PR #158, opened 2014-03-27
add SSL support

oranagra:
Adding SSL support. Tested against stunnel on the server side.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,6 +104,18 @@ $client = new Predis\Client([
 ]);
 ```
 
+When useing ssl or tls schemes you can optionally provide certificate and other SSL options like so:
+
+```php
+// SSL connection with key, cert and cafile in the local_cert file:
+$client = new Predis\Client([
+    'scheme' => 'ssl',
+    'host'   => '10.0.0.1',
+    'port'   => 6378,
+    'ssl'    => ['local_cert' => '/absolute/path/keys.pem',]
+]);
+```
+
 The actual list of supported connection parameters can vary depending on each connection backend so
 it is recommended to refer to their specific documentation for details.
 
--- a/lib/Predis/Connection/AbstractConnection.php
+++ b/lib/Predis/Connection/AbstractConnection.php
@@ -62,6 +62,10 @@ abstract class AbstractConnection implements SingleConnectionInterface
                 }
 
             case 'tcp':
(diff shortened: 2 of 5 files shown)


tomcastleman:
Hi @oranagra 

I found this PR whilst investigating the Redis Cloud service. Encryption is a key requirement when using third party services over untrusted networks. On PaaS providers like Heroku there is no option to use stunnel so the feature implemented by this PR is really important.

Have you thought about pursuing this PR along the lines @nrk suggested?

Thanks

oranagra:
Hi @tomcastleman 
I'm currently busy with other projects, if you have the time, maybe do it yourself.
Let me know if my assistance is needed to push it into the existing PR, or create a new one.
Thanks.

nrk:
@oranagra, @tomcastleman support for SSL will be added to Predis v1.1 but will be limited to `Predis\Connection\StreamConnection` (the default connection backend) at least until I find a solution to make it work for `Predis\Connection\PhpiredisStreamConnection` too (I have an idea, but also have other priorities for the library right now). I will push the code on GitHub in a couple of days, but things seems to be working fine already. The accepted URI schemes will be `tls` and `rediss`.

oranagra:
Hi,
I think that supporting SSL with the default connection class may be enough.
I don't think you can support SSL in PhpiredisStreamConnection unless support for SSL is added for hiredis first.
Thanks.

nrk:
@oranagra the `Predis\Connection\PhpiredisStreamConnection` only uses the RESP stream parser provided by hiredis, network-related stuff still happens in userland PHP code :-)

itamarhaber:
@nrk please also take into consideration the Reids URI discussion that's taking place at https://groups.google.com/forum/#!topic/redis-db/elyRy9KDRtI

nrk:
@itamarhaber both the `redis` and `rediss` schemes will be supported by Predis (well actually `redis` is already in the `v1.0` branch, will be available starting with v1.0.2) following the rules described on the respective provisional URI scheme registration documents. as opposed with `tcp` and `tls` which will stick with the current Predis-specific way of specifying both password and database.

nrk:
Implemented in commit ebb7237 available on the `master` branch (targeting the next minor release of Predis, v1.1).

PR #164, opened 2014-05-10
Recover from server disconnection by timeout

willybarro:
If Redis Server has a connection timeout set, a Predis client using the stream adapter may be left with a useless (closed) resource when Redis drops the conection per timeout.
When that happens, the user receives a Predis\Connection\ConnectionException "Error while reading line from the server.", but that one is not recoverable from outside the lib, as the connect method will return the closed resource.

More details on this patch and why it was implemented that way can be found on commit messages.

When I've set the timeout to "1" on my server, other tests started failing (see below) and are also fixed by this patch.

**Tests failing with timeout set to "1" (All of them now pass!):**
Predis\Command\KeyExpireAtTest::testCanExpireKeys
Predis\Command\KeyExpireTest::testCanExpireKeys
Predis\Command\KeyExpireTest::testConsistencyWithTTL
Predis\Command\KeyPreciseExpireAtTest::testCanExpireKeys
Predis\Command\StringSetExpireTest::testKeyExpiresAfterTTL

### DIFF
--- a/lib/Predis/Connection/AbstractConnection.php
+++ b/lib/Predis/Connection/AbstractConnection.php
@@ -24,7 +24,7 @@ use Predis\Protocol\ProtocolException;
  */
 abstract class AbstractConnection implements SingleConnectionInterface
 {
-    private $resource;
+    protected $resource;
     private $cachedId;
 
     protected $parameters;
@@ -165,7 +165,7 @@ abstract class AbstractConnection implements SingleConnectionInterface
      */
     public function getResource()
     {
-        if (isset($this->resource)) {
+        if ($this->isConnected()) {
             return $this->resource;
         }
 
--- a/lib/Predis/Connection/StreamConnection.php
+++ b/lib/Predis/Connection/StreamConnection.php
@@ -127,6 +127,14 @@ class StreamConnection extends AbstractConnection
     /**
      * {@inheritdoc}
      */
+    public function isConnected()
+    {
+        return isset($this->resource) && !feof($this->resource);    
+    }
+
+    /**
+     * {@inheritdoc}
+     */
     public function connect()
     {
         if (parent::connect() && $this->initCommands) {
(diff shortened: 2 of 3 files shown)


Milkdog:
Would really love to see this implemented. We're running into the same issue and it looks like this would solve it for us.

Seldaek:
AFAIK this just re-connects transparently in case the connection dropped right? Sounds good, I am also experiencing issues in a case where redis isn't on the same machine and long running connections sometimes get dropped.

willybarro:
I remember last year he sent me an email regarding this PR and saying something about the accessibility of the $resource property. I don't recall exactly how he wanted me to change it. This PR is old and is now failing the tests, but if you guys (and @nrk) are still interested, I'm still willing to help :)

ethanpooley:
@willybarro Could you possibly dig up that email and post it here? @Milkdog and I can participate in getting this back into shape.

My particular concern is that we pay very close attention to the pitfalls of `feof()` so as not to introduce any new symptoms. See fairly hefty warnings here: http://php.net/feof

nrk:
@ethanpooley this is the related part of the email mentioned by @willybarro:

> I'm aware of your PR but in the past I had problems (and reports of problems) when relying on feof() in terms of both performance and weird issues, so I have to think carefully about changing that part of code. I'm not saying that the current implementation is completely correct, but it's a trade-off that works unless you are disconnected by the server (and that generates the exception). Also I'm not really keen on exposing $resource to subclasses by changing its visibility, but that's a whole different matter.

Seldaek:
Couldn't we do the fix without feof? Rather catch the exception when it
occurs and attempt a reconnection/retry of the command in such cases. Or
alternatively just add an easy way to do this reconnect from the outside
and then we can have a RetryConnection class that wraps the other one and
does the catch+retry. It would be nicer as that would allow users to
customize the retry behavior (max retries, exponential back off and such).

PR #213, opened 2014-09-08
Adds magic methods to the Client phpdoc

seagoj:
Adds the basic methods (get, hget, hgetall, set, hset, expire) to the Client phpdoc. This allows mocking frameworks like [Prophecy](https://github.com/phpspec/prophecy) to mock predis objects.

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -38,6 +38,12 @@ use Predis\Transaction\MultiExec as MultiExecTransaction;
  *
  * {@inheritdoc}
  *
+ * @method get($key)
+ * @method hget($hash, $key)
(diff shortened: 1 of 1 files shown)


rodrigodiez:
@stof @nrk 

Any chance of you guys sorting out a solution for this?

Not sure how many developers I have already met who solved this lack of agreement with dirty fixes in their code base, but predis/prophecy is a recurrent topic in every meeting that I assist to. I found myself patching around this in at least 3 projects. The result is always similar: either a worse code base or lack of tests

Would you guys consider talk about this again and choose the less of the two evils considering developer use cases and good practices?

PR #229, opened 2014-12-16
Apply read_write_timeout to Unix socket connections.

srhnsn:
I noticed my Predis connections timing out on long running Unix socket connections even with read_write_timeout set to 0. It looks like it only gets set when using TCP connections.

I copied the code from the tcpStreamInitializer() method right above it which is not necessarily nice but gets the job done. Please bear with me if I missed something since this is my first contribution on GitHub ever. :)


nrk:
Thanks @srhnsn this will do for now for v1.0, I will most likely change how the underlying resource is initialized in v1.1 to reuse most of the code which is shared anyway when using PHP streams.

PR #233, opened 2014-12-19
added (failing on HHVM) test for pubsub with read_write_timeout=-1

rubensayshi:
for testing what is described (and patched) in https://github.com/nrk/predis/pull/231

fix for HHVM: https://github.com/facebook/hhvm/pull/4495 (will most likely be in next stable)


nrk:
Thanks. I added HHVM (stable) to the list of allowed failures for now, not really happy to do so but don't want to see builds marked as failed on Travis CI just for this one.

PR #235, opened 2015-01-23
add zrevrangebylex command

mihu:
please consider merging the zrevrangebylex command http://redis.io/commands/zrevrangebylex


nrk:
Thanks @mihu!

I'm really sorry that I couldn't merge your PR sooner than this especially because it was neatly prepared and you didn't miss anything, which makes it even more appreciated :+1: I had half an year off from opensource but at least better late than never. Your change will be available starting with Predis v1.0.2.

Thanks again.

mihu:
No issue, I know the feeling:) Thanks for merge!:)

PR #237, opened 2015-03-11
Emit socket error if ipredis connection has been lost/reset.

RichardHeelin:
The current code is checking for a failure (return false) or an empty buffer string, however neither of these will be the case if the connection has been reset or has errored. According to the docs for socket_recv, $buffer will be set to null if the connection is reset or their is no data. As currently null is not allowed for, we enter an infinite loop, to prevent this I've added null to the things we check before we emit a socket error. This prevents the infinite loop and correctly results in an Exception if the connection is lost/reset.

I've made this change against the 0.8 branch as that is the version I am using, however the issue still exists with later 1.0 releases.

### DIFF
--- a/lib/Predis/Connection/PhpiredisConnection.php
+++ b/lib/Predis/Connection/PhpiredisConnection.php
@@ -358,7 +358,7 @@ class PhpiredisConnection extends AbstractConnection
         $reader = $this->reader;
 
         while (($state = phpiredis_reader_get_state($reader)) === PHPIREDIS_READER_STATE_INCOMPLETE) {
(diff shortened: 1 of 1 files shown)


nrk:
Actually I'm not dead and I don't plan to die anytime soon :-)

I'm pretty much sure that in a couple of months I'll be able to resume taking care of my open source projects, but I hope to find bits of time to start clearing the backlog of pending issues and PRs as soon as possible.

nrk:
@RichardHeelin I already cherry-picked your change into `master` and `v1.0` so that your change will be available starting from `v1.0.2`. I'm still not sure if I'll push a new patch release for the `v0.8` branch, but I can merge your PR while deciding upon that :-) Thanks!

Issue #239, opened 2015-04-11
not parsing IPv6 node addresses correctly


Artlu:
Hi,

I'm using redis cluster in an IPv6 only environment. For now, I have set up my cluster with 3 masters and one master has a slave, as shown here:
```
2f4deff02f02ce70098275686acb529c08143619 fd01::85:14:243:18:6379 master - 0 1428780377665 1 connected 0-5460
879fba81ebb152c1ee7264867908fbd6c99b180a fd01::5:104:106:150:6379 myself,slave ab3e6e95138d12d303fcbbd205c31d2068d1e639 0 0 0 connected
5d6fb5ca8276cf7b1ab90cd3115f67e1f3b6bc9c fd01::85:14:243:17:6379 master - 0 1428780375658 5 connected 10923-16383
ab3e6e95138d12d303fcbbd205c31d2068d1e639 fd01::5:199:142:208:6379 master - 0 1428780376662 4 connected 5461-10922
```
My code to read/write to redis:
```
$conn = array('tcp://127.0.0.1:6379');
$options = array('cluster' => 'redis');
$c = new Predis\Client($conn, $options);
$c->set($key, $value);
```
However, I'm getting this error:
```
Array
(
 [errno] => 2
 [errstr] => stream_socket_client(): php_network_getaddresses: getaddrinfo failed: Name or service not known
 [errfile] => /usr/share/php/Predis/Connection/StreamConnection.php
 [errline] => 91
 [errcontext] => Array
 (
 [parameters] => Predis\\Connection\\ConnectionParameters Object
 (
 [parameters:Predis\\Connection\\ConnectionParameters:private] => Array
 (
 [host] => fd01
 [port] => 0
 [scheme] => tcp
 [timeout] => 5
 )

 )

 [uri] => tcp://fd01:0/
 [flags] => 4
 [errno] => 0
 [errstr] => 
 )

)
```
As you can see, the address gets parsed wrongly. Instead of taking the last number as the port and everything before as an IPv6 address, it only takes the first part as hostname/IP.

One might also consider this to be a fault of redis, as the IPv6 address isn't enclosed with brackets to make clear what's the IP and port part, like so: [fd01::85:14:243:18]:6379, but I think we have to live with that ...

nrk:
This is indeed an issue but I believe the fix proposed in #240 would solve only a part of it as we also have a problem when simply doing `$client = new Predis\Client(['host' => '::1', 'port' => 6379]);`.

nrk:
[...]

PR #240, opened 2015-04-11
IPv6 cluster compatibility

Artlu:
fixes issue #239


nrk:
No more needed, the whole issue needed more radical changes through the library (see 8aec51b).

PR #246, opened 2015-04-29
Not selecting default database

bobrik:
No reason to do `SELECT 0` on each connection. Not sure if it should go to master or v1.0.

### DIFF
--- a/src/Connection/Factory.php
+++ b/src/Connection/Factory.php
@@ -144,7 +144,7 @@ class Factory implements FactoryInterface
             );
         }
 
-        if (isset($parameters->database)) {
+        if (isset($parameters->database) && $parameters->database) {
             $connection->addConnectCommand(
                 new RawCommand(array('SELECT', $parameters->database))
             );
--- a/tests/Predis/Connection/FactoryTest.php
+++ b/tests/Predis/Connection/FactoryTest.php
@@ -130,7 +130,7 @@ class FactoryTest extends PredisTestCase
     public function testCreateConnectionWithInitializationCommands()
     {
         $parameters = new Parameters(array(
-            'database' => '0',
+            'database' => '1',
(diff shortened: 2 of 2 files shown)


nrk:
Any valid reason? I mean the `SELECT` command is sent to the server only when the `database` connection parameter is set explicitly by the user when configuring the client. Furthermore, there could be a valid reason to force `SELECT 0` when the client is configured to use persistent connections.

bobrik:
Extra round trip for every connection sounds like a valid reason. We've seen developers setting `'database' => 0` as the default configuration key without knowing the consequences.

Can you show me the code when my change breaks? If it does, it probably worth adding the test for it.

nrk:
I don't mean to sound rude to anyone, but if some developers fail to understand how certain configurations affect the behavior of the client the problem in my opinion is on user's end. In their defense, our actual issue is that this kind of nuances when configuring the client should be addressed in a more extensive set of docs but Predis is lacking documentation.

Anyway let's suppose that the client is configured to use persistent connections with a set of parameters like `tcp://127.0.0.1?database=5&persistent=1`. The client sends `SELECT 5` as soon as it connects to Redis and the underlying socket resource is left open by PHP to be reused for the next request: by keeping the socket resource around and in effect connected to Redis means that Redis is also keeping that connection associated to database 5. If the configuration of the application is then changed to use a new set of parameters like `tcp://127.0.0.1?database=0&persistent=1` (and no restart of PHP-FPM or alike is involved) some of the subsequent requests would still land on database 5 instead of 0 until the underlying socket is closed either by PHP or Redis, as there would be no explicit `SELECT 0` sent by the client.

Testing such a condition that involves persistent connection is hard and I honestly don't know how to implement a test for it using PHPUnit, especially in a deterministic way.

PR #249, opened 2015-05-08
Added phpdoc to pipeline variable

Aliance:


nrk:
Isn't @var ClientContextInterface enough since it's already referred to the variable $pipeline set below the phpdoc block? Honest question, my phpdoc-fu is admittedly weak :-)

Aliance:
Without this changes PHPStorm highlighted methods, called from `$pipeline` variable.

src/Client.php
```diff
@@ -433,8 +433,8 @@ protected function createPipeline(array $options = null, $callable = null)
             $class = 'Predis\Pipeline\Pipeline';
         }
 
-        /*
-         * @var ClientContextInterface
+        /**
```
GrahamCampbell: This change is important. `/*` is VERY different from `/**`. PHP even tokenizes it differently.

nrk: Yup that was just an oversight on my part, my doubt was about the addition of `$pipeline`.

GrahamCampbell: the `$pipeline` thing is wrong, yeh

nrk: Alright then, fixed without merging this PR (but put a note in the commit message to thank @Aliance for spotting this one).

PR #258, opened 2015-06-16
Don't check for __invoke on non-objects

### LINKED ISSUE #257: `__invoke` support on Options causes autoloading of non-existent class
In the ~1.0 `Predis\Configuration\Options`, we support `__invoke` like this:

```php
if (method_exists($value, '__invoke')) {
    $value = $value($this, $option);
}
```

But we don't check whether `$value` is an object. Unfortunately, `method_exists` can operate on string class names, and will cause autoloading. That means, using simple client options like `['cluster' => 'predis']` will cause an autoloading attempt for a class named `\predis`.

Sure, autoloading should fail, and `method_exists` should return `false`. But the autoloading should be avoided in the first place. (`method_exists` doesn't have an `$autoload = false` param like `class_exists` does.)

The simplest solution is to check for an object before looking for the `__invoke` method.

This is particularly a problem for my use of Predis, because the project I'm working on has a deep autoloader stack (ordered by likelihood of success), and a missed autoload is quite expensive. (A couple of seconds! Yeah, crazy, I know.)

dominics:
Fixes #257

Couldn't think of a good way to test autoloading within PHPUnit other than registering an actual autoloader, or putting a never-loaded class somewhere the existing autoloader can get to it (and asserting on the value of `class_exists(X, false)`). That seemed a bit messy, so I've left a regression test out (all code paths still covered by existing tests), but could do one if you want.

### DIFF
--- a/src/Configuration/Options.php
+++ b/src/Configuration/Options.php
@@ -100,7 +100,7 @@ class Options implements OptionsInterface
             $value = $this->input[$option];
             unset($this->input[$option]);
 
-            if (method_exists($value, '__invoke')) {
+            if (is_object($value) && method_exists($value, '__invoke')) {
                 $value = $value($this, $option);
             }
 


GrahamCampbell:
Why did this go to master and not 1.0?

nrk:
@GrahamCampbell because I pushed the `master` branch on GitHub but not `v1.0` yet ;-) I manually cherry-picked it into `v1.0`.

nrk:
@dominics in the end I implemented a test against regressions by temporarily registering a new autoloader function, quick and easy.

PR #261, opened 2015-07-02
add is_object before check exsist invoke method in variable

WildZero:
Issue https://github.com/nrk/predis/issues/257

### DIFF
--- a/src/Configuration/Options.php
+++ b/src/Configuration/Options.php
@@ -100,10 +100,8 @@ class Options implements OptionsInterface
             $value = $this->input[$option];
             unset($this->input[$option]);
 
-            if (is_object($value)) {
-                if (method_exists($value, '__invoke')) {
-                    $value = $value($this, $option);
-                }
+            if (is_object($value) && method_exists($value, '__invoke')) {
+                $value = $value($this, $option);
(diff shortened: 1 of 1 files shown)


src/Configuration/Options.php
```diff
@@ -100,8 +100,10 @@ public function __get($option)
             $value = $this->input[$option];
             unset($this->input[$option]);
 
-            if (method_exists($value, '__invoke')) {
-                $value = $value($this, $option);
+            if (is_object($value)) {
+                if (method_exists($value, '__invoke')) {
```
xabbuh: Why not simply `is_object($value) && method_exists($value, '__invoke'))`?

WildZero: It's my habit: separate different logical condition.

GrahamCampbell: Yeh, this should be `is_object($value) && method_exists($value, '__invoke'))`.

WildZero:
I'v fixed it

nrk:
Isn't this one just a duplicate of the already open PR #258?

GrahamCampbell:
Yeh, this is a duplicate.

PR #278, opened 2015-09-01
Slim down composer package by removing examples directory from release.

jdufresne:


nrk:
I'm not really keen to exclude the examples directory and the other *.md files especially considering their neglible impact in terms of additional bandwidth and disk occupation. This change would affect the archives generated by GitHub that can have use cases beyond composer.

Is this a trend or a practice among other libraries or just a matter of personal preference?

jdufresne:
I found the following article giving one opinion: <http://blog.madewithlove.be/post/gitattributes/>

Keep in mind that many people use composer to download dependencies for production sites. Downloading third party documentation and examples can be time consuming and goes to waste on these sites. The examples are still available to those who want them, but not forced on people during production deployment. As more and more 3rd party packages include examples, documentation and tests, the extra time spent during deployment of these sites really adds up. I've noticed these in my own deployments.

PR #279, opened 2015-09-02
Support authentication in combination with Redis Cluster

wouterlemcke:


nrk:
If the nodes of your cluster require a password you can configure the client (Predis v1.0) with the following code:

```php
$client = new Predis\Client($nodesarray, [
    'password' => $password,
    'cluster' => function ($options) {
        $cluster = new Predis\Connection\Aggregate\RedisCluster($options->connections);
        $cluster->setDefaultParameters(['password' => $options->password]);

        return $cluster;
    },
]);
```

You PR doesn't make sense because `AUTH` can __only__ be used directly against each single node. To be more specific, `$getKeyFromFirstArgument()` (as the name implies) expects a __key__ that is used to identify a specific node among the ones in the cluster as the actual target of a particular operation against a string value or data structure. Obviously this is not the case of `AUTH`.

Issue #86, opened 2012-11-02
Incorrect work with mbstring.func_overload set


msukhotin:
Hi! I have webserver with php 5.3.16 installed. 
mbstring.func_overload is set to 2 (software requirements).

Here's a php file which causes error:

```
require "Predis/Autoloader.php";
Predis\Autoloader::register();

try {
    $redis = new Predis\Client();

    $redis->set("LANG", "тест"); // any utf-8 encoded string
    $redis->set("LANG", "тест2");

}
catch (Exception $e) {
    echo $e->getMessage();
}
```

Output: ERR unknown command 'т'

If I set func_overload off it works ok. Problem is strlen which returns number of chars but not the number of bytes (PhpiredisConnection.php, function write())

PR #283, opened 2015-09-28
Example of a multibyte bug test case

urg:
Primarily adding this as a sample bug report, along with a semi-fix.  I think that probably some installation's won't have the mb_* functions, so perhaps it shoud be wrapped?


nrk:
Predis works perfectly fine with multibyte strings as-is since it treats data as binary strings and has no concept of text encoding. Could it be that your issues with multibyte strings is actually related to #86?

urg:
If you take the take the test case in the PR and run that without the changes, it will fail (at least on my computer).  It seems like that ought to work?

urg:
Sorry, you're right - that appears to the issue.  Thanks

PR #285, opened 2015-10-27
Update ClientInterface phpDoc

cheprasov:
I realized that with zrevrangebyscore() the min/max parameters must be reversed
Please, see http://redis.io/commands/zrevrangebyscore


nrk:
Whoops :-)

PR #288, opened 2015-11-09
Invalid rpush definition.

bkrukowski:


nrk:
This change is inconsistent with other signatures documented with `@method` tags as we decided to use _array_ in case of variadic commands such as `RPUSH`.

PR #299, opened 2016-01-04
When using redis cluster, add option to cache 'CLUSTER SLOTS' response

lsnider:
This is an optimization to redis cluster using `useClusterSlots(true)`.

Currently on newly instantiated PHP processes, the first called redis command may result in two extra round trips to the redis cluster.  The first call could result in a `MOVE`.  We then make a call to the new cluster calling `CLUSTER SLOTS` to learn which hosts the slots are on.  The rest of the calls in this PHP process should be correct but on the next PHP spawn, we've forgotten the location of the slots.

This PR add a `setCacheSlotMapDirectory($directory)` method.  If set, predis will save the results of a `CLUSTER SLOTS` lookup and then load the results on newly created instances saving the two extra round trips on boot.

The PR includes unit tests for the new functionality.

I tried to match the existing coding style of predis.  Let me know if interested in the PR but would like to see some changes before merging the code upstream.

### DIFF
--- a/src/Connection/Aggregate/RedisCluster.php
+++ b/src/Connection/Aggregate/RedisCluster.php
@@ -55,6 +55,8 @@ class RedisCluster implements ClusterInterface, \IteratorAggregate, \Countable
 	private $cacheSlotMapDirectory;
 	private $cacheSlotMapFile;
 
+	const CACHED_FILENAME_PREFIX = 'predis-cached-slots';
+
     /**
      * @param FactoryInterface  $connections Optional connection factory.
      * @param StrategyInterface $strategy    Optional cluster strategy.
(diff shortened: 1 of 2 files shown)


nrk:
Hi @lsnider,

I don't think that how or where the slots map is cached should have anything to do with the internals of the cluster connection class. In this case I think it's better to disable the use of `CLUSTER SLOTS` when Predis connects to redis-cluster and have the slots map saved / updated by an external process, then you can just read the configuration back from your application in order to initialize the client. Remember that _slots hints_ can be embedded in connection parameters to supply a proper cluster configuration:

```php
// The following array can be unserialized / recreated from your external slots map cache
$cluster = [
    'tcp://127.0.0.1:6379?slots=0-5460',
    'tcp://127.0.0.1:6380?slots=5461-10921',
    'tcp://127.0.0.1:6381?slots=10922-16383',
];

$options = [
    'cluster' => function ($options) {
        $connection = new Predis\Connection\Aggregate\RedisCluster($options->connections);
        $connection->useClusterSlots(false);

        return $connection;
    },
];

$client = new Predis\Client($cluster, $options);
```

wedgwood:
@lsnider  
16k array is not small. there is performance penalty when building large slots map. 
So, maybe right `guessNode` (cache cluster connection dsn without slots's infomation) would be better performance.

Issue #298, opened 2016-01-04
Implement `WAIT` command


tPl0ch:
Hello,

currently we want to use the `WAIT` - http://redis.io/commands/WAIT - command to increase data safety for certain write requests in the `redis-cluster` mode. Since `WAIT` is not implemented, we were trying to use a raw command:

``` php
$redisSuccess = $this->redis->set($key, $value, 'ex', $ttl, 'nx');
// Wait for writes to propagate through the cluster - TTL of 0.5 seconds
$slavesReached = (int) $this->redis->executeRaw(['WAIT', 2, 500]);
if ($slavesReached < 1) {
    throw new ClientException("Couldn't distribute write to slaves");
}
```

Unfortunately it tries to get a `Slot` for the `WAIT` command and fails with:
`Cannot use 'WAIT' with redis-cluster.`

Can we somehow manage to get the `WAIT` command working within the current architecture, by i.e. supplying an own `ClusterStrategy`?

Thanks in advance,
Thomas

tPl0ch:
@nrk 
> When using a client configured in cluster mode the only restriction is that all of the writes ... must be directed to the same slot

Hmm, this is dramatically limiting the usefulness of `WAIT` in a cluster setup. You have a set of key/value pairs that you want to ensure to have reached a majority of nodes, even if they are routed to different slots, don't you think?

nrk:
@tPl0ch the problem when using a cluster setup is that `WAIT` must be sent to each node interested by a write operation and, as far as I understand, this is valid for redis-cluster too. Predis cannot do this asynchronously so the first `WAIT` will block until acknowledgement / timeout before sending the next one and so forth. After that, there's the problem of how to handle different acknowledgement outcomes among the various "slots" that are target of writes.

tPl0ch:
@nrk Then it will be time to leverage coroutines and/or promises to handle cluster-wide commands in a non-blocking way :computer:

nrk:
[...]

PR #300, opened 2016-01-05
Implementing `WAIT` command

tPl0ch:
See https://github.com/nrk/predis/issues/298 for my motivation on this.

When using `redis-cluster` with a version >= 3.0, the `WAIT` command comes in handy to increase data safety with important writes by making sure that the write has been acknowledged by `n` (number of slaves).

This PR implements the command for versions >= 3.0

Since this is the first time I was digging into the `Predis` src, it might be that I missed some obvious things, although the tests are passing for me.

The tests have been set up locally with a `redis-cluster` created by the `cluster-create` util in the redis distribution.~~I still need to investigate on how to set this up in travis though, maybe someone of you can help. :)~~ I have added a Makefile to set up a redis cluster (v3.0) within travis.

### DIFF
--- a/src/Cluster/RedisStrategy.php
+++ b/src/Cluster/RedisStrategy.php
@@ -55,23 +55,4 @@ class RedisStrategy extends ClusterStrategy
             'This cluster strategy does not provide an external distributor'
         );
     }
-
-    /**
-     * @inheritDoc
-     */
-    protected function getDefaultCommands()
-    {
-        $commands = parent::getDefaultCommands();
-        $commands['WAIT'] = array($this, 'getSlotForRandomKey');
-
-        return $commands;
-    }
-
-    /**
-     * @return int
-     */
-    protected function getSlotForRandomKey()
-    {
-        return $this->hashGenerator->hash(uniqid('', true)) & 0x3FFF;
-    }
 }
--- a/src/Cluster/RedisStrategy.php
+++ b/src/Cluster/RedisStrategy.php
@@ -55,4 +55,23 @@ class RedisStrategy extends ClusterStrategy
             'This cluster strategy does not provide an external distributor'
         );
     }
+
+    /**
+     * @inheritDoc
+     */
+    protected function getDefaultCommands()
+    {
+        $commands = parent::getDefaultCommands();
+        $commands['WAIT'] = array($this, 'getSlotForRandomKey');
+
+        return $commands;
+    }
+
+    /**
+     * @return int
+     */
+    protected function getSlotForRandomKey()
+    {
+        return $this->hashGenerator->hash(uniqid('', true)) & 0x3FFF;
+    }
 }
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,11 +14,16 @@ branches:
     - php5.2_backport
     - documentation
 services: redis-server
+install: make travis-install
 before_script:
   - composer self-update
   - composer install --no-interaction --prefer-source --dev
+  - make travis-run
 script:
   - vendor/bin/phpunit -c phpunit.xml.travisci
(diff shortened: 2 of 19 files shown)


src/Cluster/RedisStrategy.php
```diff
@@ -55,4 +55,15 @@ public function getDistributor()
...
         );
     }
+
+    /**
+     * @inheritDoc
+     */
+    protected function getDefaultCommands()
+    {
+        $commands = parent::getDefaultCommands();
+        $commands['WAIT'] = array($this, 'getKeyFromFirstArgument');
```
nrk: I'm not convinced that this is the right approach to solve #298, the first argument of `WAIT` is not a key so it doesn't make much sense. I'm going to elaborate on this in a comment to your PR.

tPl0ch: @nrk Yeah, I was thinking the same. The problem is that there is no way to bypass the hash generator for the cluster strategy. But maybe it makes sense to return a random slot, since Redis is tracking the writes internally for the current client connection (see http://redis.io/commands/WAIT#implementation-details). Does this makes sense to you?

tPl0ch: @nrk Something like this:
```php
/** 
 *@return int
 */
protected function getSlotForRandomKey()
{
    return $this->hashGenerator->hash(uniqid('', true)) & 0x3FFF;
}
```

nrk:
First of all thanks for your PR @tPl0ch! It would be awesome if you could extract the part that sets up redis-cluster on Travis CI into a separate PR because I'm definitely interested to have that merged ASAP :-)

Back to the actual issue, as I've said in my inline comment I think that using a command handler to return a value that is not a key (in the case of `WAIT` the first argument is just the number of slaves) does not make much sense. But we could still play along these lines by changing our internal assumptions slightly.

Right now [`Predis\Cluster\ClusterStrategy::getSlot()`](https://github.com/nrk/predis/blob/v1.0.3/src/Cluster/ClusterStrategy.php#L336-L350), which is the actual method used by aggregate connections to obtain a valid slot for the specified command, is expected to return either:

- a slot if it can be calculated from a valid key or multiple keys
- `NULL` when the command is not supported... which until now just meant that it cannot be mapped to one specific slot.

We might just change that to augment the possible return values by including boolean `FALSE` to indicate that the command __can__ be used in the context of a certain cluster implementation even though it cannot be mapped to a specific slot. At this point when  `FALSE` is returned [our implementation](https://github.com/nrk/predis/blob/v1.0.3/src/Connection/Aggregate/RedisCluster.php#L294-L309) could stick to the last used node or just a random one to issue the command.

Thoughts? To tell you the truth I'm not a fan of having more than one type of value returned by `Predis\Cluster\ClusterStrategy::getSlot()` (or any method for that matter) but from a pragmatic POV I think it does not add much complexity to the library internals.

src/Cluster/RedisStrategy.php
nrk: Sorry I left for lunch with my comment nearly done but didn't notice your further inline comments before posting it :-) Anyway if we think about this in terms of which __node__ should be used instead of which __slot__, then I guess it would make more sense to let the decision to the actual implementation of the aggregate connection.

tPl0ch:
@nrk I made a little prototype change now: https://github.com/nrk/predis/commit/ab4296db8128c46456b24af919f9b5e7e3b5faae

Thoughts on this?

nrk:
I guess I'll need some time to think about it and play with some actual code before carrying on with the changes, in the end it still feels a bit too hackish :-)

mente:
Any updates on it? Changes looks promising

Issue #131, opened 2013-07-13
Support for Redis Sentinel


vmattila:
I would be happy to contribute to predis development by implementing automatic master/slave resolution and connection based on the configured sentinels. I tried to explore the Predis codebase, but unfortunately could not figure out where to start.

In practice, we should be able to enter the sentinel addresses instead direct master/slave ip addresses and provide the name of the master. For example:

```php
$sentinels = array(
    'tcp://10.0.0.1:26379/',
    'tcp://10.0.0.2:26379/'
);
$redis = new Predis\Client($sentinels, array('sentinel_master' => 'mymaster');
```

One of the configured sentinels should be connected and sent queries to get the master ip: `SENTINEL get-master-addr-by-name mymaster`, slaves `SENTINEL slaves mymaster` and finally create the actual aggregated connection based on the received information.

Should I create a completely new connection type, extend an existing class or modify something? I also expect that I should implement two new command classes for the two sentinel commands.

Some references:
- http://redis.io/topics/sentinel
- http://redis.io/topics/sentinel-clients

ThijsFeryn:
Any updates, Daniele?

Thx
Thijs

integrii:
Are we sure that you can read/write to any node in the cluster?  The six node cluster I have setup replies with MOVED and the correct node's location.  I then have to connect to that node to complete the write.  Making Predis aware of shards means the process goes faster.  When a MOVED comes back from a server, Predis should ask again for the node distribution.

i could see this being slightly wasteful in small applications that only do one read or write, but long running applications would benefit from it greatly.  Maybe caching the node information should be an option you can enable?

ThijsFeryn:
Hi @nrk & @vmattila 

I considered your advice again (using Sentinel as a strategy rather than a replication connection type). I tried implementing it as a strategy, but it felt wrong.

If you look at the problem Sentinel is trying to solve, it's autofailover and that's directly linked to the replication. Unlike the other strategy patterns, this is initializing new connections.

I believe the implementation of @vmattila is still the right decision.

Would you please consider merging this? I could really use this feature right now. And as mentioned in one of the comments above, Redis Cluster has its own autofailover mechanism. 

Love to hear you opinion and plans for this feature.

Thanks again
Thijs

vmattila:
Just got an email from a user testing my branch. He pointed out a problem with authentication, just copy-pasting the message here so we don't forget the issue:

"I'm testing your feature sentinel for predis.
[...]

PR #306, opened 2016-01-27
Replace automatic retry yes/no with a retry limit

crispygoth:
I've been testing the sentinel code (from issue #131) and so far the only issue I've noticed is the behaviour of the automatic retries when the Redis server is unavailable. In my testing, with auto retries turned on I sometimes had the application get stuck in an infinite loop if it was unable to contact the server, which is not ideal. However I would like it to not fail immediately as well, so I replaced the autoRetry boolean with a retryLimit integer, so you can tell it to retry a few times before giving up.

I didn't go for maintaining compatibility since this hasn't hit a stable release. If you'd prefer I could update my PR with a setAutomaticRetry method which preserves the old behaviour.

### DIFF
--- a/src/Connection/Aggregate/SentinelReplication.php
+++ b/src/Connection/Aggregate/SentinelReplication.php
@@ -56,9 +56,9 @@ class SentinelReplication extends MasterSlaveReplication
     protected $sentinelTimeout = 0.100;
 
     /**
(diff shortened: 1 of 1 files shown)


nrk:
Thanks @crispygoth!

PR #307, opened 2016-02-03
Master-Slave Replication - client side read-slave failover

umwerk-mq:
We are using a Master-Slave Redis server. 
The availability of the master is handled on the server side (i.e. a read-slave is promoted to master).
However, if one of the read-slave is not reachable, the error handling on the server side is not quick enough. The read-slave gets replaced, but it can take up to 15 min until it works again.
During those 15 min, 1 out of n read-requests will fail (where n is the number of slaves).  To avoid this we modified the MasterSlaveReplication::executeCommand() method a little bit by adding a catch block for the ConnectionException. If it is thrown, the command is resent to one of the other available slaves.

### DIFF
--- a/src/Connection/Aggregate/MasterSlaveReplication.php
+++ b/src/Connection/Aggregate/MasterSlaveReplication.php
@@ -11,9 +11,12 @@
 
 namespace Predis\Connection\Aggregate;
 
+use InvalidArgumentException;
+use RuntimeException;
 use Predis\Command\CommandInterface;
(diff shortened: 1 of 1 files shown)


nrk:
This feature has been implemented (although in a [different way](https://github.com/nrk/predis/commit/b491dff1267b89806018d179448693bff4d2cf11)) and will be available in Predis v1.1.0. Thanks anyway for taking the time to push this PR!

PR #311, opened 2016-02-16
Update useClusterNodes() docs to indicate default

JohnMaguire:
Per my comment on #91


nrk:
Cherry-picked (see 11adccd), thanks!

PR #312, opened 2016-02-16
buildSlotsMap() to handle multiple port ranges

kenotr0n:
During a cluster re-sharding event, each cluster master will have its served slot ranges mangled while the re-sharding is in progress.  Running CLUSTER NODES during such an event will show the affected masters as serving several non-contiguous ranges of ports.

This diff adds the ability to specify multiple port ranges in the 'slots' parameter to the connection URLs to minimize the number of MOVED responses that must be handled before the cluster map is corrected.

CLUSTER SLOTS parsing still works fine during these cases, but if one wanted to pre-configure which slots are on which hosts, this patch is necessary for such times.

I've also cleaned up a couple array accesses using list assignments and explode() calls.


src/Connection/Aggregate/RedisCluster.php
```diff
@@ -276,11 +280,11 @@ protected function guessNode($slot)
      */
     protected function createConnection($connectionID)
     {
-        $separator = strrpos($connectionID, ':');
+        list($host,$port) = explode(':', $connectionID, 2);
```
nrk: This change breaks some stuff as it reverts a fix purposely shipped in a7ee807 that aimed to make Predis as a whole working with IPv6 addresses, see #239 for more details.

nrk:
Hi @kenotr0n,

Indeed, it makes perfect sense to be able to specify different slot ranges via client configuration. There's one issue with your changes related to array accesses though (see my other inline comment for details), in general I suggest to submit smaller pull requests focused only on a certain issue or feature and submit other pull requests to cover any other unrelated change.

src/Connection/Aggregate/RedisCluster.php
kenotr0n: Ah, that does make sense.  The strrpos() did look a little strange.  My main improvement was just to replace blind array accesses w/ the multiple assignment.  I try to always use symbolic names instead of array indexes to make the code easier to read.  How about a little trickery to ensure only the rightmost ':' is split on:

```
  list($port, $host) = array_map('strrev', explode(':', strrev($connectionID), 2));
```
Just a little string reversal, so that a connection id like e.g. '::1:6380' would first get reversed to '0836:1::', then split into two elements around the first colon:  array('0836','1::').  Then the elements reversed again into array('6380','::1'), and the results assigned to variables.

Somehow those substr() and strrpos() calls are bugging me more than the double-reversal above.  What say you?

cebe: if your goal is to make code more readable/understandable we should go with strrpos instead of reversing the IP parts :)

nrk:
I implemented the proposed feature but with a few changes which means I didn't merge @kenotr0n's pull request (thanks anyway!). Aside from being able to specify multiple slot ranges now it is also possible to specify a single slot, but to do that I had to avoid using `list($first, $last)` because PHP would show a notice error after the `explode()` and suppressing the notices with `@` slows things down quite a bit. Here is an example:

```php
  $parameters = [
    'tcp://10.0.0.1:6379?slots=0-5460,5500-5600,11000',
    'tcp://10.0.0.2:6379?slots=5461-5499,5600-10921',
    'tcp://10.0.0.3:6379?slots=10922-10999,11001-16383',
  ];
```

Right now these changes are in the `master` branch for v1.1.0.

Thanks!

PR #315, opened 2016-03-03
Add new script command: ZDIFF.

kevinsimard:
**Returns the members of the sorted set resulting from the difference between the first sorted set and all the successive sorted sets.**

Please note that I don't have much experience with LUA and Redis so there might be a better way to do this but I thought sharing what I'm currently using to get the difference between sorted sets might be helpful to others.

```php
// key1 = {a,b,c,d}
// key2 = {c}
// key3 = {a,c,e}

$client->zdiff('key1', 'key2');             // {a,b,d}
$client->zdiff('key1', 'key2', 'key3');     // {b,d}
```

```php
/**
 * {@inheritdoc}
 */
public function getScript()
{
    return <<<LUA
local command = {'ZUNIONSTORE', 'predis:zdiff:tmp', #ARGV}

for i=1, #ARGV do table.insert(command, ARGV[i]) end

table.insert(command, 'WEIGHTS')

for i=1, #ARGV do
    if i == 1 then table.insert(command, '1')
    else table.insert(command, '0') end
end

table.insert(command, 'AGGREGATE')
table.insert(command, 'MIN')

redis.call(unpack(command))

return redis.call('ZRANGEBYSCORE', 'predis:zdiff:tmp', '1', '+inf')
LUA;
}
```


nrk:
Hi @kevinsimard, honestly I think it does not make much sense to ship custom Lua-based script commands along with Predis as it would easily grow into something out of control, but I'm pretty sure it could make for an interesting side project (that is, something like a repository for such scripts).

kevinsimard:
I understand, I just wanted to share this with you guys in case someone needed a solution for zdiff until they add it.

xlcrr:
Any plans to integrate zdiff?

PR #322, opened 2016-04-10
Fail loudly if connection initialization fails

Seldaek:
Right now if the `SELECT x` call fails for any reason, the process will continue on the wrong database. I don't know why this would happen, but we have some data in the db 0 that should not be there, so all I can think of is that sometimes it silently fails to switch DB.

### DIFF
--- a/src/Connection/PhpiredisSocketConnection.php
+++ b/src/Connection/PhpiredisSocketConnection.php
@@ -15,6 +15,8 @@ use Predis\Command\CommandInterface;
 use Predis\NotSupportedException;
 use Predis\Response\Error as ErrorResponse;
 use Predis\Response\Status as StatusResponse;
+use Predis\Response\ErrorInterface as ErrorResponseInterface;
+use Predis\Response\ResponseInterface;
 
 /**
  * This class provides the implementation of a Predis connection that uses the
@@ -308,7 +310,10 @@ class PhpiredisSocketConnection extends AbstractConnection
     {
         if (parent::connect() && $this->initCommands) {
             foreach ($this->initCommands as $command) {
-                $this->executeCommand($command);
+                $response = $this->executeCommand($command);
+                if ($response instanceof ResponseInterface && $response instanceof ErrorResponseInterface) {
+                    $this->onConnectionError($command->parseResponse($response), 0);
+                }
             }
         }
     }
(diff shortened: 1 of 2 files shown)


src/Connection/PhpiredisSocketConnection.php
```diff
@@ -308,7 +310,10 @@ public function connect()
     {
         if (parent::connect() && $this->initCommands) {
             foreach ($this->initCommands as $command) {
-                $this->executeCommand($command);
+                $response = $this->executeCommand($command);
+                if ($response instanceof ResponseInterface && $response instanceof ErrorResponseInterface) {
```
nrk: Actually `Predis\Response\ErrorInterface` extends `Predis\Response\ResponseInterface`, there's no need for the first check.

src/Connection/PhpiredisSocketConnection.php
```diff
@@ -308,7 +310,10 @@ public function connect()
...
+                    $this->onConnectionError($command->parseResponse($response), 0);
```
nrk: Error responses are not meant to be passed to `Predis\Command\CommandInterface::parseResponse()`.

nrk:
I'm not sure how `SELECT` could have failed in your specific case but I'd say it makes sense to throw an exception early-on upon connect() as other common failures are:

- `SELECT` using a database index greater than the configured number of databases in `redis.conf`.
- `AUTH` with a wrong password.

In these cases it's indeed useful to have an exception being thrown.

src/Connection/PhpiredisSocketConnection.php
Seldaek: How should I handle this then? Any chance you take over the PR if it's just a couple changes? I am not so familiar with the predis internals.

nrk: `$response->getMessage()` is enough, anyway don't worry I'll make these changes myself (and add a couple of test cases).

Seldaek: Great thanks :)

nrk:
Fixed in `master` so that it will make it into v1.1.0, I'm just not sure about backporting this change to v1.0.

Seldaek:
Yeah that sounds fine to me :)

PR #325, opened 2016-04-20
Fix bug when dsn is redis://password@localhost

thomasglachant:


nrk:
This change does not make much sense in my opinion: the authentication part of an URI actually requires both a username and a password or a username with an empty password at worts, but we know that redis has no concept of username and that's why for passwords we actually use a `password` parameter in the query string (that is `redis://localhost?password=redispassword`)

Issue #273, opened 2015-08-19
HashKey Iterator


guizdiabolik:
in Predis\Collection\Iterator\HashKey, 

    /**
     * {@inheritdoc}
     */
    protected function extractNext()
    {
        $this->position = key($this->elements);
        $this->current = array_shift($this->elements);
    }

If keys of $this->elements contain numerical values, array_shift re-indexes the array.
So, when we iterate the collection, we lost the original values after the first call of array_shift.

using the same method as Predis\Collection\Iterator\SortedSetKey fix the problem.

    /**
     * {@inheritdoc}
     */
    protected function extractNext()
    {
        if ($kv = each($this->elements)) {
            $this->position = $kv[0];
            $this->current = $kv[1];

            unset($this->elements[$this->position]);
        }
    }

itailulu:
Thanks for the fix.
Any estimation when this fix will be released?

nrk:
Fixed in 1399d38. Patch release v1.0.4 is due in a few days.

PR #330, opened 2016-04-28
Make numeric field names work with HashKey iterator

apeschar:
`array_shift` reindexes arrays that have only numeric keys. This causes all keys after the first to be incorrect when retrieving (parts of) a hash that contains only numeric keys.

This pull request replaces `array_shift` with a manual `unset`.


apeschar:
Fixes #273.

nrk:
Thanks for the PR but I preferred to use the same code (see 1399d38) I used to fix the very same issue that also affected `ZSCAN` (not sure how it went unnoticed in `HSCAN` until now).

PR #339, opened 2016-05-28
Fix parameter handling for sentinel

phofmann-trust:
Parameter handling for redis sentinel did not work. Overwriting sentinel options did not work due to incorrect use of array()+array() and the case of a ParametersInterface being passed was not handled at all.

### DIFF
--- a/src/Connection/Aggregate/SentinelReplication.php
+++ b/src/Connection/Aggregate/SentinelReplication.php
@@ -18,6 +18,7 @@ use Predis\Connection\ConnectionException;
 use Predis\Connection\FactoryInterface as ConnectionFactoryInterface;
 use Predis\Connection\NodeConnectionInterface;
 use Predis\Connection\Parameters;
+use Predis\Connection\ParametersInterface;
 use Predis\Replication\ReplicationStrategy;
 use Predis\Replication\RoleException;
 use Predis\Response\ErrorInterface as ErrorResponseInterface;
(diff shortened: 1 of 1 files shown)


nrk:
Thanks for spotting, there was indeed a bug as Predis didn't override the values of `password` and `database` from user-supplied parameters, but your fix is still wrong because now it doesn't let users override the default timeout for sentinels (set to 100ms) by specifying a value for `timeout`. I pushed a commit with a different fix and some tests to avoid regressions.

As for not handling instances of `Predis\Connection\ParametersInterface` this is the intended behaviour here, we go on the assumption that when users are passing them to `Predis\Client::__construct()` they know what they are doing just like when passing fully-initialized connection instances.

phofmann-trust:
@nrk What's about the ParametersInterface stuff from my commit. When using SncRedisBundle $parameters is no array, but an instance of ParametersInterface. Then the unset / timeout stuff is not done, which I think is an error.

nrk:
I don't know how SncRedisBundle works internally but it shouldn't feed `Predis\Client::__construct()` directly with an instance (or an array of instances) of `Predis\Connection\ParametersInterface`, there's simply no need to do that unless for very specific and obscure reasons that I honestly can't think of. If it does then I think it should be updated to strip `password` and `database` when the `replication` client option is set to `sentinel`, and provide a default `timeout` for the sentinels.

Adding support for redis-sentinel in a minor release without causing any breaking change while still providing a decent way to configure the client proved to be quite a challenge for how the internals of Predis are designed right now, some things will be changed in v2.0 to improve upon that but right now I'm not so keen to change the current established convention through the whole library that any instance of `Predis\Connection\ParametersInterface` or `Predis\Connection\ConnectionInterface` passed to the client's constructor is left untouched.

phofmann-trust:
Ok, thank you for the explanation (and this great library)! If I have time, I'll check what exactly SncRedisBundle is doing and why it is doing it, and will try to improve. I can live with using my own branch of Predis with the workaround for now.

PR #340, opened 2016-06-03
Added ClusterKeyspace iterator

crispygoth:
This is like the regular Keyspace iterator, but works across all nodes in a Redis Cluster by scanning each node in turn.

### DIFF
--- a/src/Collection/Iterator/ClusterKeyspace.php
+++ b/src/Collection/Iterator/ClusterKeyspace.php
@@ -1,0 +1,77 @@
+<?php
+
+/*
+ * This file is part of the Predis package.
+ *
+ * (c) Daniele Alessandri <suppakilla@gmail.com>
+ *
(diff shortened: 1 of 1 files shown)


lulyon:
Out of curiosity, What is it for？and what about the performance？

nrk:
@crispygoth honestly I can't see a compelling reason to add a new class at this point when you can compose the very same functionality simply by doing:

```php
foreach ($client as $node) {
    foreach (new \Predis\Collection\Iterator\Keyspace($node) as $key) {
        // DO STUFF
    }
}
```

What's the added benefit of a dedicated iterator?

crispygoth:
Actually, you're right.. I don't even know why that didn't occur to me in the first place! Sorry for the noise, I'll close this PR :)

nrk:
@crispygoth don't worry, thanks for the PR anyway :-)

PR #344, opened 2016-06-14
Correcting the PHPDocs

felixsand:
Correcting the PHPDocs for bitfield

### DIFF
--- a/src/ClientInterface.php
+++ b/src/ClientInterface.php
@@ -46,7 +46,7 @@ use Predis\Profile\ProfileInterface;
  * @method int    append($key, $value)
  * @method int    bitcount($key, $start = null, $end = null)
  * @method int    bitop($operation, $destkey, $key)
- * @method array  bitfield($key, ...)
+ * @method array  bitfield($key, $subcommand, ...$subcommandArg)
  * @method int    decr($key)
  * @method int    decrby($key, $decrement)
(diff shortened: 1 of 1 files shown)


nrk:
@felixsand is `$subcommand, ...$subcommandArg` valid phpdoc? Just asking since my phpdoc-fu is not really strong :-) BTW this should also be fixed in [`Predis\ClientContextInterface`](https://github.com/nrk/predis/blob/v1.1/src/ClientContextInterface.php#L41).

cebe:
method definition in phpdoc should be valid PHP, which this would be with php 5.6 http://php.net/manual/en/functions.arguments.php#functions.variable-arg-list

nrk:
@cebe well the last time I used an IDE that supported phpdocs the end result was a bit of a mixed bag, that why I was asking if that signature was properly recognized in annotations despite being valid PHP code :-)

cebe:
before this change, PHPStorm reports syntax error, afterwards its fine. So :+1: from me.

nrk:
@cebe great thanks, I'll merge this PR tomorrow.

felixsand:
Sorry for not responding on time. Was on vacation :-)
Thanks @cebe and @nrk !

nrk:
@felixsand no problem, don't worry and thanks for the fix :)

PR #346, opened 2016-06-14
Don't send AUTH and/or SELECT command after connecting to sentinels

phofmann-trust:
Connecting to sentinels is broken with v1.1. Bad commit: https://github.com/nrk/predis/commit/5a0dfc36020970d5ab5f58474b01797515df49ea#diff-e55f11b907debe0b57a114d0c38dc563R244.

AUTH and/or SELECT commands are sent to sentinels when password or database is specified. Sentinels don't understand these commands and thus master discovery fails and a Predis\ClientException with message "No sentinel server available for autodiscovery." is thrown.

We need to explicitly specify a null password and database.

### DIFF
--- a/src/Connection/Replication/SentinelReplication.php
+++ b/src/Connection/Replication/SentinelReplication.php
@@ -239,9 +239,10 @@ class SentinelReplication implements ReplicationInterface
         }
 
         if (is_array($parameters)) {
-            // We unset "password" and "database" from user-supplied parameters
-            // as they are not needed when connecting to sentinels.
-            unset($parameters['database'], $parameters['password']);
+            // We explicitly set "database" and "password" to null,
+            // so that no AUTH and SELECT command is send to the sentinels.
+            $parameters['database'] = null;
+            $parameters['password'] = null;
 
             if (!isset($parameters['timeout'])) {
                 $parameters['timeout'] = $this->sentinelTimeout;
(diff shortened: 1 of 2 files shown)


nrk:
My bad, it was actually the intended behavior but I failed to notice that stripping the `password` and `database` keys from the user-supplied parameters array using `unset()` would not have prevented the connection factory from applying the `password` and `database` parameters coming from the global `parameters` client option. Having both keys defined but set to `NULL` takes the precedence over the default parameters applied by the connection factory, fixing the issue.

I was never really satisfied with the code I wrote for the configuration of the redis-sentinel backend but it was the only way to provide an easy configuration for users without pushing any breaking change, now I'm definitely convinced that this needs to be reworked for v2.0 (BCs allowed).

AaronJan:
Just updated my project and got this error.. 😭

phofmann-trust:
@AaronJan For the time being this works for me in composer.json:
`         "predis/predis": "^1.1.1@dev",`

nrk:
@AaronJan this fix will be available starting with v1.1.1 but it's still unreleased (I plan to ship it tomorrow), if you need it ASAP you can tweak your `composer.json` accordingly.

AaronJan:
Thank you guys! I'll try that. 😄

nrk:
@AaronJan JFYI [Predis v1.1.1](https://github.com/nrk/predis/releases/tag/v1.1.1) is now available.

AaronJan:
@nrk 👍 Copy that!

PR #371, opened 2016-08-16
Fallback redis role discovery for versions older than 2.8.12

nicolas-cajelli:
Predis version: last stable
Redis server version: 2.8.4
PHP version: both 5.6 and 7 (we're just migrating)

@nrk :
When trying to implement Predis on our production environment (with sentinel distribution), we noticed that it failed trying to acquire a connection. Basically, it fails trying to use an error (command not found) as an array ($actualRole[0])

The reason was  that SentinelReplication::assertConnectionRole uses the command ROLE to obtain the server role status. This command was introduced on redis 2.8.12 (http://redis.io/commands/role), and the code doesn't check the return type of executeCommand (just assumes that it's gonna return an array).

We're planning to migrate our redis servers to a newer version, but this takes time on production environment, and the fix is quite easy (just check the output of executeCommand and do a fallback to the info command).

We've been using this fork on production for a month now, without any issue (and predis works great, thanks!)

### DIFF
--- a/src/Connection/Replication/SentinelReplication.php
+++ b/src/Connection/Replication/SentinelReplication.php
@@ -20,6 +20,7 @@ use Predis\Connection\NodeConnectionInterface;
 use Predis\Connection\Parameters;
 use Predis\Replication\ReplicationStrategy;
 use Predis\Replication\RoleException;
+use Predis\Response\Error;
 use Predis\Response\ErrorInterface as ErrorResponseInterface;
 use Predis\Response\ServerException;
 
@@ -538,7 +539,15 @@ class SentinelReplication implements ReplicationInterface
     {
         $role = strtolower($role);
         $actualRole = $connection->executeCommand(RawCommand::create('ROLE'));
-
+        if (is_object($actualRole) && $actualRole instanceof Error) {
+            $info = $connection->executeCommand(RawCommand::create('INFO'));
+            preg_match('#role:([a-zA-Z]+)#m', $info, $matches);
+            if (! empty($matches)) {
+                $actualRole = [$matches[1]];
+            } else {
+                $actualRole = [''];
+            }
+        }
         if ($role !== $actualRole[0]) {
             throw new RoleException($connection, "Expected $role but got $actualRole[0] [$connection]");
         }
--- a/src/Connection/Replication/SentinelReplication.php
+++ b/src/Connection/Replication/SentinelReplication.php
@@ -543,9 +543,9 @@ class SentinelReplication implements ReplicationInterface
             $info = $connection->executeCommand(RawCommand::create('INFO'));
             preg_match('#role:([a-zA-Z]+)#m', $info, $matches);
             if (! empty($matches)) {
-                $actualRole = [$matches[1]];
+                $actualRole = array($matches[1]);
             } else {
-                $actualRole = [''];
+                $actualRole = array('');
             }
         }
         if ($role !== $actualRole[0]) {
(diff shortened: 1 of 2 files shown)


src/Connection/Replication/SentinelReplication.php
```diff
@@ -538,7 +539,15 @@ protected function assertConnectionRole(NodeConnectionInterface $connection, $ro
     {
         $role = strtolower($role);
         $actualRole = $connection->executeCommand(RawCommand::create('ROLE'));
-
+        if (is_object($actualRole) && $actualRole instanceof Error) {
```
nrk: No need for `is_object()` here, you can just check if `$actualRole` is an instance of `Predis\Response\ErrorInterface` (instead of the actual `Predis\Response\Error` class).

src/Connection/Replication/SentinelReplication.php
```diff
@@ -538,7 +539,15 @@ protected function assertConnectionRole(NodeConnectionInterface $connection, $ro
...
+            $info = $connection->executeCommand(RawCommand::create('INFO'));
+            preg_match('#role:([a-zA-Z]+)#m', $info, $matches);
```
nrk: Hmm why the multiline option for the regex? I guess `/^role:([a-zA-Z]+)/i` would be a better fit (we don't care about a case-sensitive match, and I'd prefer `/` instead of `#` for the regex pattern for the sake of consistency with the rest of the library).

nrk:
Thanks, actually I knew that `ROLE` was introduced in Redis 2.8.12 but I purposely avoided implementing the fallback to `INFO` for older versions of 2.8 unless there was some interest for it. By the way I was convinced that I indicated somewhere in the README or in the release notes that redis-sentinel was supported only when using Redis >= 2.8.12 but I can't find it so well I was wrong :sweat_smile: 

It'd be cool if this PR also provided a test case for the fallback...

nicolas-cajelli:
@nrk: Just made all the changes you asked, some notes:

- I don't recall why i made the regexp multiline, just changed it and it works
- Just added a testcase (testGetConnectionByCommandFallbacksRoleOnOldVersions), i'm not sure about the naming. Also, i don't feel right about using the at($idx) for the expects, any feedback about this would be great (i'm used to Phake for mocking, so i haven't wrote pure phpunit mocks for a few years)

Thanks for your feedback!

nicolas-cajelli:
Hi @nrk : Do you have any other thought about this PR? I would like to share this with a couple of friends who have the same issue

Thanks

src/Connection/Replication/SentinelReplication.php
```diff
@@ -538,7 +538,15 @@ protected function assertConnectionRole(NodeConnectionInterface $connection, $ro
     {
         $role = strtolower($role);
         $actualRole = $connection->executeCommand(RawCommand::create('ROLE'));
-
+        if ($actualRole instanceof ErrorResponseInterface) {
+            $info = $connection->executeCommand(RawCommand::create('INFO REPLICATION'));
```
nrk: `RawCommand::create()` expects a list of arguments, so it should be `RawCommand::create('INFO', 'REPLICATION')`.

tests/Predis/Connection/Replication/SentinelReplicationTest.php
```diff
@@ -1040,6 +1040,46 @@ public function testGetConnectionByCommandThrowsExceptionOnNodeRoleMismatch()
...
+            repl_backlog_active:0
+            repl_backlog_size:1048576
+            repl_backlog_first_byte_offset:0
+            repl_backlog_histlen:0
+        ";
+        
+        $master
+            ->expects($this->at(4))
+            ->method('executeCommand')
+            ->with($this->isRedisCommand('INFO REPLICATION'))
```
nrk: Should be `$this->isRedisCommand('INFO', ['REPLICATION'])`.

nrk:
Sorry @nicolas-cajelli, got kind of busy again. Just a couple of changes and it's good to merge :)

nicolas-cajelli:
@nrk now i'm sorry, the email about your comment got lost in my inbox and never saw it. I've been running the fork all this time without any issue. I just changed the code based on your latest feedback.

PR #379, opened 2016-09-21
Fix sentinel getClientFor cannot use index

consatan:
In version 1.1.1

``` php
$client = new Predis\Client([
    'tcp://127.0.0.1:26379',
    'tcp://127.0.0.1:26380',
    'tcp://127.0.0.1:26381',
], [
    'replication' => 'sentinel',
    'service' => 'redis_service',
]);

$slave1 = $client->getClientFor(0);
```

Throw 'InvalidArgumentException' with message 'Invalid connection ID: 0.'


nrk:
Hmm why would you want to get a connection by index in this context?

PR #381, opened 2016-10-04
replication = false also enter here and we obtain a error, parameters…

Belibov:
… are array, but warning like this is returned:

Warning: call_user_func_array() expects parameter 1 to be a valid callback, no array or string given in file /var/www/dan/rec/vendor/predis/predis/src/Client.php:176


nrk:
Fixed in ca468b7 (albeit with a different approach), will be shipped in v1.1.5. Thank you.

PR #396, opened 2016-11-24
fix array_rand

b1rdex:
`array_rand` uses the libc generator, which is slower and less-random than Mersenne Twister.

https://github.com/nrk/predis/search?utf8=%E2%9C%93&q=array_rand

You better select random array value using `mt_rand`:
```php
$array = [1, 2, 3];
$random = mt_rand(0, count($array) - 1);
```

It's fixed in php7.1 - consult https://wiki.php.net/rfc/rng_fixes -, but you're targeting outdated one.

### DIFF
--- a/src/Connection/Aggregate/MasterSlaveReplication.php
+++ b/src/Connection/Aggregate/MasterSlaveReplication.php
@@ -273,7 +273,9 @@ class MasterSlaveReplication implements ReplicationInterface
     protected function pickSlave()
     {
         if ($this->slaves) {
-            return $this->slaves[array_rand($this->slaves)];
+            $keys = array_keys($this->slaves);
+            $key = $keys[mt_rand(0, count($keys) - 1)];
+            return $this->slaves[$key];
         }
     }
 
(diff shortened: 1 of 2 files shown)


b1rdex:
for #396

b1rdex:
Sorry for so many mistakes.

b1rdex:
This doesn't make sense for v2 as it's targeting 7.2+ with core fixes to the `array_rand` applied.

nrk:
> `array_rand` uses the libc generator, which is slower and less-random than Mersenne Twister.

Using `array_rand()` on older PHP versions may be slower **internally** due to libc generator, but considering the whole picture the procedure of getting a random node out of the pool ends up being at least 4 times slower since it requires the addition of costly operations in userland code (see `array_keys()`) at least from what I've tested.

Is the distribution of random nodes picked by the client really that worse compared to using `mt_rand()`? I mean, do we have some actual measurements to prove that changing our approach for PHP <= 7.2 would be indeed better or worth the penalty in terms of runtime performances?

PR #408, opened 2017-01-12
Make examples more secure

putsi:
Some Predis users clone the repository and use it as it is. This leads to a situation where a malicious user can access the sensitive Redis storage data by accessing example files with a direct URL. This pull request tries to add some protection against that.


nrk:
I guess this makes sense, but at this point woulnd't it be just better to limit examples to be exacutable only when using the CLI? They were actually meant to be used like that after all, the output of these scripts is not really formatted to be displayed as a web page.

Issue #335, opened 2016-05-22
[2.0] Drop support for PHP 5.3


nrk:
Now that (as of writing) the release of v1.1.0 is just matter of a few days and that with this new version Predis basically supports everything needed, I'm starting to play with some changes and ideas for v2.0 that would really benefit from bumping the major required version of PHP thus **dropping support for PHP 5.3**. My plan is to **switch to PHP 5.5** as the minimum required version of PHP which is already the same requirement of big and widespread fameworks such as Symfony (3.0) and Laravel (5.1), making it a sane choice I think. This would help removing some annoying stuff from the code base here simply because of PHP 5.3 and actually improve other parts (e.g. having traits would be great with command classes).

In the end I'm **strongly** biased toward going forward and forget about PHP 5.3 but I'd really love to hear what developers using Predis think about such a major decision. Keep in mind that with 2.0 I also plan to introduce some major breaking changes in the API anyway (aside from what concerns Redis commands as they will still take the same arguments and return the same values) but at the same time there are no plans for any new major feature related to Redis (there's nothing new in this respect) so I guess this is the ideal moment for such a change, and v1.x will still serve well for those that really need to stay on PHP 5.3 or 5.4 especially considered that I plan to maintain Predis v1.x for a while for bug fixes and eventual new minor features.

Your thoughs?

GrahamCampbell:
:+1: Switching to 5.5.9 would be good. That's what symfony have done, and Laravel 5.1 LTS too. It discourages people from using really old 5.5.x versions, and 5.5.9 is what shipped with ubuntu 14.04 LTS too. 16.06 LTS as 7.0.4, or something.

GrahamCampbell:
> Symfony (3.0) and Laravel (5.1), making it a sane choice I think

Ah, you said. :)

GrahamCampbell:
Laravel 5.3 is actually PHP 5.6.4 or higher now btw. ;)

nrk:
@GrahamCampbell I considered PHP 5.6 but I guess it'd be too much of a version bump especially since I can't find any compelling reasons for it. I mean, variadic functions and arguments unpacking are indeed very nice additions to the language to have but I guess I can live without using them in Predis for the sake of a broader diffusion. Thanks!

elazar:
[...]

PR #418, opened 2017-03-07
Upgrade tests to be compatible with PHPUnit 5.7 and drop support for dead PHP versions

Milad:
I have done all the necessary changes to the tests so that they become compatible with PHPUnit 5.7 which also requires PHP 5.6+.


.gitignore
```diff
@@ -7,3 +7,4 @@ package.xml
 composer.lock
 experiments/
 vendor/
+.idea
```
GrahamCampbell: 👎 This belongs in your global ignore file.

Milad: work in progress. I have created the pull request to run the pipelines at travis!

composer.json
```diff
@@ -19,7 +19,7 @@
         "php": ">=5.3.9"
```
GrahamCampbell: This needs upgrading too.

GrahamCampbell:
👎 On these changes. They have not been made, because it's simply not needed to upgrade phpunit. It's more important to have wider PHP version support.

Milad:
I think it's not a good idea to allow dead php versions hold us back. In my project I use PHPUnit 6 and I had to downgrade to PHPUnit 5.7 just to satisfy the widest possible PHP versions here.

nrk:
@Milad I'm sorry to say this but you should have discussed these changes on a separate issue **prior** to actually investing your time in this PR. Unfortunately I can't accept it for the sole reason that in Predis 1.x we won't ever bump the minimum supported version of PHP and there's no reason to change this just for upgrading phpunit. It'll be a whole different matter for Predis 2.x as you can see in #335 but I still haven't decided which minimum target version of PHP will be supported in the end (most likely 7.x at this point, not really sure yet). That being said, I really don't see how a `require-dev` dependency in Predis can affect your whole project since phpredis is being vendored anyway by composer.

PR #421, opened 2017-03-24
Ignore replication when value it's `false|null`

jorge07:
`const VERSION = '1.1.1';`

I'm working with [SncRedisBundle](https://github.com/snc/SncRedisBundle) and I saw that `replication` parameter its always passed to the `Predis\Client` even when it's null or false.
You may think it's a bad implementation, and it is because its causing errors like:
`Warning: call_user_func_array() expects parameter 1 to be a valid callback, no array or string given`

However I think its a good idea ignore this parameters when they doesn't make sense or when the value its a non action value like `false|null` considering that a `string|callable` are the only values that require actions.

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -120,7 +120,7 @@ class Client implements ClientInterface, \IteratorAggregate
             if ($options->defined('aggregate')) {
                 $initializer = $this->getConnectionInitializerWrapper($options->aggregate);
                 $connection = $initializer($parameters, $options);
-            } elseif ($options->defined('replication')) {
+            } elseif ($options->defined('replication') && $options->replication) {
                 $replication = $options->replication;
 
                 if ($replication instanceof AggregateConnectionInterface) {
--- a/tests/Predis/ClientTest.php
+++ b/tests/Predis/ClientTest.php
@@ -297,19 +297,15 @@ class ClientTest extends PredisTestCase
     {
         $arg1 = array('tcp://host1', 'tcp://host2');
 
-        $arg2 = array(
+        $clientReplicationFalse = new Client($arg1, array(
             'cluster' => 'redis',
             'replication' =>  false,
-        );
+        ));
 
-        $clientReplicationFalse = new Client($arg1, $arg2);
-
-        $arg2 = array(
+        $clientReplicationNull = new Client($arg1, array(
             'cluster' => 'redis',
             'replication' =>  null,
-        );
-
-        $clientReplicationNull = new Client($arg1, $arg2);
+        ));
 
         $this->assertInstanceOf('Predis\Connection\Aggregate\ClusterInterface', $clientReplicationFalse->getConnection());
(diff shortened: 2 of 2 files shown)


jorge07:
I don't have idea about why the tests failed on my prev commit and now are passing. 5.4 anyway...

jorge07:
Any update on that?

nrk:
This [should actually be handled](https://github.com/nrk/predis/blob/v1.1.1/src/Configuration/ReplicationOption.php#L26-L59), but it probably doesn't work when both `replication` and `cluster` are present. Will check later.

emri99:
Facing same issue, I don't think it's handled as it happens before creating the `ReplicationOption` instance during `PRedis\Client` constructor.

Here's a stack trace:

> at ErrorHandler->handleError('2', 'call_user_func_array() expects parameter 1 to be a valid callback, no array or string given', 'vendor/predis/predis/src/Client.php', '162', array('callable' => null))
> at call_user_func_array(null, array(array(array('scheme' => 'tcp', 'host' => '127.0.0.1', 'port' => '6379', 'timeout' => '300')), object(Options))) in Client.php line 162
> at ***Client***->***Predis***\{***closure***\}(array(array('scheme' => 'tcp', 'host' => '127.0.0.1', 'port' => '6379', 'timeout' => '300')), object(Options)) in Client.php line 131
> at ***Client->createConnection***(array(array('scheme' => 'tcp', 'host' => '127.0.0.1', 'port' => '6379', 'timeout' => '300'))) in Client.php line 56
> at ***Client->__construct***(array(array('scheme' => 'tcp', 'host' => '127.0.0.1', 'port' => '6379', 'timeout' => '300')), array('prefix' => 'bunny:', ***'replication' => null***, 'service' => 'session', 'parameters' => array('database' => '1', 'password' => null)))

jorge07:
Any updates here?

nrk:
Fixed in ca468b7 for v1.1.5... or better said, now the invalid value is handled without raising a PHP warning but throwing an exception with a message that informs users that passing `false` or any value that can be interpreted as `false` to this option is not supported.

I say _invalid value_ because using `false` does not make sense anyway, if we add this for `replication` then it should be done for `cluster` too (but it doesn't work). Furthermore, setting `replication => false` with this change would simply fall back to aggregating nodes in a cluster setup even when `cluster` is not being explicitly set, this would be confusing at best.

By the way the fact that `replication` in v1.1 accepts `true` is only due to backwards compatibility with older versions of Predis when it only had one backend for replication (Redis Sentinel came later) and we needed whatever value for this option simply because options are passed as a dictionary... `false` was never intended to be an accepted value. Admittedly, choosing `true` as a "trigger" value was not a smart choice in the end, more in general client options for aggregate connections turned into a bit of a mess in v1.1 as things did not scale well with the addition of newer features. In v2.0 they will behave in a completely different (and more consistent) way.

jorge07:
Nice. Thanks for the effort @nrk

PR #422, opened 2017-04-03
Stop treating empty strings as valid parameters

shadowhand:
Ideally, these parameters would always be `null` but may end up as empty strings `''` and then invalid `AUTH` commands will be sent.

### DIFF
--- a/src/Connection/Factory.php
+++ b/src/Connection/Factory.php
@@ -173,13 +173,13 @@ class Factory implements FactoryInterface
     {
         $parameters = $connection->getParameters();
 
-        if (isset($parameters->password)) {
+        if (!empty($parameters->password)) {
             $connection->addConnectCommand(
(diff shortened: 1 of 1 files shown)


nrk:
Using `empty()` is not good as it would break things when explicitly trying to connect to the database index `0` (after all `empty('0') == true`... yay PHP!), same thing for `password`, even though having `0` as a password would be kind of silly. I not inclined to put some convoluted checks in place for that honestly.

shadowhand:
Well right now you have a situation where defining `'password' => ''` will send an `AUTH` command that is invalid. Perhaps an auto-cast boolean check instead?

```php
if (isset($parameters->password) && $parameters->password)) { ... }
```

nrk:
Fixed for v1.1 in 31baa9b (will be included in v1.1.4).

PR #436, opened 2017-05-03
fix null password config

star1989:
fix null password config


nrk:
Fixed for v1.1 in 31baa9b.

PR #441, opened 2017-05-25
Disconnect sentinelConnection on SentinelReplication::disconnect()

bartbutler:
Without this the sentinel connection is left open which can cause weird effects if the script is forked, etc.

### DIFF
--- a/src/Connection/Aggregate/SentinelReplication.php
+++ b/src/Connection/Aggregate/SentinelReplication.php
@@ -640,6 +640,10 @@ class SentinelReplication implements ReplicationInterface
         foreach ($this->slaves as $connection) {
             $connection->disconnect();
         }
+
+        if ($this->sentinelConnection) {
+            $this->sentinelConnection->disconnect();
+        }
     }
 
     /**


nrk:
Makes sense but some tests now fail, most likely it's only due to the changed behaviour so they just need to be tweaked. On a related note I don't think the explicit `connect()` to the sentinel node is really needed, this action is handled lazily anyway.

bartbutler:
Funny about the tests, nothing should really have changed there in terms of the outward behavior of the class. I'm not sure the explicit connect is needed, but on the other hand it is present in the switchTo function for the master/slaves and the executeCommand function does not seem to check for connection first, just write to the resource, which if disconnected doesn't exist. I figured adding the explicit connect() was safer.

bartbutler:
Aha, yes getResource() in AbstractConnection performs a connect() if the resource does not exist. This might be cleared though.

bartbutler:
Okay, the tests work when I roll back the last commit with the 'connect'. I think the issue is that if the connect happens there and fails, it isn't inside a try-catch block, so the next sentinel does not get tried. In any case, should work now.

nrk:
Yes that's why I said that it's handled lazily 😄 See the README:

> Connections to Redis are lazy meaning that the client connects to a server only if and when needed. While it is recommended to let the client do its own stuff under the hood, there may be times when it is still desired to have control of when the connection is opened or closed: this can easily be achieved by invoking $client->connect() and $client->disconnect(). Please note that the effect of these methods on aggregate connections may differ depending on each specific implementation.

This is not written in the phpdocs of `Predis\Connection\NodeConnectionInterface::getResource()` simply because it may depend on the actual implementation (see `Predis\Connection\WebdisConnection`) but I guess it's worth making it more clear.

As for `Predis\Connection\Aggregate\SentinelReplication::switchTo()`, we explicitly connect to the selected node because it's useful to know ASAP if that server is actually alive (it's an operation that overrides the whole redis-sentinel automatisms after all) but for sentinel nodes this is not really a problem, Predis handles just handles a failure transparently.

Tests failed with your second commit because some specific test units (usually those marked with the `@group disconnected`) rely on the assumption that an actual connection is not opened.

bartbutler:
Aha, good to know!

PR #443, opened 2017-06-09
Fix travis — use trusty dist

b1rdex:
Or drop hhvm support as Symfony and others — https://github.com/symfony/symfony/issues/18922

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,5 @@
 language: php
 sudo: false
-dist: trusty
-php:
-  - 5.3
-  - 5.4
-  - 5.5
-  - 5.6
-  - 7.0
-  - 7.1
-  - hhvm
 branches:
   except:
     - v0.5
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,5 +1,6 @@
 language: php
 sudo: false
+dist: trusty
```
GrahamCampbell: better to just go trusty on the hhvm build only, to avoid getting really old php versions.

b1rdex: How you supposed to do it?

GrahamCampbell:
```yaml
matrix:
  include:
    - php: 5.3
    - php: 5.4
    - php: 5.5
    - php: 5.6
    - php: 7.0
    - php: 7.1
    - php: hhvm
      dist: trusty
```

b1rdex:
Thank you, @GrahamCampbell. Changed matrix, but hhvm have one test failed.

PR #448, opened 2017-07-07
Fix each

chihiro-adachi:
When running in the environment of php 7.2, the following warning was issued.
`The each() function is deprecated. This message will be suppressed on further calls: 1x`

`each()` function seems to be deprecated in 7.2.
http://php.net/manual/en/function.each.php

I tried to fix it, please merge if necessary.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,6 @@ php:
   - 5.6
   - 7.0
   - 7.1
-  - hhvm
 branches:
   except:
     - v0.5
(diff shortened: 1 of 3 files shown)


nrk:
In the end I fixed it by simply using `key()` and `current()` (see 157f658), painless fix as there was no real need for using `each()` to begin with. Thank you anyway @chihiro-adachi for reporting this! By the way, the next time please do not mix commits with different scopes in a PR (that is, the fix that removed `each()` and the change targeting TravisCI), two separate PRs even if tiny are way better.

chihiro-adachi:
thank you 😄

PR #450, opened 2017-07-12
Update CRC16.php

tongsq:
Hello, I think i have meet a bug when use redis cluster. 
if the redis key is a number, this function always return 0. it will cause an error when use redis cluster. i think this is a bug need repair


nrk:
Thanks, the bug was indeed there but the fix was not optimal as it would be extremely slow. An explicit cast to string of the input value is enough and adds almost no overhead. Fixed in 343afd3.

Issue #447, opened 2017-07-03
Replication + pipeline = always on master ?


eymerich:
Hi, seems like it is not possible to execute a pipeline of read only commands to a slave, everything in this configuration is routed through the master, is there a reason?
src/Pipeline/Pipeline.php
if block @ line 114

We have worked on a workaround that evaluates each command of the impending pipeline and if it contains all read only operation it is routed through a suitable slave.

nrk:
The reason is basically to keep things simple. At the time of implementation I thought of doing something similar (at least conceptually) to your workaround but in the end I decided to stick with the current behavior. In theory nothing stops you from using `Predis\Client::getClientFor()` to get a client instance for a specific slave and then execute the pipeline against it.

Anyway I'm not against the idea per se, depending on the implementation.

nrk:
I'm closing this as it won't be implemented for v1.1, using your own custom pipeline is easy enough like demonstrated by your PR so such cases can be easily covered..

That said in v2.0 we're going to use a completely different approach for pipelines and this behaviour is going to be supported, developers will actually be able to customize the underlying behaviour of a pipeline to a greater extent and without messing with the _front-facing_ pipeline abstraction.

PR #451, opened 2017-07-12
Target replica nodes when pipelines contain only read operations

eymerich:
About #447


Lusitaniae:
Any update here?

nrk:
This will be implemented in v2.0.

In v2.0 we're going to use a completely different approach for pipelines, developers will actually be able to customize the underlying behaviour of a pipeline to a greater extent and without messing with the front-facing pipeline abstraction.

PR #456, opened 2017-08-14
Add null return value to PHPDoc

teohhanhui:


nrk:

PR #464, opened 2017-09-24
modify zadd hint so they indicate infinite number of arguments

janmyszkier:
following example of:
http://manual.phpdoc.org/HTMLSmartyConverter/HandS/phpDocumentor/tutorial_tags.param.pkg.html

### DIFF
--- a/src/ClientInterface.php
+++ b/src/ClientInterface.php
@@ -113,7 +113,7 @@ use Predis\Profile\ProfileInterface;
  * @method array  sscan($key, $cursor, array $options = null)
  * @method array  sunion(array|string $keys)
  * @method int    sunionstore($destination, array|string $keys)
- * @method int    zadd($key, array $membersAndScoresDictionary)
+ * @method int    zadd($key, ...$membersAndScoresDictionary)
  * @method int    zcard($key)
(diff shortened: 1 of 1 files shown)


janmyszkier:
This is because the second argument is not an array, in fact it requires you to give at least 3 params:
```
$this->redis->zAdd($setKey,$score,$data)
```

nrk:
Actually for @method annotations we chose to stick with a more PHP-friendly approach for command arguments instead of using the underlying order of arguments as defined by the Redis documentation (provided that Predis always supports the latter for commands). In our case `ZADD key score1 member1 [scoreN memberN]` translates to `$client->zadd($key, ["member1" => "score1", "memberN" => "scoreN"])` that's why we have `array` in the method signature (unfortunately it's not possible with phpdoc annotations to specify that we really expect a named array).

On of the reasons for chosing this approach is also related to the fact that Redis commands historically changed by adding more and more variants and options, in fact since Redis 3.0.2 `ZADD` was expanded to accept [some modifiers](https://redis.io/commands/zadd#zadd-options-redis-302-or-greater) so the proposed approach would be pretty much impossible to document, instead in our case we can just add a third optional argument of type `array` to specify optional modifiers.

PR #467, opened 2017-10-12
Fixed stream read blocking signals

imcj:
this code demo blpop will block pcntl_signal

```
<?php
declare(ticks=1);
# I want ctrl + c terminal this process
# but that unavailable because int signal is blocking by fgets
include __DIR__ . '/../autoload.php';
pcntl_signal(SIGINT, function($signal) {
    echo("INT signal\n");
    exit(0);
});
$client = new \Predis\Client();
$client->blpop('test', 10);
```

https://github.com/imcj/predis/blob/demo-blocking/examples/stream_connection_blocking_signals.php

I fixed that. Add select in read, select not blocking signals.when signal arrived , select will raise an exception.and then dispatch signals again.

now, i can listen SIGINT, save process state. and quit.

why did that?

because i use predis in the imcj/mqk , that library start multi processes consume redis list message, and mapping to the event.i want use signal SIGINT graceful all processes, but predis hold all signals, i cann't listen SIGINT signal.

thx for all contributors.

```
    public function read()
    {
        $socket = $this->getResource();
        while (true) {
            try {
                $this->select([$socket], [], []);
            } catch (\Exception $e) {
                if ($e == 'Interrupted system call') {
                    pcntl_signal_dispatch();
                    continue;
                }
            }
            $chunk = fgets($socket);
            break;
        }
```

### DIFF
--- a/src/Connection/StreamConnection.php
+++ b/src/Connection/StreamConnection.php
@@ -312,7 +312,7 @@ class StreamConnection extends AbstractConnection
             try {
                 $this->select([$socket], [], []);
             } catch (\Exception $e) {
-                if ($e == 'Interrupted system call') {
+                if ($e->getMessage() == 'Interrupted system call') {
                     pcntl_signal_dispatch();
                     continue;
                 }
@@ -401,7 +401,7 @@ class StreamConnection extends AbstractConnection
                 return;
             }
             // raise all other issues to exceptions
-            throw new \Exception($errstr, 0, $errno, $errfile, $errline);
+            throw new \Exception($errstr, $errno);
         });
 
         $count = stream_select($r, $w, $e, 10.0);
--- a/src/Connection/StreamConnection.php
+++ b/src/Connection/StreamConnection.php
@@ -1,5 +1,5 @@
 <?php
-
+declare(ticks=1);
 /*
  * This file is part of the Predis package.
  *
@@ -308,7 +308,18 @@ class StreamConnection extends AbstractConnection
     public function read()
     {
         $socket = $this->getResource();
-        $chunk = fgets($socket);
+        while (true) {
+            try {
+                $this->select([$socket], [], []);
+            } catch (\Exception $e) {
+                if ($e == 'Interrupted system call') {
+                    pcntl_signal_dispatch();
+                    continue;
+                }
+            }
+            $chunk = fgets($socket);
+            break;
+        }
 
         if ($chunk === false || $chunk === '') {
             $this->onConnectionError('Error while reading line from the server.');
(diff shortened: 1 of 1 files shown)


Majkl578:
Hi, I'm currently investigating possibilities of non-blocking reads, since I want a long-running SUBSCRIBE-based script, which should also be able to respond to signals properly. Current implementation blocks in a hard way and there is no way around.

src/Connection/StreamConnection.php
```diff
@@ -1,5 +1,5 @@
 <?php
-
+declare(ticks=1);
```
Majkl578: Ticks have huge overhead and generally are not a good idea. May I suggest to check Async signals implemented in PHP 7.2?
https://wiki.php.net/rfc/async_signals

src/Connection/StreamConnection.php
```diff
@@ -308,7 +308,18 @@ protected function write($buffer)
     public function read()
     {
         $socket = $this->getResource();
-        $chunk = fgets($socket);
+        while (true) {
+            try {
+                $this->select([$socket], [], []);
```
Majkl578: Is there any reason why this isn't implemented using non-blocking streams instead? http://php.net/manual/en/function.stream-set-blocking.php

imcj: predis cannot  using non-blocking, because that socket io read must in loop.if you want real non-blocking, please google php redis react-php.

non-blocking need io event loop.

Majkl578: Imho you can use non-blocking stream even without event loop. You just need to loop synchronously, just like you did here with `while( true)`, and check `fgets(...) !== false`. The point is that in such synchronous loop, signals are dispatched properly, unlike blocking stream.

src/Connection/StreamConnection.php
```diff
@@ -308,7 +308,18 @@ protected function write($buffer)
...
+            } catch (\Exception $e) {
+                if ($e->getMessage() == 'Interrupted system call') {
+                    pcntl_signal_dispatch();
```
imcj: @Majkl578 look at here.if signal deliveryed, select will throw an exception, pcntl_signal_dispatch.

imcj:
sorry,my english so bad.need some times to think what's do you say.maybe
few hours, becuase my country is am 11:28.

Michael Moravec <notifications@github.com>于2018年3月5日周一 上午10:25写道：

nrk:
@Majkl578 it's definitely true that in the last 12 months I didn't have much time to work on my open-source projects but this library is definitely not abandoned... _on hiatus_ would probably be a better definition of its status, after all it doesn't have major issues (I can't reproduce most of those currently opened TBH) and there are a few improvements for the current stable branch that would be nice to have but not urgent. It doesn't explicitly come with a profile for Redis 4.x and this is probably the most striking issue right now, but it's not like it won't work and devs can still inject their own implementation of commands if they really need them. 

That said I'd like to work on it again very soon but since I don't have a good track record on my promises in this regard I won't dare to give a timeframe. By the way good contributions would be a great help, unfortunately in the past, aside from very few exceptions, I didn't really get much help beyond minor fixes which are greatly appreciated anyway, but they sstill don't change the fact that this basically remains a one-man project.

As for support for old PHP releases, the fact that it still maintains backwards compatibility with PHP 5.x is quite a given for versions in the 1.x branch... I mean there's no real need to shake-up the internals of this library just for the fun of it when it's stable and such, these new features are for the most part not really required for Predis to work fine. Obviously it's a completely different matter for the 2.x branch: [as I've stated before](https://github.com/nrk/predis/issues/335#issuecomment-318052071) it's pretty much decided that Predis 2.x will drop support for PHP 5.x exactly because v1.x, with minor fixes and maintenance, will still do fine for past versions of PHP for many years to come.

I hope this clarifies a bit the current state of Predis.

src/Connection/StreamConnection.php
nrk: Indeed an event loop is most likely not really needed in such cases (see the socket-based phpiredis connection backend in which [we used `socket_select()`](https://github.com/nrk/predis/blob/v1.1.1/src/Connection/PhpiredisSocketConnection.php#L310) to provide the same timeout behaviour of `stream_socket_client()`, Predis remains mainly blocking so it doesn't need to coordinate with other I/O events or libraries that plugs into a full fledged event loop.

Issue #280, opened 2015-09-03
Question: predis multi slaves with a slave down


endymon:
My question is, with a predis setup involving multiple configured slaves, If 1 or more slaves are down, how do you get predis to stop trying to send queries to the down slave(s)?

When I  tested taking a slave down I got spammed by these:
```
Predis\\Connection\\ConnectionException: from 'EXISTS XXXXXX{THE_KEY}XXXXX' while connected to host: 'XXX{THE_HOST}XXX' MESSAGE: Connection refused [tcp://localhost:XX{PORT}XX]
```
And then after a short while the error message switched to:
```
'LOADING Redis is loading the dataset in memory'
```

These exceptions make sense if there is only one slave, but I would have assumed that in a multiple slave configuration predis either would by default (or by configuration parameter) fail-over silently to a NON down slave host. ONLY if all slaves were down could I see an exception propagating up to the client, but not in this case.

How does predis handle multiple slaves when 1 or more are failed or unavailable?
Is there such a configuration parameter?

Thanks
-- Matt

xiconfjs:
Right now we are testing the replication-sentinel function of the predis 1.1-sentinal branch. And we are experiencing the same phenomenon. Is there an option to skip a slave with is re-syncing itself?

nrk:
@endymon starting with Predis v1.1, trying to execute a read-only command on a unreachable slave will discard that slave and try again on the next one, or even the master if all of the slaves are offline ([see the related changes](https://github.com/nrk/predis/blob/c936621/src/Connection/Aggregate/MasterSlaveReplication.php#L279-L340)). As for errors such as `-LOADING` they are not handled yet but I guess it should be fairly easy to intercept them to apply the same logic described above.

@xiconfjs the same applies to redis-sentinel, I will look into it.

nrk:
Now Predis intercepts `-LOADING` responses, removes the "busy" slave from the pool and executes the command on the next slave (or master as last resort) instead of throwing an exception right away.

PR #475, opened 2017-12-08
Prevent early failure when a slave is loading dataset.

esler:
Same problem like #280 

I'm getting this exception
```
Exception: LOADING Redis is loading the dataset in memory
  Type: Predis\Response\ServerException
  File: vendor/predis/predis/src/Client.php
  Line: 370
```

I'm using `Predis\Connection\Aggregate\SentinelReplication` which is missing previous fix. I've prepared simple fix in the fashion of previous fix. I'm not sure about not removing failing slave from the pool but it works for my problem.


chasebolt:
any chance of this getting merged in? we are experiencing this issue with using sentinel.

Issue #488, opened 2018-02-19
exists command breaks with tr_TR.utf8 locale


coffeerocket:
The library uses the method strtoupper to capitalize the command.
In the case of the Turkish locale, the "i" is not capitalized, cause it is a special character in Turkish.
The command cannot be found which leads to an error.

nrk:
I'm closing this issue as this is a duplicate of #372.

Issue #372, opened 2016-08-18
Command 'iNCRBY' is not a registered Redis command error


deepmk:
strtoupper function and different character sets in the problem PREDIS package

sample
`setlocale(LC_ALL, 'tr_TR.utf8', 'tr_TR.utf8', 'tr_TR');`

used in a project that brings problems 

sample
`if (isset($this->commands[$commandID = strtoupper($commandID)])) {
            return $this->commands[$commandID];
        }`

INCRBY = iNCRBY

I solved the problem by using mb_strtoupper instead strtoupper.

nrk:
Apparently the Turkish language just decided to make things weird here! Jokes aside, I would've never guessed such an issue when uppercasing strings with plain latin characters but I've [just read about it](http://blog.yjl.im/2011/08/i-and-i-in-turkish.html) here and well... you learn something new every day :smile: Not sure how to fix this right now but unfortunately `mbstring` is a non-default extension so we can't really rely on it.

baskan:
@widlturk  fix is actually just replace ReplaceProfile.php line 85 to:
>         $commandID = mb_strtoupper($commandID, 'UTF-8');

@nrk is there a way that we can use our own Profiler implementing ProfileInterface ?

egurelli:
So same issue happened to me also when I using Laravel (5.3) Cache. I have investigated the issue and codes both Laravel lib and Predis lib.

Looks like Predis proxy commands to Redis. From Laravel (5.3) documents :

> You may interact with Redis by calling various methods on the Redis facade. The Redis facade supports dynamic methods, meaning you may call any Redis command on the facade and the command will be passed directly to Redis.

So my solution on Laravel (probably will be same on other framework platforms or native users) :

use commands like in [Redis documents](http://redis.io/commands/) with "**case sensitive**"

`Redis::incrby`
Result : Command 'iNCRBY' is not a registered Redis command.

`Redis::INCRBY`
Result : Success

Note : Packages/Vendors using "lowercase" Redis commands and that cause same problem : Probably extending Class and redefine functions/methods may solve the issue.

nrk:
[...]

PR #489, opened 2018-02-19
Fix and test case for issue #488 (command capitalization error under turkish locale)

coffeerocket:


nrk:
I'm closing this PR as it doesn't actually provide a comprehensive fix. Also read [my comment here](https://github.com/predis/predis/pull/566#issuecomment-674395781) on #566 and the first issue created about this on #372.

PR #494, opened 2018-03-05
Fixup readme badge with badge poser!

JellyBellyDev:
before:
schermata 2018-03-05 alle 15 40 08

after: 
schermata 2018-03-05 alle 15 48 08

### DIFF
--- a/README.md
+++ b/README.md
@@ -476,10 +476,10 @@ found [on its project page](http://travis-ci.org/nrk/predis).
 
 The code for Predis is distributed under the terms of the MIT license (see [LICENSE](LICENSE)).
 
-[ico-license]: https://img.shields.io/github/license/nrk/predis.svg?style=flat-square
-[ico-version-stable]: https://img.shields.io/packagist/v/predis/predis.svg?style=flat-square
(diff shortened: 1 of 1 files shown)


nrk:
shields.io works perfectly fine overall, most likely it was simply a temporary hiccup on their part so I think switching just because of this is kind of overkill TBH.

JellyBellyDev:
I love [Badge Poser](https://poser.pugx.org/), also because it is written in php!
But if you do not want to change, no problem. Close the PR too!
I tried it!

drAlberT:
I had a lot of such issues in the past too with shields.io ...  We could try to change :)

nrk:
@drAlberT the fact is, it's not I have anything against Badge Poser and in the past I was actually [using it in the README](https://github.com/nrk/predis/commit/257791edcfcfee6db3fafb89671d87fb4c9f055b#diff-04c6e90faac2675aa89e2176d2eec7d8), but I switched to **shields.io** because it supports Travis CI and Gitter (and well it even provides an HHVM badge) and I vastly prefer their flat style. Not sure if anything changed about that for Badge Poser...

tillkruss:
Seems to work 👍🏻

PR #497, opened 2018-04-11
Correcting the return type hint for BRPOPLPUSH method

phpguru:
I could be missing something, but in my tests, `brpoplpush` returns a string. [The Redis docs](https://redis.io/commands/brpoplpush) seem to confirm this. My IDE was complaining about the return type not matching in a method I wrote that returns a string after calling `brpoplpush`, and when I dug into the IDE warning, I found the `ClientInterface` class to be providing an `array` type instead of a `string` return type.


nrk:

tillkruss:
@nrk: How should we handle these PRs in terms of pushing them into `2.x` as well?

PR #499, opened 2018-04-12
Disconnect when server returns bad response in an atomic pipeline

jimbonator:
Predis' atomic pipelines are not throwing a ServerException when Redis
returns an error.  Instead, the code considers it to be a ClientException
and maintains the connection.

This patch will convert server errors to ServerExceptions and disconnect.
The Client object will re-connect on the next command attempt.

This patch also disconnects from the server if no response or the partial
response does not include a server error, as the connection state appears
to be in a bad state and should not be reused.

### DIFF
--- a/src/Pipeline/Atomic.php
+++ b/src/Pipeline/Atomic.php
@@ -18,6 +18,8 @@ use Predis\Connection\NodeConnectionInterface;
 use Predis\Response\ErrorInterface as ErrorResponseInterface;
 use Predis\Response\ResponseInterface;
 use Predis\Response\ServerException;
+use Predis\Transaction\AbortedMultiExecException;
+use Predis\Transaction\MultiExec;
 
 /**
  * Command pipeline wrapped into a MULTI / EXEC transaction.
@@ -62,70 +64,82 @@ class Atomic extends Pipeline
     protected function executePipeline(ConnectionInterface $connection, \SplQueue $commands)
     {
         $profile = $this->getClient()->getProfile();
-        $connection->executeCommand($profile->createCommand('multi'));
+        $multi = $profile->createCommand('MULTI');
+        $exec = $profile->createCommand('EXEC');
 
+        // open transaction: MULTI
+        $connection->writeRequest($multi);
+
+        // transmit all commands in transaction
         foreach ($commands as $command) {
             $connection->writeRequest($command);
         }
 
+        // execute transaction: EXEC
+        $connection->writeRequest($exec);
+
+        // response to MULTI
+        $response = $connection->readResponse($multi);
+        if ($response instanceof ErrorResponseInterface) {
+            $this->exception($connection, $response); // close connection and throw ServerException
+        }
+
+        // responses to commands (all should be QUEUED result)
+        $ex = null;
         foreach ($commands as $command) {
             $response = $connection->readResponse($command);
 
-            if ($response instanceof ErrorResponseInterface) {
-                $connection->executeCommand($profile->createCommand('discard'));
-                throw new ServerException($response->getMessage());
+            // throw first error as a ServerException
+            if (($response instanceof ErrorResponseInterface) && !$ex) {
+                $ex = new ServerException($response->getMessage());
             }
         }
 
-        $executed = $connection->executeCommand($profile->createCommand('exec'));
+        // response to EXEC
+        $responses = $connection->readResponse($exec);
 
-        if (!isset($executed)) {
-            // consider connection fouled
-            $connection->disconnect();
-
-            // TODO: should be throwing a more appropriate exception.
-            throw new ClientException(
-                'The underlying transaction has been aborted by the server.'
-            );
+        // if one of the commands returned an error, only throw after all responses have been received
+        if ($ex) {
+            throw $ex;
         }
 
-        if (count($executed) !== count($commands)) {
-            // convert first returned error to ServerException
-            if ($this->throwServerExceptions()) {
-                foreach ($executed as $response) {
-                    if ($response instanceof ErrorResponseInterface)
-                        $this->exception($connection, $response);
-                }
-            }
+        // if EXEC returned null, the transaction was aborted
+        if (!isset($responses)) {
+            // fake a MultiExec, which is not used in an atomic pipeline
+            $multi_exec = new MultiExec($this->getClient());
+            throw new AbortedMultiExecException($multi_exec, 'The underlying transaction has been aborted by the server.');
+        }
 
-            // consider connection fouled
-            $connection->disconnect();
+        // if EXEC returned an error, throw ServerException (but don't close connection, all responses
+        // have been read and are accounted for)
+        if ($responses instanceof ErrorResponseInterface) {
+            throw new ServerException($responses->getMessage());
+        }
 
+        if (count($responses) !== count($commands)) {
             $expected = count($commands);
-            $received = count($executed);
+            $received = count($responses);
 
             throw new ClientException(
                 "Invalid number of responses [expected $expected, received $received]."
             );
         }
 
-        $responses = array();
-        $sizeOfPipe = count($commands);
         $exceptions = $this->throwServerExceptions();
 
-        for ($i = 0; $i < $sizeOfPipe; ++$i) {
-            $command = $commands->dequeue();
-            $response = $executed[$i];
+        // parse all unparsed results
+        $i = 0;
+        foreach ($commands as $command) {
+            $response = $responses[$i];
 
             if (!$response instanceof ResponseInterface) {
-                $responses[] = $command->parseResponse($response);
+                $responses[$i] = $command->parseResponse($response);
             } elseif ($response instanceof ErrorResponseInterface && $exceptions) {
                 $this->exception($connection, $response);
-            } else {
-                $responses[] = $response;
             }
 
-            unset($executed[$i]);
+
+            $i++;
         }
 
         return $responses;
--- a/src/Pipeline/Atomic.php
+++ b/src/Pipeline/Atomic.php
@@ -80,6 +80,9 @@ class Atomic extends Pipeline
         $executed = $connection->executeCommand($profile->createCommand('exec'));
 
         if (!isset($executed)) {
+            // consider connection fouled
+            $connection->disconnect();
+
             // TODO: should be throwing a more appropriate exception.
             throw new ClientException(
                 'The underlying transaction has been aborted by the server.'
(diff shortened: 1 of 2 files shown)


jimbonator:
I should add, this patch emerged from problems in production (a Redis cluster under constant load).  This patch appears to have resolved our problems entirely.

jimbonator:
> There was 1 failure:
> 1) Predis\Command\ConnectionSelectTest::testThrowsExceptionOnUnexpectedDatabase
> Failed asserting that exception message 'ERR DB index is out of range' contains 'ERR invalid DB index'.

I don't think my patch caused these errors.  When I test on my dev machine I see the same error with the `v1.1` branch.

jimbonator:
I've rebased this branch against `v1.1`.  I've also gone further with Atomic to solve some of its error-handling issues (where the socket is not flushed when an error occurs during the transaction, similar to #547) and fixed the code so an aborted transaction properly throws an `AbortedMultiExecException`.  This implementation is also more performant because all commands (incl. `MULTI` and `EXEC`) are pipelined.

src/Pipeline/Atomic.php
```diff
@@ -62,56 +64,82 @@ protected function getConnection()
...
                 "Invalid number of responses [expected $expected, received $received]."
             );
         }
 
-        $responses = array();
-        $sizeOfPipe = count($commands);
         $exceptions = $this->throwServerExceptions();
 
-        for ($i = 0; $i < $sizeOfPipe; ++$i) {
-            $command = $commands->dequeue();
```
nrk, line 103: Is it really necessary to change this?

This was meant to free memory as soon as possible by discarding useless instances of `Predis\Command\Command` stored in `$commands`: basically when a `$response` is parsed by the associated `$command` we don't need to keep `$command` in memory, so doing a `SplQueue::dequeue()` removes the only reference for that instance that can be reclaimed by the GC at the next iteration). When you have some hefty pipelines this actually makes a difference in terms of memory usage.

Even by going with your change, you are still not emptying the `SplQueue` (which is kept in reference by the underlying implementation of `Predis\Pipeline\Pipeline`). Truth to be told `SplQueue` should probably be emptyed even when exception occurs but this is not done even in my original code, this should be investigated but it's not the scope of this PR as it also involves the base class.

src/Pipeline/Atomic.php
```diff
@@ -62,56 +64,82 @@ protected function getConnection()
...
             if (!$response instanceof ResponseInterface) {
-                $responses[] = $command->parseResponse($response);
+                $responses[$i] = $command->parseResponse($response);
             } elseif ($response instanceof ErrorResponseInterface && $exceptions) {
                 $this->exception($connection, $response);
-            } else {
-                $responses[] = $response;
             }
 
-            unset($executed[$i]);
```
nrk, line 114: Same as above: the reason for this is to free as much memory as possible early on during the iteration.

src/Pipeline/Atomic.php
```diff
@@ -62,56 +64,82 @@ protected function getConnection()
...
+        if ($ex) {
+            throw $ex;
         }
 
-        if (count($executed) !== count($commands)) {
+        // if EXEC returned null, the transaction was aborted
+        if (!isset($responses)) {
+            // fake a MultiExec, which is not used in an atomic pipeline
+            $multi_exec = new MultiExec($this->getClient());
+            throw new AbortedMultiExecException($multi_exec, 'The underlying transaction has been aborted by the server.');
```
nrk, line 110: I do understand the logic behind this as a specific exception would be ideal in this context, but I don't like it anyway as you are throwing an exception which expects a certain type as an argument and you work around this by instantiating a fake object (with an internal state that potentially differs from the current state of the pipelined transaction) that has no meaning when intercepted by outside code in a try / catch block.

Would it be possible to go with a generic `Predis\ClientException` for now in v1.1?

In v2.0 we may consider having `Predis\Transaction\AbortedMultiExecException` accept a `Predis\ClientContextInterface` as an argument for `__construct()` (pipelines and plain transactions implement this interface after all) and we could even add a specific interface like `Predis\Transaction\TransactionInterface` if we can find some specific common ground that could work with both pipelined and non-pipelined transactions, or eventually just make it a marker interface that implements `Predis\ClientContextInterface`.

src/Pipeline/Atomic.php
```diff
@@ -62,56 +64,82 @@ protected function getConnection()
...
 
-        if (!isset($executed)) {
-            // TODO: should be throwing a more appropriate exception.
-            throw new ClientException(
-                'The underlying transaction has been aborted by the server.'
-            );
+        // if one of the commands returned an error, only throw after all responses have been received
+        if ($ex) {
+            throw $ex;
         }
```
nrk, line 104: I'm most likely missing something here but how is this any different than using [`DISCARD`](https://redis.io/commands/discard) (it tells Redis to discard the queued commands and reset the server-side state without having the client to drop the connection) and throw the exception as soon as we get a non `+QUEUED` response from the server, like did in [previous code](https://github.com/predis/predis/pull/499/files#diff-018d40c6b8462f1b9a5e89720186e531L74-L76)?

And if we use `DISCARD` to clear the transaction there's no actual reason to spend time and resources reading back the responses from the server just to prevent protocol desynchonization issues (just to prove it, if you do a `DISCARD` and then try to do an `EXEC` you'll get a `-ERR EXEC without MULTI`).

nrk: Alright nevermind, I was indeed missing the fact that we are pipelining here and that's exactly the reason why we had a bug in the first place 🤦‍♂️ I'll leave my original comment above as a reference anyway.

In this context, using [`DISCARD`](https://redis.io/commands/discard) after pipelining all of the writes would be pointless anyway as we have already sent all of the commands to Redis anyway and are just reading back status responses.

That said,  I still think that as soon as we read back an error instead of a `+QUEUED` response it'd be probably more efficient to drop the connection, clear the `SplQueue` object and then throw the exception immediatly instead of consuming all of the outstanding reads. Wouldn't it work the same to avoid protocol desynchronization issues?

src/Pipeline/Atomic.php
jimbonator: I wasn't aware emptying the queue was a design goal.  I'll update the branch with that in mind.

src/Pipeline/Atomic.php
jimbonator: (as with above)

src/Pipeline/Atomic.php
jimbonator: Trust me, I don't like it either, but this is pretty important to me.  We have systems relying on `WATCH/MULTI/EXEC` to perform certain operations, and we keep stats on which operations failed due to errors (which is a problem to be fixed) versus those which failed due to concurrency issues (which is part of the design).  This change permitted us to (finally) distinguish between the two cases without performing string comparisons of exception messages.  (It also meant we could migrate plain `transaction()` code to atomic pipelines without further code changes.)

If making this change is that troublesome, I'll revert.  I filed a ticket for this at #640.

nrk: The description of your use case makes perfect sense so we'll fix #640 for sure in v2.0, thanks for creating a separate issue so that we can keep track of it.

Back to this on the stable branch, just thinking out loud... if we really want to change I'd probably prefer having a new exception extending `Predis\Transaction\AbortedMultiExecException` and implemented with something along these lines (I'm skipping phpdocs just for the sake of brevity):

```php
namespace Predis\Pipeline;

use Predis\Transaction\AbortedMultiExecException;

class AbortedAtomicPipelineException extends AbortedMultiExecException
{
    public function __construct($message, $code = null)
    {
        parent::__construct(null, $message, $code);
    }

    public function getTransaction()
    {
        return null;
    }
}
```

This would make transactions and atomic pipelines somewhat interchangeable when aborted transaction errors are returned by Redis as you'd just need to catch `Predis\Transaction\AbortedMultiExecException`. Furthermore, I'd vastly prefer `getTransaction()` to return `NULL` instead of an unrelated instance of `Predis\Transaction\MultiExecTransaction`. Then again, my concern is that there would be no way to retrieve the original instance of `Predis\Client` in a `catch` block. Hmmmm....

PS: generally speaking, this can be indeed considered a bug but would such a change be considered backwards-compatible for a patch release?

src/Pipeline/Atomic.php
nrk: > That said, and unless I'm wrong once again, I still think that as soon as we read back an error in place of a +QUEUED response it'd be probably more efficient to explicitly drop the connection (which also clears the pending transaction on the Redis side), clear the SplQueue object and then throw the exception immediatly instead of consuming all of the outstanding reads and throw the exception after reading back the response from EXEC. Wouldn't it work the same to avoid protocol desynchronization issues?

@jimbonator do you have any feedback about my observation ⬆️

jimbonator: Sorry, I've been caught up in other work and didn't get a chance to come back to this yesterday.

You're right, that would avoid the desynchronization issues, but I was hoping to avoid dropping the connection for server (rather than connection) errors.

src/Pipeline/Atomic.php
jimbonator: I would be ecstatic to add a subclass for the exception, but as you pointed out, there are other concerns there with the base class contract.  Personally, I'm not overly concerned about retrieving the `Client` in the `catch` block, but I can understand if other callers have structured their code around that.

> this can be indeed considered a bug but would such a change be considered backwards-compatible for a patch release?

That's ... a good point.  If there is existing code out there expecting a `ClientException` for the abort, then there would be some compatibility issues.  But this would be a good change for v2.0, since those assumptions can be revised.  Or, for v1.x, there could be a field stored in the client's `OptionsInterface` we could add signalling "throw AbortedMultiExecException in atomic pipelines" is allowed (and make it mandatory for v2.0.)

src/Pipeline/Atomic.php
jimbonator: I've reverted this change.  The queue is drained as the responses are processed.

src/Pipeline/Atomic.php
jimbonator: (reverted this as well.)

src/Pipeline/Atomic.php
jimbonator: Reverted.  Now throws `ClientException` as before.

src/Pipeline/Atomic.php
jimbonator: I've gone ahead and reverted this approach.  When an error is encountered, the connection is dropped and the exception immediately thrown.

PR #504, opened 2018-05-12
fix `AUTH` failed: ERR Client sent AUTH, but no password is set

hncg:
for example:
in laravel or other projects  .env file

```
REDIS_HOST=redis
REDIS_PASSWORD=
REDIS_PORT=6379
```

### DIFF
--- a/src/Connection/Factory.php
+++ b/src/Connection/Factory.php
@@ -173,7 +173,7 @@ class Factory implements FactoryInterface
     {
         $parameters = $connection->getParameters();
 
-        if (isset($parameters->password)) {
+        if (empty($parameters->password)) {
             $connection->addConnectCommand(
                 new RawCommand(array('AUTH', $parameters->password))
             );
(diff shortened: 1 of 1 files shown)


src/Connection/Factory.php
```diff
@@ -173,7 +173,7 @@ protected function prepareConnection(NodeConnectionInterface $connection)
     {
         $parameters = $connection->getParameters();
 
-        if (isset($parameters->password)) {
+        if (empty($parameters->password)) {
```
mikebranderhorst: if ( ! empty($parameters->password)) {

poisa:
@hncg Thank you! This has been driving me insane with Laravel until I saw your PR. I hope this gets merged ASAP!

the-veloper:
`REDIS_PASSWORD` should be set to 'null' instead of nothing.
Ex:
```
REDIS_HOST=redis
REDIS_PASSWORD=null
REDIS_PORT=6379
```

hncg:
Yes, but we shouldn’t  expect that everyone knows exactly how to set.
@the-veloper

nrk:
Fixed for v1.1 in 31baa9b (will be included in v1.1.4).

PR #514, opened 2018-07-21
PUB/SUB and MULTI/EXEC support for replication

josiasmontag:
The [limitations](http://redis.io/topics/cluster-tutorial#migrating-to-redis-cluster) of Redis clusters are currently applied to all aggregated connections. This prevents PUB/SUB and MULTI/EXEC to work for Redis replication setups.

In case of a Redis master/slave (non-cluster) setup these commands will just work fine. These commands are [not whitelisted](https://github.com/predis/predis/blob/v1.1/src/Replication/ReplicationStrategy.php#L212) as read-only operations and thus will always be executed on the master.


josiasmontag:
I just updated the PR. Would be great if we could get this bugfix into `1.1` :)

ShishkinAR:
Hi, @tillkruss! Is it possible to reopen this PR and fix an issue for v1.1? We use v1.1.10 and cannot update to 2.x for now, but we need for MultiExec support for replication connection.

tillkruss:
@ShishkinAR We don't maintain v1.x anymore, at least not actively. Why are you unable to use v2.x?

PR #527, opened 2018-09-13
Balance sentinel connections

audriusbugas:
On bigger scale there is possibility to kill first sentinel just because of connection rate


dark9911:
Very nice! I'd love to see this feature get committed.

PR #542, opened 2018-12-04
Do not add slave connection on broken link

tomas7777:
When the slave is being reconnected after being disconnected and is out of sync link with master is broken and making read calls could return either `null` or `LOADING Redis is loading the dataset in memory` response. To ensure that client connects to good health slave, sentinel provides `master-link-status` which shows current status.


coconut30:
Hi, is there any chance that this PR will be part of an 1.1.7 release ?

PR #545, opened 2019-01-09
 fix small mistake

gitsrc:
fix small mistake


nrk:
Merged into branch `v1.1` so this fix will be included in the next patch release (1.1.4).

JFYI the original commit message was too vagues so I edited it for clarity, see 2440a61.

Thank you!

PR #553, opened 2019-03-15
Update rpush\lpush doc

MichaelRain:
In a fact lpush\rpush doesn't support mass values insertion. It just insert it to the list.


Baojs:
“/src/Command/Command.php“ 
You can see that there is a normalizeVariadic function in this class.

nrk:
Both [`LPUSH`](https://redis.io/commands/lpush) and [`RPUSH`](https://redis.io/commands/rpush) are variadic indeed.

Issue #468, opened 2017-10-18
SORT with read-only slave


buffcode:
Affects: v1.1.1
Redis Version: 3.2.8

According to https://github.com/antirez/redis/blob/4.0/src/server.c#L252 the `SORT` command is a write command (even without `STORE`), also in 3.x.

I'm currently running into `Predis\Response\ServerException - #0: READONLY You can't write against a read only slave.` with following code using no STORE at all:
```php
$redis->sort(
  'somekey',
  [
    'by' => 'NOSORT',
    'get' => [
      '#',
      'somekey:*->event',
      'somekey:*->location'
    ]
  ]
);
```

nrk:
Fix will be available in Predis v1.1.4. Thank you.

PR #554, opened 2019-03-18
update sort command to be a write operation

mfettig:
Addresses #468 

As noted in the linked issue, Redis treats all SORT commands as 'write' operations
https://github.com/antirez/redis/blob/4.0/src/server.c#L252

This patch removes 'SORT' from the array returned by ReplicationStrategy::getReadOnlyOperations() and removes the isSortReadOnly() helper method. Related tests were also updated.

### DIFF
--- a/src/Replication/ReplicationStrategy.php
+++ b/src/Replication/ReplicationStrategy.php
@@ -90,31 +90,6 @@ class ReplicationStrategy
     }
 
     /**
-     * Checks if a SORT command is a readable operation by parsing the arguments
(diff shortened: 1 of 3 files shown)


mfettig:
build appears to fail because travis is using redis-server with version 4.0.x, which has a different error message than expected when selecting a DB out of range

https://github.com/antirez/redis/commit/c7a4e694ad3689d934897b0d2c37144d7b2d0b97

nrk:

Issue #552, opened 2019-03-12
Client class is an improper implementation of \IteratorAggregate, causes false positive errors in CI builds and when debug libraries are present.


mopsyd:
Class `Predis\Client` has a malformed implementation of `\IteratorAggregate`, which causes a number of debugging libraries to throw false positive warnings on exceptions generated when attempting to get an iterator from the object. This causes unwarranted exceptions when using debugging libraries like Kint, and also can lead to false positive errors that break PHPUnit and similar unit testing frameworks, falsely flagging builds as failing.

A proper implementation of `\IteratorAggregate` does not throw an exception under any circumstances. This is the wrong place to check for the error condition. The Predis class that expects the iterable needs to be performing the check and raising the exception if it is not as expected, not the `getIterator` method, which should always return an iterable result cleanly.

As this interface implements "magic" behavior that is affects the native PHP language construct `foreach`, the `getIterator` method is expected not to raise errors, throw exceptions, or otherwise behave in a manner that is not expected of an already defined traversable (eg a native array).

There is no stacktrace to provide because warnings do not raise exceptions and are not "catchable", however they do cause CI builds to fail, breaks downloads such as PDFs with its error text (such as generating a pdf for client download), and causes numerous debugging libraries to throw false positives when deeply analyzing the object (eg Kint). The exception thrown *does not halt execution in lieu of a catch block*, and is instead converted to a warning by PHP.

As no stack trace is available due to the error causing conflicts with PHP internals that raises warnings and is not "catchable", I have taken the liberty of isolating the root cause for you.

In the provided block below, the offending piece of logic is `throw new ClientException('The underlying connection is not traversable');`

This is the offending function:

```php
public function getIterator()
    {
        $clients = array();
        $connection = $this->getConnection();

        if (!$connection instanceof \Traversable) {
            throw new ClientException('The underlying connection is not traversable');
        }

        foreach ($connection as $node) {
            $clients[(string) $node] = new static($node, $this->getOptions());
        }

        return new \ArrayIterator($clients);
    }
```

This leads to errors such as the following spamming the error log/cli console:
```
PHP Warning:  An exception (Predis\\ClientException) was thrown in /path/to/application/vendor/predis/predis/src/Client.php on line 538 while executing Kint Parser Plugin "Kint\\Parser\\IteratorPlugin". Error message: The underlying connection is not traversable in /path/to/application/vendor/kint-php/kint/src/Parser/Parser.php on line 589, referer: http://sandbox.localhost/
```

[...]

PR #556, opened 2019-03-25
Iterate not iterable connections

peter-gribanov:
See #552 for more details


nrk:
Merged into branch `v1.1` so this fix will be included in the next patch release (1.1.4).Thank you!

PR #566, opened 2019-06-19
change all strtoupper method to mb_strtoupper

eroltukenmez:


nrk:
I'm closing this PR for a few reasons:

- it tries to change stuff unrelated to the underlying issue (TravisCI)
- the `mbstring` extension must be installed for `mb_strtoupper()` to work, but it's a non-default extension.
- regardless, `mb_strtoupper()` is slow as hell which might not be a problem in most normal usage cases for this function, but when you start issuing thousands of commands (see pipelines) it can add seconds just for the conversion. Doing a plain `strtr($string, 'abcdefghijklmnopqrstuvwxyz', 'ABCDEFGHIJKLMNOPQRSTUVWXYZ');` would be way faster (still with a greater overhead than `strtoupper()` though) and it would be enough for our use-case since Redis uses plain ASCII letters for command IDs by convention, but it'd would be a PITA to maintain such code spread through the whole codebase.

You could try mitigating this by doing a `$client->INFO()` instead of `$client->info()` but I'm plenty aware this is *not* a fix as this won't solve anything which is outside your control, e.g. other libraries or applications.

Having said that I do aknowledge that when the Tukish locale is in use we have a problem and that's why I never closed the first issue reported about this (see #372). I'm still thinking about how we could solve this in a future release with a more maintainable approach and without impacting performances unless strictly needed. I have a few ideas, but nothing set in stone yet.

PR #567, opened 2019-06-24
Fix issues when using setlocale

Dawnthorn:
If you set the locale to Turkish before running $redis->expire(...) you
will get this error:
Command 'EXPÝRE' is not a registered Redis command.

This is because strtoupper capitalizes expire to EXPÝRE. mb_strtoupper
dosen't use the locale, so it doesn't have this problem.

Here's an example. You need to have the Turkish locale installed on your system for this to show the bug:
```
setlocale(LC_ALL, 'tr_TR');
print(strtoupper('expire') . "\n");
```

### DIFF
--- a/src/Profile/RedisProfile.php
+++ b/src/Profile/RedisProfile.php
@@ -45,7 +45,7 @@ abstract class RedisProfile implements ProfileInterface
      */
     public function supportsCommand($commandID)
     {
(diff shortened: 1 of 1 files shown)


Dawnthorn:
I just saw PR https://github.com/nrk/predis/pull/566 which fixes my problem, but it's more comprehensive. So I'm closing this one.

Issue #570, opened 2019-07-26
Wiki link broken


adjenks:
![image](https://user-images.githubusercontent.com/10687569/61984201-30dace00-afb8-11e9-865c-2376e3d3da17.png)
![image](https://user-images.githubusercontent.com/10687569/61984207-39cb9f80-afb8-11e9-88be-92677dba5686.png)

xemose:
Wiki link should be fixed to https://github.com/nrk/predis/wiki

MeroveeCharrue:
There is an open PR to fix exactly this : #589.

PR #589, opened 2019-11-25
Fix wiki link

funivan:


MeroveeCharrue:
I was about to offer this exact fix. Thank you for doing it.

It should be linked to this issue : #570

Issue #592, opened 2019-12-31
Adding support for ZPOPMAX command


AhmedRaafat14:
**I would like to start my contribution** to this project by adding support for [**ZPOPMAX command**](https://redis.io/commands/zpopmax) available since Redis 5.0.

PR #593, opened 2020-01-05
Add new command: ZPOPMAX (Redis 5.0)

AhmedRaafat14:
Added [**ZPOPMAX command**](https://redis.io/commands/zpopmax).

Related to issue #592

Example of usgae:
```php
$client = new \Predis\Client('tcp://127.0.0.1:6379', array('profile' => '5.0'));

$values = array(
    'one' => 1,
    'two' => 2,
    'three' => 3,
    'four' => 4,
);

$client->zadd('myzset', $values);

# $popedValues = $client->zpopmax('myzset');
# With count specified
$count = 2;
$popedValues = $client->zpopmax('myzset', $count);

print_r($popedValue);
```

### DIFF
--- a/src/Profile/RedisVersion320.php
+++ b/src/Profile/RedisVersion320.php
@@ -276,6 +276,11 @@ class RedisVersion320 extends RedisProfile
             'GEODIST' => 'Predis\Command\GeospatialGeoDist',
             'GEORADIUS' => 'Predis\Command\GeospatialGeoRadius',
             'GEORADIUSBYMEMBER' => 'Predis\Command\GeospatialGeoRadiusByMember',
+
+            /* ---------------- Redis 5.0 ---------------- */
+
+            /* commands operating on sorted sets */
+            'ZPOPMAX' => 'Predis\Command\ZSetRemoveMaxByScore',
         );
     }
 }
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -38,7 +38,7 @@
 
     <php>
         <!-- Redis -->
-        <const name="REDIS_SERVER_VERSION" value="3.2" />
+        <const name="REDIS_SERVER_VERSION" value="5.0" />
         <const name="REDIS_SERVER_HOST" value="127.0.0.1" />
         <const name="REDIS_SERVER_PORT" value="6379" />
         <const name="REDIS_SERVER_DBNUM" value="15" />
--- a/src/ClientContextInterface.php
+++ b/src/ClientContextInterface.php
@@ -163,6 +163,7 @@ use Predis\Command\CommandInterface;
  * @method $this geodist($key, $member1, $member2, $unit = null)
  * @method $this georadius($key, $longitude, $latitude, $radius, $unit, array $options = null)
  * @method $this georadiusbymember($key, $member, $radius, $unit, array $options = null)
+ * @method $this zpopmax($key)
  *
  * @author Daniele Alessandri <suppakilla@gmail.com>
  */
--- a/src/ClientInterface.php
+++ b/src/ClientInterface.php
@@ -171,6 +171,7 @@ use Predis\Profile\ProfileInterface;
  * @method string geodist($key, $member1, $member2, $unit = null)
  * @method array  georadius($key, $longitude, $latitude, $radius, $unit, array $options = null)
  * @method array  georadiusbymember($key, $member, $radius, $unit, array $options = null)
+ * @method array  zpopmax()
  *
  * @author Daniele Alessandri <suppakilla@gmail.com>
  */
--- a/src/Command/Processor/KeyPrefixProcessor.php
+++ b/src/Command/Processor/KeyPrefixProcessor.php
@@ -166,6 +166,8 @@ class KeyPrefixProcessor implements ProcessorInterface
             'GEODIST' => 'static::first',
             'GEORADIUS' => 'static::georadius',
             'GEORADIUSBYMEMBER' => 'static::georadius',
+            /* ---------------- Redis 5.0.0 ---------------- */
+            'ZPOPMAX' => 'static::first'
         );
     }
 
--- a/src/Command/ZSetRemoveMaxByScore.php
+++ b/src/Command/ZSetRemoveMaxByScore.php
@@ -1,0 +1,46 @@
+<?php
+
+/*
+ * This file is part of the Predis package.
+ *
+ * (c) Daniele Alessandri <suppakilla@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Predis\Command;
+
+/**
+ * Class ZSetRemoveMaxByScore
+ * @package Predis\Command
+ *
+ * @link https://redis.io/commands/zpopmax
+ *
+ * @author Ahmed Raafat <ahmed.raafat1412@gmail.com>
+ */
+class ZSetRemoveMaxByScore extends Command
+{
+    /**
+     * @inheritDoc
+     */
+    public function getId()
+    {
+        return 'ZPOPMAX';
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function parseResponse($data)
+    {
+        $response = array();
+
+        $responseCount = count($data);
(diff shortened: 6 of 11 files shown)


src/Profile/RedisVersion320.php
```diff
@@ -276,6 +276,11 @@ public function getSupportedCommands()
             'GEODIST' => 'Predis\Command\GeospatialGeoDist',
             'GEORADIUS' => 'Predis\Command\GeospatialGeoRadius',
             'GEORADIUSBYMEMBER' => 'Predis\Command\GeospatialGeoRadiusByMember',
+
+            /* ---------------- Redis 5.0 ---------------- */
+
+            /* commands operating on sorted sets */
+            'ZPOPMAX' => 'Predis\Command\ZSetRemoveMaxByScore',
```
hannesvdvreken, line 283: probably this should be a different profile, no? Like `RedisVersion500.php`?

AhmedRaafat14: I think you are right.
I will update that :).

I thought no one gonna look into that anymore 😃.

Thank you.

AhmedRaafat14: Added the new profile.
@hannesvdvreken & @tillkruss re-check :)

AhmedRaafat14:
I made `v5.0` is the default one and also the default for the test cases, if no one objects!

nrk:
In general, adding new commands to Predis is not as straightforward as it might seem if anyone wants to do it "properly".

Considering the ongoing transition, I'll write down some notes here based on the `v1.0` branch to anyone benefits. I'll try expanding these bits later to have them highlighted in a more appropriate and accessible place.

Aside from having the actual command class prepared by extending `Predis\Command\Command` or implementing `Predis\Command\CommandInterface`, depending on the situation, and added to the list of supported commands of the respective [`Predis\Profile\RedisProfile`](https://github.com/nrk/predis/blob/v1.1.1/src/Profile/RedisVersion320.php) class, there are a few places that needs to be taken care of through Predis to have the library fully and properly support a new command:

1. The command signature should be added as @method in the phpdocs of both [`Predis\ClientInterface`](https://github.com/nrk/predis/blob/v1.1.1/src/ClientInterface.php#L27-L173) and [`Predis\ClientContextInterface`](https://github.com/nrk/predis/blob/v1.1.1/src/ClientContextInterface.php#L19-L165).

2. If the command has keys in the argument list, then it should be handled accordingly by [`Predis\Command\Processor\KeyPrefixProcessor`](https://github.com/nrk/predis/blob/v1.1.1/src/Command/Processor/KeyPrefixProcessor.php)

3. For replication scenarios such as redis-sentinel, it should be handled accordingly by [`Predis\Replication\ReplicationStrategy`](https://github.com/nrk/predis/blob/v1.1.1/src/Replication/ReplicationStrategy.php) depending if the command represents a read-only operation (meaning that the client can stick to a slave node) or it's disallowed altogheter. The client by default treats commands not explicitly handled here as a write operations, forcing a switch to the master node (but Redis could also return a -ERR response).

4. For cluster scenarios, either simple client-sharding or redis-cluster, it should be handled accordingly by [`Predis\Cluster\ClusterStrategy`](https://github.com/nrk/predis/blob/v1.1.1/src/Cluster/ClusterStrategy.php). The client by default treats commands not explicitly handled here as disallowed when using cluster, thus raising an exception.

5. Last but not least a proper PR for each new command should also ship with tests not because this should be the common practice, but also because when there's so much stuff going on it's hard to do that at a later stage and the risk of forgetting something along the way becomse pretty high. By the way the current test suite has [a few helpers](https://github.com/nrk/predis/tree/v1.1.1/tests/PHPUnit) that would most likely benefit from some explanations.

As a full example of all the steps involved, you can see my past commit that adds [support for `BITFIELD`](https://github.com/nrk/predis/commit/3362e474b66c4de8b5639991ec61bae8984a0566).

By the way, in this specific case we actually miss profiles for both Redis 4.0 and 5.0 and the commands introduced in these respective major releases. Provided that the priority for now is getting a patch release of Predis in the `v1.1` branch and adding new commands should come at a later stage (see the roadmap drafted by @tillkruss), my proposal would be to start by committing two separate profile classes with no new commands added and then implement each new command in a separate PR, with all the proper bits and with no rush. Getting a list of commands by each Redis release is actually easy thanks to [the commands.json file in redis-doc](https://github.com/redis/redis-doc/blob/master/commands.json), this definitely helps to see what we are missing.

We should also keep in mind that `v1.1` and `master` diverged quite a lot which is one more reason to think twice about implementing all of the new commands in both branches as this would impact on the time and effort required to get things actually done, so we should clear this aspect before start committing new commands IMHO.

nrk:
In general, adding new commands to Predis is not as straightforward as it might seem if anyone wants to do it "properly".

Considering the ongoing transition, I'll write down some notes here based on the `v1.0` branch to anyone benefits. I'll try expanding these bits later to have them highlighted in a more appropriate and accessible place.

Aside from having the actual command class prepared by extending [`Predis\Command\Command`](https://github.com/nrk/predis/blob/v1.1.1/src/Command/Command.php) or implementing [`Predis\Command\CommandInterface`](https://github.com/nrk/predis/blob/v1.1.1/src/Command/CommandInterface.php), depending on the situation, and added to the list of supported commands of the respective [`Predis\Profile\RedisProfile`](https://github.com/nrk/predis/blob/v1.1.1/src/Profile/RedisVersion320.php) class, there are a few places that needs to be taken care of through Predis to have the library fully and properly support a new command:

1. The command signature should be added as @method in the phpdocs of both [`Predis\ClientInterface`](https://github.com/nrk/predis/blob/v1.1.1/src/ClientInterface.php#L27-L173) and [`Predis\ClientContextInterface`](https://github.com/nrk/predis/blob/v1.1.1/src/ClientContextInterface.php#L19-L165).

2. If the command has keys in the argument list, then it should be handled accordingly by [`Predis\Command\Processor\KeyPrefixProcessor`](https://github.com/nrk/predis/blob/v1.1.1/src/Command/Processor/KeyPrefixProcessor.php)

3. For replication scenarios such as redis-sentinel, it should be handled accordingly by [`Predis\Replication\ReplicationStrategy`](https://github.com/nrk/predis/blob/v1.1.1/src/Replication/ReplicationStrategy.php) depending if the command represents a read-only operation (meaning that the client can stick to a slave node) or it's disallowed altogheter. The client by default treats commands not explicitly handled here as a write operations, forcing a switch to the master node (but Redis could also return a -ERR response).

4. For cluster scenarios, either simple client-sharding or redis-cluster, it should be handled accordingly by [`Predis\Cluster\ClusterStrategy`](https://github.com/nrk/predis/blob/v1.1.1/src/Cluster/ClusterStrategy.php). The client by default treats commands not explicitly handled here as disallowed when using cluster, thus raising an exception.

5. Last but not least a proper PR for each new command should also ship with tests not because this should be the common practice, but also because when there's so much stuff going on it's hard to do that at a later stage and the risk of forgetting something along the way becomse pretty high. By the way the current test suite has [a few helpers](https://github.com/nrk/predis/tree/v1.1.1/tests/PHPUnit) that would most likely benefit from some explanations.

As a full example of all the steps, you can see my past commit that adds [support for `BITFIELD`](https://github.com/nrk/predis/commit/3362e474b66c4de8b5639991ec61bae8984a0566).

In the specific case

AhmedRaafat14:
Thank you @nrk for sharing :).

You are right I will create a separate PR that introduces the two missing profiles (`400 & 500`), and keep this PR only focuses on the new command.


I think your whole comment needs to be moved as an announcement for the future contributors to be able to see it as well :).

nrk:
> You are right I will create a separate PR that introduces the two missing profiles (400 & 500), and keep this PR only focuses on the new command.

TBH I suggest we coordinate a bit on how to proceed before going ahead with these PRs, let's decide on the roadmap in detail on #615 so that any new work (beside fixes I mean) ends up targeting the proper branch. I'd still avoid any new stuff in a patch release on the `v1.1` branch, but we still need to decide if we do need an intermediate minor release which means a `v1.2` branch or if we want to go ahead with v2.0 and then target the `master` branch directly.

> I think your whole comment needs to be moved as an announcement for the future contributors to be able to see it as well :)

Yes that's exactly what I meant with "_a more appropriate and accessible place_" 😉

PR #599, opened 2020-02-21
Added possibility to call info on Master/Slave Replication

DemigodCode:
Hi,

Symfony uses the "info"-command to get the server version on which it is connected to.
This is done by the cache RedisTrait.

Afaik phpredis will just forward this command, predis should do the same.
I guess it's a better idea to use the master for such commands, so it's temporarily used for non-write Requests to redis.

I also fixed the travis.yml and removed a check of an exception message which is different between older/newer versions of redis.


DemigodCode:
Could not completly activate PHP 7.4 test. There's more work needed.
It's added but allowed to fail, cause of deprecation warnings.

PR #609, opened 2020-06-16
Cluster improvements

ebogdanov:
Hi.

Thanks for your project.

I'd like suggest to review my changes in code, some of them can be helpful for others.
All of them about better stability in Cluster mode.

1. Pipeline handling if during command workflow **MOVED** response was sent from Redis side.
2. Connection veryfication. This feature is mostly for checking issues with persistent connection, to ensure that network sequence of request-reply is not broken.
3. Try to reconnect once on network issues. Current logic deletes connection from pool if some problem was met. This can degradate local state saved on client side, since that there is not any nodes left to interact with. Or in case of "network fast up&down" code will throw away current connection to node, connect to some other node, ask state of slots, and only after that will return back to first slot. So that I've added 1 try to re-connect to the failed node, which will help to ensure that problem with selected node is permanent, and there is any purpose of ask cluster state from other node.

Any of your feedback is much appreciated.

Regards,


src/Connection/Aggregate/RedisCluster.php
```diff
@@ -196,7 +200,7 @@ public function buildSlotsMap()
             }
         }
 
-        return $this->slotsMap;
+        return $this->slotsMap ?? [];
```
tillkruss, line 203: This won't work on older PHP versions.

tillkruss (changes requested):

composer.json
```diff
@@ -16,7 +16,7 @@
         }
     ],
     "require": {
-        "php": ">=5.3.9"
+        "php": ">=7.0"
```
tillkruss, line 19: We can't change the PHP version for the v1.1 branch.

tillkruss (changes requested):

PR #611, opened 2020-07-29
Bring travis config up to date

hannesvdvreken:


dereuromark:
Also PHP7.4 (and maybe 8) should be added to matrix soon IMO.

PR #617, opened 2020-08-11
Allow failures on TravisCI for PHP 5.3 to 5.5

srmklive:


nrk:
Why?

srmklive:
If its not needed, i can close it.

tillkruss:
@srmklive: Can you explain your reasoning behind it?

srmklive:
The reasoning is the build on TravisCI shouldn't fail due to old php versions. If you think, this is not needed. Then lets close the PR.

nrk:
The reason why I don't think it makes much sense to allow failures for those versions on `v1.1` is that if tests fail (and they do not fail, right now) then they should be fixed because we'll still support PHP 5.3, 5.4 and 5.5 for patch releases in the above mentioned branch.

If we'd be willing to accept failures for ancient versions of the runtime then we should just drop support for them but as I've said this shouldn't be the case for `v1.1`, but might be OK for the next minor release and will be a gived for the next major release.

On the contrary, allowing failures for PHP 8 would be OK since it's still a work in progress and things may break anytime just because of that.

tillkruss:
See 7da604b 👍🏻

Issue #620, opened 2020-08-11
Support PHP 8, drop PHP 5 & 7.0 - 7.1


tillkruss:
As discussed in #615. Related #616.

dereuromark:
This seems to be already done in https://github.com/nrk/predis/blob/main/.travis.yml (main)
and can be closed then.

tillkruss:
@dereuromark: I'll keep this open until the tests pass again.

Are you up for fixing the test failures on `main` that came with the PHPUnit upgrade from 4 to 8?

AhmedRaafat14:
> @dereuromark: I'll keep this open until the tests pass again.
> 
> Are you up for fixing the test failures on `main` that came with the PHPUnit upgrade from 4 to 8?


I can help with this if needed as well :).

dereuromark:
I didnt see the test fails
@AhmedRaafat14 much appreciated.

tillkruss:
@AhmedRaafat14: Go for it 👍🏻

tillkruss:
@emmanix2002 @AhmedRaafat14 You can split it up, too, so we have many small PRs and not one massive one 🙂

There seem to be quite a few sectors:

- Depreciation notices
- Redis 6 changes
- Broken mocks

AhmedRaafat14:
I was just thinking about the same thing.

@emmanix2002 you can tell what you already started fixing and I can take another point, what do you think :)!

Otherwise, I will just start looking into the docs 😉

emmanix2002:
@AhmedRaafat14 fixed issues for some of the files inside of `tests/PHPUnit` for now, specifically:

- `tests/PHPUnit/ArrayHasSameValuesConstraint`
- `tests/PHPUnit/PredisTestCase`
- `tests/PHPUnit/RedisCommandConstraint`

Currently on:

- `tests/PRedis/ClientExceptionTest`

There are quite a number of issues but the ones in the `PHPUnit` directory are the lowest hanging fruits

emmanix2002:
@AhmedRaafat14 quick question, I don't know if we want to set `namespace` declarations inside the test files. IMO, we should only use namespace declarations for the actual source file.

I ask because the files inside `tests/Predis` have namespace declarations on them.

What do you think?

AhmedRaafat14:
> I ask because the files inside `tests/Predis` have namespace declarations on them.


[...]

PR #616, opened 2020-08-11
Support PHP 7.4 in v1.1.x

tillkruss:
Looks like for PHP 7.4 we need to patch PHPUnit 😑

PR #625, opened 2020-08-12
Support PHP 8.0 in v1.1.x

AhmedRaafat14:
Related to #620

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,3 +31,5 @@ matrix:
     - php: 7.3
     - php: 7.4
     - php: nightly # This is how Travis call PHP 8.0.0-dev - php-nightly
+  allow_failures:
+    - php: nightly
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,6 +30,6 @@ matrix:
     - php: 7.2
     - php: 7.3
     - php: 7.4
-    - php: nightly # This is how Travis call PHP 8.0.0-dev - php-nightly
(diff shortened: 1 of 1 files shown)


nrk:
... and it even passes the test run, how cool is that? 😎

Still, I think you should add `php: nightly` to `allow_failures` in the matrix as things might break unexpectedly without being our fault considered PHP 8 is still a work in progress.

By the way, note the difference

PHP 5.4 `Time: 1.27 minutes, Memory: 106.50MB`
PHP 7.4 `Time: 38.76 seconds, Memory: 42.00MB`
PHP 8.0 `Time: 13.15 seconds, Memory: 28.00MB`

AhmedRaafat14:
> ... and it even passes the test run, how cool is that? 😎

Yes, it is so cool 😎 and it makes me happy when it did and thanks also go to the patch applied in #616 by @tillkruss.

> Still, I think you should add `php: nightly` to `allow_failures` in the matrix as things might break unexpectedly without being our fault considered that PHP 8 is still a work in progress.

Good point, doing it.

> By the way, note the difference:
> 
> PHP 5.4 `Time: 1.27 minutes, Memory: 106.50MB`
> PHP 7.4 `Time: 38.76 seconds, Memory: 42.00MB`
> PHP 8.0 `Time: 13.15 seconds, Memory: 28.00MB`

Yes, it is so incrediable to see this huge difference on an awesome package like predis :relaxed:

PR #626, opened 2020-08-12
Remove PEAR/PHAR

### LINKED ISSUE #619: Drop PEAR support
As discussed in #587 and #615.

tillkruss:
Resolves #619.

@nrk: What are your thoughts on the `bin/create-single-file` script, does that still work? Should we drop it in v2.0?


nrk:
@tillkruss the script `bin/create-single-file` despite being an hack should still be working, at least in v1.1, but I don't think I've ever tested it with v2.0 and I wasn't really thinking about keeping it around as it's something that dates back from ten years ago or so, sometime before Composer, when people nagged me because they had to use autoloading with v0.7 🤷 I guess we can drop it with no remorse.

Actually the script used to generate a Phar archive was born for the same reason above, but by today's standards Phar archive make sense mostly for CLI applications.

Issue #624, opened 2020-08-11
Recover changes from old master branch


tillkruss:
Is there anything we should recover from the stale master branch?

From #587: 

> The master branch is meant to become v2.0, but since it diverged quite a lot from v1.1 most fixes cannot be easily merged. At first I intended to maintain both v1.1 and master-to-be-v2.0 but in retrospect I guess that plan actually backfired so now the plan is to just fix v1.1 for both implementation bugs and compatiblity issues with newer PHP versions, maybe a few new Redis commands, and leave it at that. On the master branch, aside from applying the same fixes, we should drop support once and for all for PEAR and older PHP versions and rework parts of code that could benefit from new features introduced in PHP 7 before implementing anything new.

AhmedRaafat14:
That is really a good question.
Personally, I have to check the differences and see, but mostly I think we just need to follow the roadmap and don't port allot of changes.

But yes we will see after we check :).

nrk:
During the next few days I'll review every change I applied back then, but as far as I remember everything I pushed there had a reason to exist in the next major version. Just to make an example, one of the biggest changes in v2.0 is how commands are handled by default with a convention-over-configuration approach (see 62b421f, 4b47639 and 822f02b) which basically supersedes the whole concept of profiles. Maybe some things could be improved now that we'll finally drop older PHP versions, but the overall direction was indeed the right one.

PR #627, opened 2020-08-12
Drop HHVM

### LINKED ISSUE #618: Drop HHVM support
Like most PHP packages, let's drop HHVM support in v2.0. v1.1 will continue to support it, but will only receive bug fixes.

tillkruss:
Resolves #618.

### DIFF
--- a/tests/PHPUnit/PredisConnectionTestCase.php
+++ b/tests/PHPUnit/PredisConnectionTestCase.php
@@ -436,13 +436,7 @@ abstract class PredisConnectionTestCase extends PredisTestCase
      */
     public function testThrowsExceptionOnConnectionTimeout()
     {
-        // TODO: float timeouts for connect() under HHVM 3.6.6 are broken and,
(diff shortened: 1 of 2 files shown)


tillkruss:
https://github.com/predishq/predis/blob/main/src/Configuration/ProfileOption.php#L32-L42

@nrk: Can we drop, too?

nrk:
Dropping support for HHVM means that we can remove any hack related to it from the codebase.

On a related note, honestly I'd slow things down a bit for now with merging stuff that modifies too much of the source code on `main` as it might end up giving us some headaches later when porting the changes back from `master-stale`, as I'm pretty sure most stuff from there is still valid and could even be re-applied as-is, with just some squashing and a few clean-ups (see #624).

Just give me a few days at least to investigate 😃

tillkruss:
@nrk sure, no problem!

nrk:
I didn't even notice but I removed all of the HHVM-related hacks in #654 which is now merged into `main`, so we can close this.

PR #629, opened 2020-08-13
Fix ArrayHasSameValuesConstraint

b1rdex:

Issue #615, opened 2020-08-10
Roadmap to v2.0


tillkruss:
Some thoughts on the future roadmap.

## v1.1

- Bug fixes only
- Maintain support for PHP 5.6 - 7.4 and Redis 3-5

_PHP 7.2 is EOL will get security patches only until 30 Nov 2020._

## v2.0

- Drop HHVM & PEAR
- Support PHP 7.2 - 8.0
- Merge new features and new Redis commands
- Support for Redis 3-6?
- API documentation?
- Move to GH Actions?

tillkruss:
Ping @AhmedRaafat14 @b1rdex @dereuromark @emmanix2002 @srmklive @rlunar.

b1rdex:
V1 branch should get full php < 8 support. May be php 8 too if that doesn't require too much changes. We use predis with php 7.4 in production with no bugs found.

tillkruss:
@b1rdex That depends on PHPUnit. If we can find a version that supports 5.3 - 7.4, sure. If not we might need to branch off v1.2 that supports 5.6 - 7.4, but drops anything older.

b1rdex:
It’s possible to patch phpunit using composer patches to get even phpunit 4.8 to work with php 8. Example: https://github.com/yiisoft/yii/pull/4306

Отправлено с iPhone

tillkruss:
@b1rdex Neat, can you open a PR test v1.1 with 7.3 and 7.4?

emmanix2002:
What do you need yes to do @tillkruss ?

AhmedRaafat14:
Cool!
I completely agree with @b1rdex `v1.1` should get the full support for `PHP < 8`, and the same for `v2.0`.

As @tillkruss suggested we can call it `v1.2`.

We use Predis with PHP 7.2 & Symfony 4.4 in production with no bugs so far.

So it should be easy & neat.

nrk:
I'd also drop PEAR support, at least from v2.0 and going forward.

Honestly it's always been a nightmare to have it build packages of Predis on each release as I had various problems in the past with getting the build environment running properly from time to time (the actual build process also requires [some hackish stuff](https://github.com/nrk/predis/blob/dbfc1a74ba524636af2e3ef2c6d330d39098345f/bin/create-pear)). Furthermore I guess PEAR is a dead man walking anyway.

I'll try to get it work for the next patch release at least.

dereuromark:
[...]

PR #630, opened 2020-08-13
Downgrade phpunit

b1rdex:
I tried to rewrite the tests in #629 and found it very hard.
IMO we should better stay at PHPUnit 4.8 for now then have all the tests broken.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -31,7 +31,8 @@
         "php": "^7.2"
     },
     "require-dev": {
-        "phpunit/phpunit": "^8.0"
+        "phpunit/phpunit": "~4.8",
+        "cweagans/composer-patches": "^1.6"
     },
     "suggest": {
         "ext-phpiredis": "Allows faster serialization and deserialization of the Redis protocol",
@@ -41,5 +42,17 @@
         "psr-4": {
             "Predis\\": "src/"
         }
-    }
+    },
+    "extra": {
+ 		"composer-exit-on-patch-failure": true,
+ 		"patches": {
+ 			"phpunit/phpunit-mock-objects": {
+ 				"Fix PHP 7 and 8 compatibility": "./tests/phpunit_mock_objects.patch"
+ 			},
+ 			"phpunit/phpunit": {
+ 				"Fix PHP 7 compatibility": "./tests/phpunit_php7.patch",
+ 				"Fix PHP 8 compatibility": "./tests/phpunit_php8.patch"
+ 			}
+ 		}
+ 	}
 }
--- a/tests/phpunit_mock_objects.patch
+++ b/tests/phpunit_mock_objects.patch
@@ -1,0 +1,38 @@
+diff --git a/src/Framework/MockObject/Generator.php b/src/Framework/MockObject/Generator.php
+index 3df3abb..fc76d5d 100644
+--- a/src/Framework/MockObject/Generator.php
++++ b/src/Framework/MockObject/Generator.php
+@@ -1031,16 +1031,29 @@ protected function getMethodParameters(ReflectionMethod $method, $forCall = fals
+             $typeDeclaration = '';
+ 
+             if (!$forCall) {
++                if (PHP_VERSION_ID >= 80000) {
++                    $isArray = $parameter->getType() && $parameter->getType()->getName() === 'array';
++                    $isCallable = $parameter->getType() && $parameter->getType()->getName() === 'callable';
++                } else {
++                    $isArray = $parameter->isArray();
++                    $isCallable = version_compare(PHP_VERSION, '5.4.0', '>=') ? $parameter->isCallable() : false;
++                }
++
+                 if ($this->hasType($parameter)) {
+-                    $typeDeclaration = (string) $parameter->getType() . ' ';
+-                } elseif ($parameter->isArray()) {
++                    $type = $parameter->getType();
++                    $typeDeclaration = (PHP_VERSION_ID >= 70100 ? $type->getName() : (string) $type) . ' ';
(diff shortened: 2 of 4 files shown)


AhmedRaafat14:
Hmm so this means we drop updating the tests and stick to PHPUnit 4?!
> what we discuessed [here](https://github.com/predishq/predis/issues/620#issuecomment-672960734)

b1rdex:
I'd propose staying on the PHPUnit 4 for now. Just to have the tests passing.
Then we can upgrade it in a PR, rewriting/fixing all the tests.

AhmedRaafat14:
> I'd propose staying on the PHPUnit 4 for now. Just to have the tests passing.
> Then we can upgrade it in a PR, rewriting/fixing all the tests.

I agree but that was the plan for v2.2, we kept it as it is for v1.1 now.

Maybe @tillkruss have a different opnion! lets see what he thinks.

b1rdex:
I see no point in having the main branch tests broken.
I'm not against upgrading phpunit, I just want to have the tests working to check new code/fixes.

AhmedRaafat14:
> I see no point in having the main branch tests broken.
> I'm not against upgrading phpunit, I just want to have the tests working to check new code/fixes.

yes I agree with you it should be working and someone else doing the tests update.

My point was to see if we really have to do the upgrade or not will keep it like this in v2, will see.

nrk:
> I'd propose staying on the PHPUnit 4 for now. Just to have the tests passing.
> Then we can upgrade it in a PR, rewriting/fixing all the tests.

@b1rdex I agree. I mean I **do** think we should upgrade to a newer PHPUnit version for Predis v2.0 but I just don't see the point of doing it right now if PHPUnit 4 still works with PHP 7.4 / PHP 8.0 after patching. Investigating on what needs to be done for the upgrade is fine, obviously, but doing the actual work of rewriting tests should be done when the codebase on `main` will be stable enough in terms of changes in order to avoid wasting time.

On a kind of related note, I understand the excitement of reviving Predis and it's awesome but IMHO we are going a bit too fast and sometimes getting too trigger-happy with merges and commits, for no practical reasons. Clearly this doesn't mean we should stick with how I was getting things done back when I was actively working on this library alone, it wouldn't work now and I was probably doing the exact opposite to the extreme anyway (test stuff locally first on multiple runtimes, think about changes twice or thrice for a few days and then finally push stuff). Just my point of view, that goes without saying 😉

> My point was to see if we really have to do the upgrade or we will keep it like this in v2, will see.

@AhmedRaafat14 it really depends on the effort required for upgrading to a newer PHPUnit, but I think it will be worth doing so for v2.

AhmedRaafat14:
@nrk I completely agree with you as we decided to get a little bit slow, I also would love to merge this PR at least to have a successful build on the main branch so people can feel trusty on their code changes exactly as @b1rdex suggested.

Also, I will take a lazy approach to upgrade PHPUnit and fix the tests step by step as we go & will always rebase to latest changes on the main so be inline with the code changes and not get many merge conflicts at the end 😅.

tillkruss:
I agree with upgrading PHPUnit and fixing the tests slowly over the next weeks. No rush, but keeping PHPUnit v4 going seems like dragging on baggage.

tillkruss:
Should we merge #629 in the meantime to fix up some of the tests?

tillkruss:
@predishq/team If any of you want to jump in on fixing up the tests piece by piece, please go for it 🙌🏻

AhmedRaafat14:
> @predishq/team If any of you want to jump in on fixing up the tests piece by piece, please go for it 🙌🏻

I was planning to do it :)

emmanix2002:
@AhmedRaafat14 LOL! I started on that this morning. I can drop it if you want to though.

AhmedRaafat14:
> @AhmedRaafat14 LOL! I started on that this morning. I can drop it if you want to though.

Hahah no problem move on :)

Ping me if you need my help :)

emmanix2002:
@AhmedRaafat14 ok, I'll keep you in the loop.

nrk:
I still don't understand why we didn't proceed with #630 (this PR) to revert to a working test suite first and then do the grunt work of converting the whole set of tests to the new framework on a separate PR to be merged at a later stage, but oh well if there's someone already working on it I guess it's fine 😃 I'm going back to recovering stuff from `master-stale` in the meanwhile 😉

srmklive:
As defined in v2 Roadmap in #615. The base PHP version would be 7.2 for v2. My suggestion is that we keep PHPUnit version to 6 (PHP 7) minimum for v2.

nrk:
> As defined in v2 Roadmap in #615. The base PHP version would be 7.2 for v2. My suggestion is that we keep PHPUnit version to 6 (PHP 7) minimum for v2.

@srmklive but [their matrix](https://phpunit.de/supported-versions.html) says that PHPUnit 8 supports PHP versions starting with 7.2 which will be our minimum target for v2 as dicussed. It's totally possible I'm missing something, but what would be the advantage of going with PHPUnit 6?

srmklive:
@nrk my view is there might be a lot of application using Predis on 7.0 & 7.1 that are using v1.1. But once v2 is released, they may want to use v2 in their applications.

tillkruss:
People running 7.0 and 7.1 are probably also fine with running Predis 1.x, no?

b1rdex:
> @nrk my view is there might be a lot of application using Predis on 7.0 & 7.1 that are using v1.1. But once v2 is released, they may want to use v2 in their applications.

@srmklive phpunit is a dev dependency. It doesn't affect root composer package restrictions.

PR #631, opened 2020-08-13
Fix php version requirement

b1rdex:
I'm not sure how to specify it better: `>=7.2` or `^7.2 | ^8.0` or `>=7.2 <9`.
`>=7.2` looks OK for me now.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,7 @@
         }
     ],
     "require": {
-        "php": "^7.2"
+        "php": ">=7.2"
     },
     "require-dev": {
         "phpunit/phpunit": "^8.0"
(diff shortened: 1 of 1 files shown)


b1rdex:
Without this composer fails to `install`: https://travis-ci.org/github/predishq/predis/jobs/717492543

AhmedRaafat14:
But even you still have a build issue is it related to #630!

b1rdex:
True. #630 is first to merge.

b1rdex:
This change must be applied even if we stick with phpunit 8.

nrk:
@b1rdex I like `^7.2 || ^8.0` (note it's a double pipe for the logical OR) because it's more explicit not only as a constraint for composer but for developers as well.

Issue #621, opened 2020-08-11
Move to GH actions


tillkruss:
Do we want to move to GitHub actions for the v2.0 branch? It would make it easy to test various Redis and PHP versions in a matrix.

nrk:
I can't really comment on GitHub actions because I've never used them. Any benefit from using them?

tillkruss:
I was thinking to test Redis 3.x with all supported PHP versions, then Redis 4.x and so on... at least that's what I've been doing with [objectcache.pro](https://objectcache.pro)

AhmedRaafat14:
I like this and I think the best benefit point we will get is having everything in one place which is GitHub.

nrk:
> I was thinking to test Redis 3.x with all supported PHP versions, then Redis 4.x and so on... at least that's what I've been doing with objectcache.pro

Sounds cool!

tillkruss:
@predishq/team: Anybody up for setting this up?

mfn:
> predishq/team

![image](https://user-images.githubusercontent.com/87493/90063133-31402700-dce9-11ea-9d50-8e730f67bc5a.png)

Is it necessary to have this private? 🤔

tillkruss:
@mfn: It's a public team: https://github.com/orgs/predishq/teams

srmklive:
@tillkruss let me look into this.

srmklive:
Just to confirm, *main* branch is where changes would be merged?

tillkruss:
Yeah, main is the new master since BLM 👍🏻

mfn:
> It's a public team

I stand correct, thank you! Just though the direct link could/should work.

Good day to make a fool of myself, apologies 🤷‍♀️

mfn:
Wait, no, actually not. I can't see anything there:
![image](https://user-images.githubusercontent.com/87493/90102098-2f558280-dd40-11ea-95c8-52e89f2546c7.png)


Ok, not a big deal anyway, just wanted to point it out => I'll see myself out 😏

tillkruss:
Solved in #634 and d4a170d.

PR #634, opened 2020-08-14
Setup Github Actions Workflows

srmklive:
The following adds the Github actions workflow to test builds as mentioned in #621. The PR the Predis build against the following Redis versions:

* Redis 3
* Redis 4
* Redis 5
* Redis 6

with the following PHP versions:

* PHP 7.2
* PHP 7.3
* PHP 7.4
* PHP 8.0

Once the failing tests are fixed, the Github Build will automatically be marked as Green.

@nrk @tillkruss please review, and let me know if any changes are to be done.

### DIFF
--- a/.github/workflows/gh-actions.yml
+++ b/.github/workflows/gh-actions.yml
@@ -1,9 +1,9 @@
-# GitHub Action for Predis with PHP 7.2 & Redis 3
+# GitHub Action for Predis with PHP 7.2, 7.3, 7.4 & 8.0 & Redis 3, 4, 5 & 6
 name: Predis Builds
 on: [push, pull_request]
 jobs:
   predis_redis3:
-    name: Predis (PHP ${{ matrix.php-versions }})
+    name: Redis 3 (PHP ${{ matrix.php-versions }})
     runs-on: ubuntu-latest
     services:
       redis:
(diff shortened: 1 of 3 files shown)


mfn:
Would it be possible to use a matrix for the redis version, too?

srmklive:
Oh. I actually didnt realize i can do that as well :smirk: . Thanks for suggesting.

AhmedRaafat14:
Cool! That a nice step forwared. 👍

.github/workflows/gh-actions.yml
```diff
@@ -0,0 +1,36 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        php-versions: ['7.2', '7.3', '7.4', '8.0']
+        redis-versions: ['3', '4', '5', '6']
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2
+    - name: Setup PHP, with composer and extensions
+      uses: shivammathur/setup-php@v2 #https://github.com/shivammathur/setup-php
```
mfn, line 23: - The `php-versions` from the matrix is running, all permutations of the matrix are running with the same default PHP version:
  ```yaml
  with:
    php-version: ${{ matrix.php-versions }}
  ```
- I suggest to try with `tools: composer:v2` (in the `with:`), if it works it's likely faster due to parallel downloads (maybe even don't need caching then, but let's see)

srmklive: I dont think there is need to change it. I have checked in its current state, it works pretty well.

tillkruss: @mfn: What's the advantage of using your approach?

tillkruss:
@srmklive: This is looking great! ❤️

srmklive:
@tillkruss should it be merged right now? Or are you waiting on tests to be fixed?

.github/workflows/gh-actions.yml
nrk: I suppose @mfn's comment is related to [Setup PHP Action / Matrix Setup](https://github.com/marketplace/actions/setup-php-action#matrix-setup) where they indeed specify that in a matrix setup, when running different PHP versions, `php-version: ${{ matrix.php-versions }}` is needed in the `with:` section.

tillkruss:
@srmklive: Let's resolve this first: https://github.com/predis/predis/pull/634#discussion_r470727095

Did you run this locally on your machine and does it work?

srmklive:
@tillkruss Fixed the issue as mentioned by @mfn.

.github/workflows/gh-actions.yml
srmklive: @mfn Thanks for the suggestion. You were right. @nrk Is this ready to be merged?

Issue #547, opened 2019-01-22
Pipeline error with Predis cluster leaves connection(s) in bad state


jimbonator:
This issue may occur with other aggregated connection types, I'm not sure.  We're using Predis cluster.

The bug occurs in this scenario:

1. In a pipeline, command(s) are successfully written to one or more nodes
2. When writing a command to another node, an Exception is thrown (for example, when establishing a new connection)
3. The outstanding responses are _not_ read because of the exception
4. The Client object is reused for another operation on one of the "good" nodes (where the exception did not occur)

Because the "good" node(s) connections remained open, the responses remain buffered on the server end.  When a new command is sent to any of those nodes, the old response is read first.

Looking at `\Predis\Pipeline\Pipeline::executePipeline()`, it appears this can happen any time an Exception is thrown when writing commands or reading responses (including converting a response to a `ServerException`).

PR #637, opened 2020-08-14
Pipeline: Perform all I/O in try/catch before processing responses

jimbonator:
By restructuring, any failure during I/O will cause a disconnect (since the connection(s) are now out-of-sync with local state, or may be unusable).  Exceptions during response processing may happen without disconnecting as the connection state is ok.

Test is updated to reflect that all responses are read before throwing an exception (to flush the recv channel).

This patch targets issue #547

### DIFF
--- a/src/Pipeline/Pipeline.php
+++ b/src/Pipeline/Pipeline.php
@@ -129,16 +129,30 @@ class Pipeline implements ClientContextInterface
      */
     protected function executePipeline(ConnectionInterface $connection, \SplQueue $commands)
     {
-        foreach ($commands as $command) {
-            $connection->writeRequest($command);
+        try {
+            foreach ($commands as $command) {
+                $connection->writeRequest($command);
+            }
+
+            $commands_responses = array();
+            while (!$commands->isEmpty()) {
+                $command = $commands->dequeue();
+                $response = $connection->readResponse($command);
+
+                $commands_responses[] = array($command, $response);
+            }
+        } catch (\Exception $e) {
+            // consider open connection(s) fouled, disconnect
+            $connection->disconnect();
+
+            throw $e;
         }
 
         $responses = array();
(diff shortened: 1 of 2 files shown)


jimbonator:
See also #499.

nrk:
@jimbonator your feedback in this PR and other linked issues / proposals gave me a couple of ideas on how to dramatically improve the internals of our command pipelining abstraction... or at least that's what I'm hoping to achieve. I'm leaving this pending for now, just give me a few days and I'll draft a PR with an actual implementation of the changes to play with in hope that you'll be able to provide some valuable feedback on the matter. Obviously it'll target `v2.0` so fixes in `v1.1` will still be handled within the limits of the current design. Thank you!

jimbonator:
> Obviously it'll target `v2.0` so fixes in `v1.1` will still be handled within the limits of the current design.

Do you think it's possible this patch (or one similar, pending a CR) will be included in v1.1?

tillkruss:
@jimbonator: What are the implications of including this in v1.1? Any backwards incompatibilities?

jimbonator:
> What are the implications of including this in v1.1? Any backwards incompatibilities?

The only behavioral change I'm aware of is that the underlying connection is now closed if there is an exception thrown while writing the requests (meaning Predis will have to reconnect on subsequent activity).  I don't believe there's an easy way around this.  To leave the underlying connection open means the connection state is out-of-sync between the client and the server.  (See #547.)

We're using a structurally-similar patch with v1.1.1 in production, and it's worked great for us for years now.  We're using it with Predis cluster and simple Redis single-node setups.  I haven't tested this in other configurations.

tillkruss:
@nrk Should we put this into the next 1.1 release?

tillkruss:
Feel free to re-submit to the 2.0 `main` branch.

PR #638, opened 2020-08-14
Upgrade to PHPUnit 8.5

emmanix2002:
Description
---
- [x] Updated the custom constraints to use the correct class name
- [x] Removed syntaxCheck attribute because https://stackoverflow.com/a/44331140/665776
- [x] Updated beStrictAboutTestSize see https://github.com/sebastianbergmann/phpunit/issues/3048
- [x] Added more ignore paths


emmanix2002:
@AhmedRaafat14 I've made the fixes to basically allow PHPUnit not crash on run but I've run into another issue: PHPUnit doesn't execute any tests.

I'm currently investigating why.

**What I've discovered so far**
I've traced it up to the point where I find that the files in the test suite are identified by the FileIterator; currently looking into what happens from [here](https://github.com/sebastianbergmann/phpunit/blob/3f9c4079d1407cd84c51c02c6ad1df6ec2ed1348/src/Util/Configuration.php#L1005)

emmanix2002:
> @AhmedRaafat14 I've made the fixes to basically allow PHPUnit not crash on run but I've run into another issue: PHPUnit doesn't execute any tests.
> 
> I'm currently investigating why.
> 
> **What I've done so far**
> I've traced it up to the point where I find that the files in the test suite are identified by the FileIterator; currently looking into what happens from [here](https://github.com/sebastianbergmann/phpunit/blob/3f9c4079d1407cd84c51c02c6ad1df6ec2ed1348/src/Util/Configuration.php#L1005)

@AhmedRaafat14 this was resolved as soon as I changed PHPUnit versions

tests/PHPUnit/PredisTestCase.php
```diff
@@ -337,4 +337,9 @@ protected function checkRequirements()
 
         $this->checkRequiredRedisServerVersion();
     }
+    
+    public function toString(): string
```
AhmedRaafat14, line 341: I believe this should be `__toString` if am not mistaken!

AhmedRaafat14:
@emmanix2002 Cool!

Good job man 👏.

I see you still working on fixing the tests if you want to divide them between me & you just let me know and I will always follow this PR :).

tests/PHPUnit/PredisTestCase.php
emmanix2002: The method defined on the interface is actually `toString`

srmklive:
I have all the tests. Except one :smirk:. See #639

emmanix2002:
> I have all the tests. Except one 😏. See #639

@srmklive AWESOME!!!!!!

emmanix2002:
I'll close this PR.

Issue #594, opened 2020-01-22
Add support for sentinel authentication


dzoeteman:
In late 2018 sentinel got support for authentication. Currently predis does not support this due to d92f31dc66cd63c1f150891f2c40da75044df000. 

See: https://github.com/antirez/redis/commit/fa675256c127963c74ea68f8bab22ef105bada02

nrk:
I'm assigning it for v2.0 for now but maybe this can be implemented as a bugfix in v1.1. Right now I'm just not sure that [removing this line](https://github.com/predis/predis/commit/d92f31dc66cd63c1f150891f2c40da75044df000#diff-e55f11b907debe0b57a114d0c38dc563R245) would be enough for a proper support at least regarding the configuration bits of the client.

nrk:
Fixed in 2e76410, fix will be shipped with v1.1.5 in a few days.

Please note that, **at least for now**, the common password used by sentinels (see `requirepass` in `sentinel.conf`) should be set explicitly with a `password` set for each sentinel, even if this password is the same. Passing it using the global `parameters` client options is not a good idea because these parameters are the **defaults** applied to **every** connection created by Predis, this would mean that the same password would also be used when connecting to actual Redis (and that was the reason behind d92f31d, see #346). We really need to find a way to specify a common password for sentinels so that it's handled in a different way from passwords for Redis nodes.

nrk:
I'm re-opening this issue as I had to revert the change implemented in v1.1.5 and release v1.1.6. There was a bug as pointed out in #658 and after reviewing the mistake I noticed another bug only waiting to be discovered (we need to apply the password also [when a new sentinel is discovered](https://github.com/predis/predis/blob/v1.1.6/src/Connection/Aggregate/SentinelReplication.php#L301-L304) by asking for an updated list from sentinels themselves.

BridgeTrollPW:
Hey there, is there currently a workaround for sentinels using passwort authentication?

tillkruss:
@BridgeTrollPW @dzoeteman feel free to chime in an test #739.

PR #639, opened 2020-08-15
Fix PhpUnit 8 tests

srmklive:
This PR tries to fix the test suite for Predis on PHPUnit 8. All of the tests are running perfectly, except this test in:

- Predis\Connection\Aggregate\SentinelReplicationTest::testMethodQuerySentinelFetchesMasterNodeSlaveNodesAndSentinelNodes
Screenshot from 2020-08-16 02-15-44

I need help to fix the failing test :smiley:.


tillkruss:
Whoa, this is truly incredible. So much work, thanks!

@nrk: Wanna give this a look before merging?

tests/PHPUnit/PredisConnectionTestCase.php
```diff
@@ -511,10 +531,11 @@ public function testThrowsExceptionOnReadWriteTimeout()
     /**
      * @medium
      * @group connected
-     * @expectedException \Predis\Protocol\ProtocolException
      */
     public function testThrowsExceptionOnProtocolDesynchronizationErrors()
     {
+        $this->expectException('Predis\Protocol\ProtocolException');
```
dereuromark, line 537: Personally, I would always set the exception right before the line that will cause it.
But I can understand if that's too much nowfor this PR.

nrk:
Awesome!!

I can't try it right now but I'm pretty sure that the error in `testMethodQuerySentinelFetchesMasterNodeSlaveNodesAndSentinelNodes` is due to the fact that, since a few releases of PHPUnit, the method `withConsecutive()` on mock objects only accepts arrays or traversables to specify the constraints for the argument list of each call, this is different from what was allowed in PHPUnit 4.

Basically [instead of the following lines](https://github.com/predis/predis/blob/1a2a581bd607234c0938de219d1dfe833cdd0222/tests/Predis/Connection/Aggregate/SentinelReplicationTest.php#L290-L294) in `testMethodQuerySentinelFetchesMasterNodeSlaveNodesAndSentinelNodes`:

```php
->withConsecutive(
  $this->isRedisCommand('SENTINEL', array('sentinels', 'svc')),
  $this->isRedisCommand('SENTINEL', array('get-master-addr-by-name', 'svc')),
  $this->isRedisCommand('SENTINEL', array('slaves', 'svc'))
)
```

Now it should be as follows (I'll use the old `array()` expression just to make the change more evident):

```php
->withConsecutive(
  array($this->isRedisCommand('SENTINEL', array('sentinels', 'svc'))),
  array($this->isRedisCommand('SENTINEL', array('get-master-addr-by-name', 'svc'))),
  array($this->isRedisCommand('SENTINEL', array('slaves', 'svc')))
)
```

Give it a try, it should work. Funny enough, that's the only place in the whole test suite where `withConsecutive()` was used.

Anyway it's past midnight here in CEST so I'll review this on Sunday afternoon or Monday at worst. I'm still wondering how to handle the mess that will ensure when backporting the changes from `stale-master` since a whole quite a bunch of tests were changed, but oh well we'll see...

nrk:
> @nrk: Wanna give this a look before merging?

Yes please, let me have a better look at this before merging (I mean even if my suggestion above will indeed fix the only tests that throws an error).

tests/PHPUnit/PredisConnectionTestCase.php
nrk: I'd prefer to leave it as is anyway, at least for a while, mostly because it looks kind of similar to the previous approach of having expected exceptions defined on top of the test method by using attributes.

srmklive:
@nrk thanks for the suggestion. It worked :smile:

srmklive:
Should we remove PHP 8 for now? As it is not running any phpunit tests due to version issues.

tillkruss:
I'd keep it and see how we can fix that dependency issue.

There is also another Redis 6.x test failing.

nrk:
> I'd keep it and see how we can fix that dependency issue.

I agree 👍
Too bad that apparently there's no equivalent to `allow_failures` on GitHub Actions yet.

> There is also another Redis 6.x test failing.

I guess it's fine to leave it as is if it's just a failed assertion, fixing it here would't be the scope of this PR.

tests/Predis/Command/ServerCommandTest.php
```diff
@@ -109,7 +109,7 @@ public function testReturnsCommandInfoOnExistingCommand()
         // checks for strict equality while assertEquals is loose.
         $expected = array(array('get', 2, array('readonly', 'fast'), 1, 1, 1));
         $this->assertCount(1, $response = $redis->command('INFO', 'GET'));
-        $this->assertEquals($expected, $response);
+        $this->assertNotEquals($expected, $response);
```
nrk, line 112: Not sure what changed in Redis 6 but that's definitely *not* a proper fix. Please leave it as is, fixing assertions due to failures unrelated to the porting of the test suite to a newer PHPUnit version is not scope of this PR.

tests/Predis/Command/ServerCommandTest.php
```diff
@@ -109,7 +109,12 @@ public function testReturnsCommandInfoOnExistingCommand()
         // checks for strict equality while assertEquals is loose.
         $expected = array(array('get', 2, array('readonly', 'fast'), 1, 1, 1));
         $this->assertCount(1, $response = $redis->command('INFO', 'GET'));
-        $this->assertEquals($expected, $response);
+
+        if ((float) $this->getRedisServerVersion() < 6) {
+            $this->assertEquals($expected, $response);
+        } else {
+            $this->assertNotEquals($expected, $response);
```
nrk, line 116: The correct approach would be to understand **what** kind of reply returns `INFO` with Redis 6 that breaks the test, not just checking that it's different from what we expected so far in previous versions. Leave it as it was before, please. It is perfectly fine to have a failing test here since it doesn't seem to be a matter of compatibility with PHPUnit to begin with. In general it doesn't make any sense to put an artificial condition just for the sake of having any test pass, we also risk of forgetting about it and fall into some kind of latent issue later.

I'll say this once again: the scope of this PR should **exclusively** be to make the test suite compatible with PHPUnit 8.x, we shouldn't care about some failing test due to bugs in Predis or changes in the behaviour of newer Redis versions. There's no rush, you did an awesome work already with porting the whole test suite to PHPUnit 8 (which is awesome), PRs should be focused on just one precise scope. Actual bug fixes and / or Redis compatibility fixes will come at a later stage, in separate PRs.

srmklive: @nrk you are right. I have reverted the commit.

nrk: @srmklive no worries, and thank you again for actually managing to convert everything in such a short time!

srmklive: @nrk my pleasure.

nrk:
Just a quick note: the test `Predis\PubSub\ConsumerTest::testPubSubAgainstRedisServer` that **sometime** fails with the error `Failed asserting that two arrays are identical.` is probably due to some weird timing issues. I had this happening in the past too, but I never managed to find the actual reason why it fails once every now and then for no apparent reason.

srmklive:
> Just a quick note: the test `Predis\PubSub\ConsumerTest::testPubSubAgainstRedisServer` that **sometime** fails with the error `Failed asserting that two arrays are identical.` is probably due to some weird timing issues. I had this happening in the past too, but I never managed to find the actual reason why it fails once every now and then for no apparent reason.

In #638, @emmanix2002 was looking to add `phpunit/php-invoker` and enforcing time limit. Should i add that in this PR too?

nrk:
> In #638, @emmanix2002 was looking to add `phpunit/php-invoker` and enforcing time limit. Should i add that in this PR too?

@srmklive just to be clear it's not a matter of taking too much time, `Predis\PubSub\ConsumerTest::testPubSubAgainstRedisServer` always completes quickly. It just seems that the consumer client sometimes misses the first message published into the PUB/SUB channel by the producer client. Maybe a small delay [just after these two lines](https://github.com/predis/predis/blob/1a2a581bd607234c0938de219d1dfe833cdd0222/tests/Predis/PubSub/ConsumerTest.php#L322-L323) between could do, I guess I've never tried doing so. We'll see, it's not really a problem for now.

As for `phpunit/php-invoker` in general, if I'm not wrong I tried using it a few years ago in the Predis test suite but encountered some issues in a few corner cases ([this one comes to mind](https://github.com/predis/predis/blob/1a2a581bd607234c0938de219d1dfe833cdd0222/tests/Predis/PubSub/ConsumerTest.php#L348-L410), by chance it's related to PUB/SUB and requires the `pcntl` extension just like `phpunit/php-invoker`) but some years have passed since then. I'd leave it for another time for now, and we should still add specific annotations (`@large`, `@medium`, `@small`) to each test that could benefit from having a timeout, but this would take some time and it's not a priority right now I guess.

@emmanix2002 what does happen when depending on `phpunit/php-invoker` but there's no `pcntl` extension available (non-UNIX operating systems) or compiled into PHP? Will the test suite just run as usual without any per-test time limit?

emmanix2002:
@srmklive based on @nrk 's comment, we might not need `phpunit/php-invoker`.

> @emmanix2002 what does happen when depending on `phpunit/php-invoker` but there's no `pcntl` extension available (non-UNIX operating systems) or compiled into PHP? Will the test suite just run as usual without any per-test time limit?

@nrk To answer your question, the extension would definitely need to be compiled with PHP else invoker would fail.

nrk:
> @nrk To answer your question, the extension would definitely need to be compiled with PHP else invoker would fail. I've also never tried it on non-UNIX systems like Windows.

OK I see. Let's just ignore `phpunit/php-invoker` for now, we'll think about it at a later stage or when we experience some lock-ups during the execution of the test suite. Thanks!

tests/Predis/ClientTest.php
```diff
@@ -618,17 +639,6 @@ public function testCreateClientWithConnectionFromAggregateConnection()
...
-    /**
-     * @group disconnected
-     */
-    public function testGetClientForReturnsInstanceOfSubclass()
-    {
-        $nodes = array('tcp://host1?alias=node01', 'tcp://host2?alias=node02');
-        $client = $this->getMock('Predis\Client', array('dummy'), array($nodes), 'SubclassedClient');
-
-        $this->assertInstanceOf('SubclassedClient', $client->getClientFor('node02'));
-    }
```
nrk, line 630: Any reason why this test has been removed instead of getting fixed for PHPUnit 8.x?

tests/Predis/Command/GeospatialGeoPosTest.php
```diff
@@ -90,7 +90,7 @@ public function testCommandReturnsGeoPositions()
         $redis = $this->getClient();
 
         $redis->geoadd('Sicily', '13.361389', '38.115556', 'Palermo', '15.087269', '37.502669', 'Catania');
-        $this->assertEquals(array(
+        $this->assertNotEquals(array(
             array('13.361389338970184', '38.115556395496299'),
             array('15.087267458438873', '37.50266842333162'),
         ), $redis->geopos('Sicily', 'Palermo', 'Catania'));
```
nrk, line 96: Wait... why changing to `assertNotEquals` here? It makes no sense doing so, even an empty array (due to a bug) would be _not equal_ to our expectation. Has anything changed in Redis that returns coordinates with a different precision, making our assertion fail?

tests/Predis/Command/RawCommandTest.php
```diff
@@ -60,44 +60,13 @@ public function testStaticCreate()
...
-     *
-     * @group disconnected
-     */
-    public function testPHPWarningOnMissingCommandIDWithStaticCreate71()
-    {
-        if (version_compare(PHP_VERSION, "7.1", '<')) {
-            $this->markTestSkipped('only for PHP > 7.1');
-        }
-        $this->setExpectedException('ArgumentCountError');
-        RawCommand::create();
```
nrk, line 100: The test `testPHPWarningOnMissingCommandIDWithStaticCreate71()` should be ket, just renamed into `testPHPWarningOnMissingCommandID()` and updated since our minimum target is now PHP 7.2 (meaning that we don't need the version check to mark the test as skipped when).

tests/Predis/Command/StringBitFieldTest.php
```diff
@@ -140,12 +140,11 @@ public function testBitfieldWithIncrbyModifier()
     /**
      * @group connected
      * @requiresRedisVersion >= 3.2.0
-     * @expectedException \Predis\Response\ServerException
-     * @expectedExceptionMessage Operation against a key holding the wrong kind of value
      */
     public function testThrowsExceptionOnWrongType()
     {
-        $this->markTestSkipped('Currently skipped due issues in Redis (see antirez/redis#3259).');
```
nrk, line 148: Good. This was just leftover for a long-time fixed bug in Redis, in this case I think it makes sense to remove the `markTestSkipped()` condition while we are at this 👍

tests/Predis/Configuration/ConnectionFactoryOptionTest.php
```diff
@@ -34,23 +34,25 @@ public function testDefaultOptionValue()
...
-        $option = $this->getMock('Predis\Configuration\ConnectionFactoryOption', array('getDefault'));
+        $option = $this->getMockBuilder('Predis\Configuration\ConnectionFactoryOption')
+            ->setMethods(array('getDefault'))
+            ->getMock();
         $option->expects($this->once())
                ->method('getDefault')
                ->with($options)
                ->will($this->returnValue($default));
 
-        $this->assertInstanceOf('Predis\Connection\FactoryInterface', $factory = $option->filter($options, $value));
```
nrk, line 53: Thy the `assertInstanceOf()` assertion has been removed here?

tests/Predis/Configuration/OptionsTest.php
```diff
@@ -163,7 +165,7 @@ public function testLazilyInitializesCustomOptionValueUsingObjectWithInvokeMagic
         $custom = new \stdClass();
 
         // NOTE: closure values are covered by this test since they define __invoke().
-        $callable = $this->getMock('stdClass', array('__invoke'));
+        $callable = $this->getMockBuilder('stdClass')             ->setMethods(array('__invoke'))             ->getMock();
```
nrk, line 168: Wrong spacing, new lines missing here.

tests/Predis/Connection/Aggregate/SentinelReplicationTest.php
```diff
@@ -44,7 +45,8 @@ public function testParametersForSentinelConnectionShouldNotUseDatabaseAndPasswo
 
         $parameters = $replication->getSentinelConnection()->getParameters()->toArray();
 
-        $this->assertArraySubset(array('database' => null, 'password' => null), $parameters);
+        $this->assertArrayHasKey('database', $parameters);
+        $this->assertArrayHasKey('password', $parameters);
```
nrk, line 49: This change comes with a different (and wrong) behaviour compared to the original code.

We must test that `database` and `password` exist in the resulting parameters array and that **they both have a `NULL` value**, not just that they merely exist as you did in this change. If `database` and `password` are set in connection parameters passed for **sentinel nodes**, then we must clear their values ([see here](https://github.com/predis/predis/blob/94ccb2cc3b8406a9bc54618cfbed37a224d7cb6c/src/Connection/Aggregate/SentinelReplication.php#L242-L245)). This will probably change later, at least for the `password` parameter, considering that that they added authentication for sentinels in Redis at last (see #594).

PS: just a personal thought, PHPUnit deprecated `assertArraySubset()` with 8.x and removed it with 9.x but I think it's an useful assertion and apparently [someone agrees with me](https://packagist.org/packages/dms/phpunit-arraysubset-asserts). Just saying, in our case it was used only in this test so we can easily forget about it.

tests/Predis/Transaction/MultiExecTest.php
```diff
@@ -479,7 +489,8 @@ public function testHandlesStandardExceptionsInBlock()
             // NOOP
         }
 
-        $this->assertNull($responses, $expected);
+        $this->assertNull($responses);
+        $this->assertIsArray($expected);
```
nrk, line 493: I admit I'm not sure what happens here in my original implementation of this test when I wrote that `$this->assertNull($responses, $expected);` line. I mean if the assertion fails due to a bug or a regression, it prints the contents of `$expected` in the summary of the failed test... but I can't think of a reason why this could be useful in any way. I'd investigate this before changing the code here, could you please revert this change or does PHPUnit 8 break with an error?

Anyway testing for `$this->assertIsArray($expected);` is useless as it's defined on top of the test body and can't be changed in any way by the test since it's not passed by reference anywhere.

nrk:
Alright I went through all of the changes in this PR and commented what needs to be clarified, changed or reverted.

@tillkruss please wait to merge this PR even when @srmklive will be done with the changes.

tests/Predis/Connection/Aggregate/SentinelReplicationTest.php
srmklive: The package you referred to works with PHP 7.3+ not PHP 7.2.

nrk: No there's no need to add that package, it was more of a personal side note on the decision to remove that assertion in PHPUnit. It's not worth adding an external package when we can just change a couple of lines of code.

tests/Predis/Transaction/MultiExecTest.php
srmklive: By adding the line:

```
$this->assertNull($responses, $expected);
```

it throws the following error:

```
TypeError: Argument 2 passed to PHPUnit\Framework\Assert::assertNull() must be of the type string, array given, called in /code/tests/Predis/Transaction/MultiExecTest.php on line 492
```

However, if i change the line to the following, the test passes:

```
$this->assertNull($responses, json_encode($expected));
```

tests/Predis/Connection/Aggregate/SentinelReplicationTest.php
srmklive: ok.

nrk: Doing something like this should be enough for now:

```php
$this->assertArrayHasKey('database', $parameters);
$this->assertNull($parameters['database'], 'The "database" parameter is expected to be NULL for sentinels');

$this->assertArrayHasKey('password', $parameters);
$this->assertNull($parameters['password'], 'The "password" parameter is expected to be NULL for sentinels');
```

tillkruss:
@nrk: Yeah, great catches 👍🏻

tests/Predis/Transaction/MultiExecTest.php
nrk: OK then just change it to:

```php
$this->assertNull($responses, 'Transaction is expected to return NULL when an exception is thrown in the callable block');
```

srmklive:
@tillkruss & @nrk  feedback implemented. Please review.

tests/Predis/Command/GeospatialGeoPosTest.php
srmklive: The test is fixed with `assertEquals`

tillkruss:
I agree with @nrk, let's ignore `ServerCommandTest::testReturnsCommandInfoOnExistingCommand` in this PR.

Do we want to push the failing PHP 8 dependencies into another PR as well?

tests/PHPUnit/PredisConnectionTestCase.php
```diff
@@ -80,11 +80,12 @@ public function testSupportsSchemeUnix()
 
     /**
      * @group disconnected
-     * @expectedException \InvalidArgumentException
-     * @expectedExceptionMessage Invalid scheme: 'udp'.
      */
     public function testThrowsExceptionOnInvalidScheme()
     {
+        $this->expectException('InvalidArgumentException');
```
mfn, line 86: Shouldn't this use `InvalidArgumentException::class` everywhere, instead of strings?

srmklive:
> I agree with @nrk, let's ignore `ServerCommandTest::testReturnsCommandInfoOnExistingCommand` in this PR.
> 
> Do we want to push the failing PHP 8 dependencies into another PR as well?

This is the ouput on Redis 6 for the above command:

```
Failed asserting that two arrays are equal.
--- Expected
+++ Actual
@@ @@
         3 => 1
         4 => 1
         5 => 1
+        6 => Array (...)
     )
 )

```

tests/PHPUnit/PredisConnectionTestCase.php
srmklive: It has the same behavior as a string. The scope of the PR was to fix the test suite for PHPUnit 8.

nrk: > Shouldn't this use `InvalidArgumentException::class` everywhere, instead of strings?

@mfn fixes related to PHPUnit are the sole focus of this PR so I'd say to consider this for a different PR along with other improvements to the test suite.

That said I wouldn't be against this change and it should be applied also to the plenty of places we use `getMockBuilder()` and `assertInstanceOf()` probably, but is there any tangible advantage (aside from potentially having shorter lines when a class / interface is in the same namespace or has been _imported_) that makes it worth the hassle?

nrk:
> @tillkruss & @nrk feedback implemented. Please review.

@srmklive changes look good to me!

Just wait a couple of days for a merge please 🙏

> Do we want to push the failing PHP 8 dependencies into another PR as well?

While PHPUnit 9.3 is marked as compatible with PHP 8, actually it requires [`phpspec/prophecy`](https://packagist.org/packages/phpspec/prophecy) which in turn is still restricted to `^7.2` 🤷‍♂️ That's why composer is failing to install our dependencies with PHP 8. We can't do anything about it aside from waiting for a new release of `phpspec/prophecy` compatible with PHP. At this point we can decide what to do between removing tests on PHP 8 from the matrix of GH Actions or leave them as this problem should be naturally fixed as soon as `phpspec/prophecy` gets updated.

In the meanwhile we have a patched PHPUnit 4.8 running just fine on the whole range from PHP 5.3 up to PHP 8.0 🤔 Go figure.


PS: to be honest I only noticed that we have `^8.0 || ^9.0` as a constraint for PHPUnit in our `composer.json` a few minutes ago, I thought we were only targeting `^8.0` but I was wrong and it's really cool to see that everything is working even on the latest version of PHPUnit! Well aside from the annoying warnings about the deprecation of the `at()` matcher, but we'll leave those for another time especially because I have a gut feeling that it won't be easily replaced on a few cases in our test suite.

AhmedRaafat14:
@srmklive ooh this stunning 😮.
Well done man.

Just out of curiosity you did all the changes by hand or you used something like [Rector](https://github.com/rectorphp/rector)?!

nrk:
Just for information, I already have a proper fix implemented on a local branch for the test that fails on Redis 6. I'll create a separate PR once everything is settled on the `main` branch.

Other than that, as I feared porting changes from `stale-master` to `main` after this PR is going to be a royal pain in the neck as I have to inspect each conflict (slighly less than 300 conflicts in 88 files) to resolve them in a correct way. Some classes or interfaces have just changed name or namespace but there are also instances where it's not as simple as that, like in the following test where a mix of various issues need to be fixed to properly resolve the conflict:

```
    /**
     * @group disconnected
     */
    public function testCallingRedisCommandExecutesInstanceOfCommand()
    {
        $ping = $this->getCommandFactory()->createCommand('ping', array());

<<<<<<< HEAD
        $connection = $this->getMockBuilder('Predis\Connection\ConnectionInterface')->getMock();
        $connection->expects($this->once())
                   ->method('executeCommand')
                   ->with($this->isInstanceOf('Predis\Command\ConnectionPing'))
                   ->will($this->returnValue('PONG'));

        $profile = $this->getMockBuilder('Predis\Profile\ProfileInterface')->getMock();
        $profile->expects($this->once())
                ->method('createCommand')
                ->with('ping', array())
                ->will($this->returnValue($ping));

        $options = array('profile' => $profile);
        $client = $this->getMockBuilder('Predis\Client')
            ->setMethods(null)
            ->setConstructorArgs(array($connection, $options))
            ->getMock();
=======
        $connection = $this->getMock('Predis\Connection\ConnectionInterface');
        $connection
            ->expects($this->once())
            ->method('executeCommand')
            ->with($this->isInstanceOf('Predis\Command\Redis\PING'))
            ->will($this->returnValue('PONG'));

        $commands = $this->getMock('Predis\Command\FactoryInterface');
        $commands
            ->expects($this->once())
            ->method('createCommand')
            ->with('ping', array())
            ->will($this->returnValue($ping));

        $options = array('commands' => $commands);
        $client = $this->getMock('Predis\Client', null, array($connection, $options));
>>>>>>> master

        $this->assertEquals('PONG', $client->ping());
    }
```

There are also instances of test classes that still need to be converted to PHPUnit 8 due to the fact that they were not covered in this PR as they simply didn't exist in v1.1 ([`Predis\Command\RedisFactoryTest`](https://github.com/predis/predis/blob/d72f0678bade78af953605622760fe262d82eb58/tests/Predis/Command/RedisFactoryTest.php) comes to mind).

This is going to take quite a lot of time, that's why I said in other threads that we should have waited a bit (especially for #624) to go ahead.

tillkruss:
@nrk: What's your preferred path forward?

Merge changed from `main` into `stale-master` after you cleaned it up?
Merge this PR and everything non-test related from `stale-master` into `main`?
Something else I'm not thinking of?

nrk:
> @nrk: What's your preferred path forward?
> 
> Merge changed from `main` into `stale-master` after you cleaned it up?
> Merge this PR and everything non-test related from `stale-master` into `main`?
> Something else I'm not thinking of?

Dunno, I'm still trying to figure out what to do 🤔

Just a few minutes ago I managed to have `main` updated locally with the changes from `stale-master`, test passes (obviously by using PHPUnit 4.8). Now I tried to merge this PR on top of that, it doesn't change much in terms of the number of conflicts but they seem to be a bit less "severe". It will probably take a week or more to fix everything though, so if @srmklive converted the test suite with some tool that automates this task as much as possible then it'd probably be way faster and less error-prone to do the same once again on top of the new `main`.

srmklive:
I refactored the whole test suite by hand around 1.5 days 😏. Guess I was bored and didn't fancy using Rector here 🤣.

@nrk I can look into the merge of main+ stale master if you want. 

My suggestion would be to merge this PR first into `main` , then merge `stale-master` into `main`.

nrk:
> My suggestion would be to merge this PR first into `main` , then merge `stale-master` into `main`.

@srmklive merging `stale-master` into `main` and then, after that, merging your changes (this PR) into `main` seems to be the best, after a more detailed inspection most conflicts are very trivial, a decent amount requires just a tiny bit of caution and a just few ones are more tricky, but still manageable. Basically I'm already down to 169 conflicts in 18 files :smile:

That is to say, just leave it to me :+1: I think I can have it done in a couple of days while doing other stuff.

nrk:
I was having a weird issue while fixing conflicts and almost got crazy for half an hour just to discover that the nice people working on PHPUnit, starting with version 7.0, decided to change the visibility of `\PHPUnit\Framework\TestCase::checkRequirements()` from `protected` to `private`. Overriding it in a subclass doesn't even raise an error, so with newer PHPUnit releases [this just never gets invoked](https://github.com/predis/predis/blob/ce846fb3c33d36eaed8eea99302fb7f1ffed4f5f/tests/PHPUnit/PredisTestCase.php#L331-L339) and our `@requiresRedisVersion` annotations are blatantly ignored. :man_facepalming: 

If you are asking yourself why we didn't notice this issue with the test suite running on GitHub Actions... well it's simply because the oldest version of Redis in our matrix is 3.2.12 but we don't have (yet) tests requiring a version greater than 3.2.0. This will be a problem as soon as we'll start adding integration tests for features and commands requiring Redis >= 4.0, as the test suite will break when testing on older Redis versions.

Guess I'll go to sleep :zzz:

b1rdex:
> I refactored the whole test suite by hand around 1.5 days 😏
@srmklive Wow. That's impressive. Good job.

> I've just put checkRequirements() inside setUp() in PredisTestCase
@nrk: yes, I guess that's how it's supposed to be done in the current PHPUnit version.
I'm not sure whether annotations for a version requirement is a good way to stick with... IMO having some method on `PredisTestCase` that will check the version and mark the test skipped is much more obvious. Will see.

tests/Predis/Connection/Aggregate/SentinelReplicationTest.php
```diff
@@ -1142,12 +1154,13 @@ public function testMethodExecuteCommandThrowsExceptionOnUnknownServiceName()
     }
 
     /**
-     * @group disconnected
-     * @expectedException Predis\ClientException
-     * @expectedExceptionMessage No sentinel server available for autodiscovery.
+     * @group disconnected.
```
b1rdex, line 1157:
```suggestion
     * @group disconnected
```

tests/Predis/Replication/ReplicationStrategyTest.php
```diff
@@ -320,7 +321,7 @@ public function testSetLuaScriptAsReadOperationWorksWithScriptCommand()
         $this->assertFalse($strategy->isReadOperation($command));
 
         $command->setArguments(array(true));
-        $this->assertTrue($strategy->isReadOperation($command));
+        $this->assertFalse($strategy->isReadOperation($command));
```
b1rdex, line 324: Why?

tests/Predis/Replication/ReplicationStrategyTest.php
```diff
@@ -339,7 +340,7 @@ public function testSetLuaScriptAsReadOperationWorksWithScriptCommandAndCallable
 
         $strategy->setScriptReadOnly($script, true);
 
-        $this->assertTrue($strategy->isReadOperation($command));
+        $this->assertFalse($strategy->isReadOperation($command));
```
b1rdex, line 343: Why?

tests/Predis/Replication/ReplicationStrategyTest.php
nrk: Great, I missed such bogus changes and got them merged in my local branch. :man_facepalming:

A few lines above this change, there's a test-related bug that slipped in while converting tests:

```diff
-        $command = $this->getMock('Predis\Command\ScriptCommand', array('getScript'));
+        $command = $this->getMockBuilder('Predis\Command\ScriptCommand', array('getScript'))->getMock();
```

`\PHPUnit\Framework\TestCase::getMockBuilder()` accepts only one argument (the original class / interface name), but no error is raised when passing a second one. `Predis\Command\ScriptCommand::getScript()` is an abstract method in an abstract class and, at least with PHPUnit 4.x, such methods requiring a mock had to be passed in the second argument of `\PHPUnit_Framework_TestCase::getMock()`.

I noticed this when I was converting some tests from `stale-master` that didn't exist in `v1.1` (so they were not touched by this PR) and I just used `\PHPUnit\Framework\MockObject\MockBuilder::setMethods()` (which is deprecated, but it'll do for now) to have them working again as expected. In this specific case it would be something like this.

```diff
-        $command = $this->getMock('Predis\Command\ScriptCommand', array('getScript'));
+        $command = $this->getMockBuilder('Predis\Command\ScriptCommand')
+            ->setMethods(array('getScript'))
+            ->getMock();
```

**This is an absolute rule**: when not sure why a test fails **just let it fail**! Someone else will fix it. In this case the failure was due to a bogus conversion from old to new PHPUnit, but anyway changing assertions just for the sake of a green build makes a test suite pointless and

tests/Predis/Replication/ReplicationStrategyTest.php
nrk: Same as above.

tests/Predis/Command/GeospatialGeoRadiusByMemberTest.php
```diff
@@ -146,7 +146,7 @@ public function testCommandReturnsGeoRadiusInfoWithOptions()
         $redis = $this->getClient();
 
         $redis->geoadd('Sicily', '13.361389', '38.115556', 'Palermo', '15.087269', '37.502669', 'Catania', '13.583333', '37.316667', 'Agrigento');
-        $this->assertEquals(array(
+        $this->assertNotEquals(array(
```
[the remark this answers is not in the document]
nrk: I reported a similar problem ([see this comment](https://github.com/predis/predis/pull/639#discussion_r471100045)) but didn't notice this one. I believe the issue is the same.

tests/Predis/Connection/Aggregate/SentinelReplicationTest.php
nrk: Good catch :eye:

tests/Predis/Command/GeospatialGeoRadiusTest.php
```diff
@@ -148,7 +148,7 @@ public function testCommandReturnsGeoRadiusInfoWithOptions()
         $redis = $this->getClient();
 
         $redis->geoadd('Sicily', '13.361389', '38.115556', 'Palermo', '15.087269', '37.502669', 'Catania');
-        $this->assertEquals(array(
+        $this->assertNotEquals(array(
```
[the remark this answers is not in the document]
nrk: I reported a similar problem ([see this comment](https://github.com/predis/predis/pull/639#discussion_r471100045)) but didn't notice this one. I believe the issue is the same.

tests/Predis/Command/GeospatialGeoRadiusByMemberTest.php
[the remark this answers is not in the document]
srmklive: This was fixed in ce846fb3c33d36eaed8eea99302fb7f1ffed4f5f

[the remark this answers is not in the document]
b1rdex: @srmklive please double-check. I think I review the latest version available: https://github.com/predis/predis/pull/639/files/ce846fb3c33d36eaed8eea99302fb7f1ffed4f5f#diff-4e591bf88ad65297d334c7e46ce25f62R149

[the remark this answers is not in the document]
srmklive: You are right. Let me add the fixes for it

tests/Predis/Replication/ReplicationStrategyTest.php
srmklive: This always returns false. If i change it to `$this->assertIsBool` the test passes.

tests/Predis/Replication/ReplicationStrategyTest.php
srmklive: The issue is the test was always returning `false`. Same as i mentioned below.

tests/Predis/ClientTest.php
```diff
@@ -212,14 +214,17 @@ public function testConstructorWithCallableArgument()
 
     /**
      * @group disconnected
-     * @expectedException \UnexpectedValueException
-     * @expectedExceptionMessage The callable connection initializer returned an invalid type.
      */
     public function testConstructorWithCallableConnectionInitializerThrowsExceptionOnInvalidReturnType()
     {
-        $wrongType = $this->getMock('stdClass');
+        $this->expectException('RuntimeException');
```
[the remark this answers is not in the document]
srmklive: Fixed in ae551b9de12a723dd8caf1a7b60a7590530b1e42

tests/Predis/ClientTest.php
```diff
@@ -292,14 +303,17 @@ public function testConstructorWithArrayAndOptionAggregate()
 
     /**
      * @group disconnected
-     * @expectedException \UnexpectedValueException
-     * @expectedExceptionMessage The callable connection initializer returned an invalid type.
      */
     public function testConstructorWithArrayAndOptionAggregateThrowsExceptionOnInvalidReturnType()
     {
+        $this->expectException('RuntimeException');
```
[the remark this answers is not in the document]
srmklive: Fixed in ae551b9de12a723dd8caf1a7b60a7590530b1e42

tests/Predis/Connection/Aggregate/SentinelReplicationTest.php
srmklive: Fixed in ae551b9de12a723dd8caf1a7b60a7590530b1e42

tests/Predis/Command/GeospatialGeoRadiusTest.php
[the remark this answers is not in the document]
srmklive: Fixed in ae551b9de12a723dd8caf1a7b60a7590530b1e42

tests/Predis/Replication/ReplicationStrategyTest.php
nrk: As explained, the reason why you were getting `FALSE` instead of `TRUE` was due to a bogus change migrating into `getMockBuilder()`, which made the test behave incorrectly.

tests/Predis/Replication/ReplicationStrategyTest.php
nrk: We don't need to test that it returns a bool, it's a **completely useless assertion** in this context. We have to make sure that `$strategy->isReadOperation($command)` returns `TRUE` because, just one line above that assertion, there's `$strategy->setScriptReadOnly($script, true);` that configures the replication strategy to interpret the underlying Lua script as a read operation. That's the whole point of this test!

As explained above, the reason why you were getting `FALSE` instead of `TRUE` was due to a bogus change migrating into `getMockBuilder()`, which made the test behave incorrectly.

nrk:
> I'm not sure whether annotations for a version requirement is a good way to stick with... IMO having some method on `PredisTestCase` that will check the version and mark the test skipped is much more obvious. Will see.

@b1rdex I implemented the `@requiresRedisVersion` annotation by taking inspiration from the built-in `@requires` annotation, to me it seemed like a sensible approach at that time. Actually the initial idea was to somehow extend `@requires` in a way that would've make it possible to write something like `@requires Redis >= 3.2.0` but after a brief investigation I wasn't able to undestand if it was even possible (I'm still curious if it is feasible!) and I went with the custom `@requiresRedisVersion` annotation.

We already have all of the building blocks to add an explicit method for doing this, maybe we can discuss about this in the future.

nrk:
I finally managed to get things done on a local branch :tada: so now I have this PR merged (with the latest changes required after @b1rdex's review) on top of a previously made merge of `stale-master` into `main`, after that I applied another bunch of changes and fixes to adapt tests that were not covered by this PR as it was based on the code base of `v1.1`.

Everything seems to be working fine, I tested it locally with the latest patch releases of PHP 7.2, 7.3 and 7.4 and against Redis 3 and Redis 6 (I didn't bother with intermediate release for now).

I'd suggest leaving this PR as is and avoid making any further changes, I'll push a new PR tomorrow based on my local branch.

nrk:
Thank you @srmklive this PR can be closed now that #644 (which includes your commits) and #646 have been merged into `main`.

PR #641, opened 2020-08-17
use present for copyright range

heithemmoumni:


nrk:
I'm not a lawyer, I did a quick search because I don't remember having ever seen using _present_ in the copyright notice of `LICENSE` files. I've found [this question on StackOverflow](https://opensource.stackexchange.com/q/4885) with an accepted answer that I read as "_it's better to avoid it_". [This is also interesting](https://opensource.stackexchange.com/questions/5778/why-do-licenses-such-as-the-mit-license-specify-a-single-year) regarding dates in copyright notices. It's a bit of a gray zone so I'd avoid this and just stick with the most common way of bumping the year.

LICENSE
```diff
@@ -1,4 +1,4 @@
-Copyright (c) 2009-2016 Daniele Alessandri
+Copyright (c) 2009-present Daniele Alessandri
```
b1rdex, line 1:
```suggestion
Copyright (c) 2009-2020 Daniele Alessandri
```

PR #644, opened 2020-08-19
[v2.0] Port previous development branch to new main + PHPUnit 8.x support

nrk:
With this pull request we restore the old development branch into the current `main` branch and, on top of that, we merge changes from #639 (based on the codebase from `v1.1`) and convert to PHPUnit 8.x all remaining tests for classes that didn't exist in the stable branch. Quite a bunch of things changed for v2.0 compared to v1.1 and after reviewing all of the changes I decided to preserve all of them as they are for this PR as I'm still OK with the overall direction. This doesn't mean that everything is set in stone, some changes still need work and other things might change again.

Commit 592dd91 was required due to the nature of the change or PHPUnit would have been unable to complete running the test suite. On the other hand there's still a failing test related to compatibility with Redis 6.0 (slightly different response payload to `COMMAND INFO`), I already have a fix but purposely left this change for a later pull request.

This doesn't address any deprecated function in recent PHPUnit releases (`PHPUnit\Framework\TestCase::at()` is the worst offender but there are a couple of more) which will need to be fixed, some fixes are easy others are not.

This pull request addresses #624 and integrates work from #639 so they can both be closed as soon as this one is merged into `main`.


tests/PHPUnit/PredisTestCase.php
```diff
@@ -204,31 +195,48 @@ protected function createClient(array $parameters = null, array $options = null,
...
-     * @return mixed
+     * @return \Predis\Connection\NodeConnectionInterface
      */
-    protected function getMockConnection($parameters = null)
+    protected function getMockConnectionOfType($interface, $parameters = null)
     {
-        $connection = $this->getMock('Predis\Connection\NodeConnectionInterface');
+        $connection = $this->getMockBuilder($interface)->getMock();
+
+        // TODO: verify that $connection is an instance of \Predis\Connection\NodeConnectionInterface
```
b1rdex, line 209: `$interface` can be checked with `is_a()`

tests/PHPUnit/PredisTestCase.php
```diff
@@ -333,8 +341,6 @@ protected function checkRequiredRedisServerVersion()
      */
     protected function checkRequirements()
```
b1rdex, line 342: Can we drop this method and call `checkRequiredRedisServerVersion` directly in `setUp`?

b1rdex (commented):
Huge! Great work. Though I haven't looked much the `tests` folder.

src/Command/Factory.php
```diff
@@ -109,10 +96,12 @@ public function createCommand($commandID, array $arguments = array())
      */
     public function defineCommand($commandID, $class)
     {
-        $reflection = new \ReflectionClass($class);
+        if ($class !== null) {
```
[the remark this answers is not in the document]
dereuromark: maybe the docblock needs adjusting

I would recommend adding phpstan up to level 5 (minimum for packages usually for good quality).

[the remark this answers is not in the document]
nrk: > Its type is `string`, where `null` comes from?

Indeed the docblock does not fully reflect the current behaviour of this method.

Here `NULL` is used to "undefine" the class handler for the specified command ID ([link to tests](https://github.com/predis/predis/blob/c26909699b6ad9bf17afb2becb619a59101026dd/tests/Predis/Command/RedisFactoryTest.php#L95-L128) just for reference), basically it says to the command factory to forget about any known handler for that command. By default `Predis\Command\Factory` uses a convention-over-configuration approach (different to command profiles in v1.1.) where `$commandID` is translated to the fully-qualified class name `Predis\Command\Redis\$commandID` when invoking `Predis\Command\Factory::createCommand()`. This is just the default behaviour as it can be overridden on a command basis by using `Predis\Command\Factory::defineCommand()`.

When `NULL` is passed as the second argument to this command, `NULL` is stored as the value in the internal named array that explicitly maps command IDs to PHP classes (this also prevent the command factory to pick the class handler by autoloading `Predis\Command\Redis\$commandID`).

Admittedly using `NULL` like this is not good here, and by doing so we are also exposing an implementation detail of the class. It would make more sense to have a `Predis\Command\Factory::undefineCommand()` method and use it accordingly [here in `Predis\Configuration\Option\Commands`](https://github.com/predis/predis/blob/4b2b35202acbaf79c2f543fc0be1d3520acdbad3/src/Configuration/Option/Commands.php#L35-L43) (`commands` is a new client option replacing the old `profile` one).

I'll fix this separately.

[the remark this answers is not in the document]
nrk: > I would recommend adding phpstan up to level 5 (minimum for packages usually for good quality).

Running phpstan against Predis is indeed something I'd like to try sooner or later, I'm just not sure how it would handle the mix used in the library with a (mostly) strongly-typed approach for the inner parts and an heavily loosely-typed approach for the front-facing stuff (`Predis\Client` and related interfaces) most commonly used by developers.

tests/PHPUnit/PredisTestCase.php
nrk: Makes sense.

src/Command/Factory.php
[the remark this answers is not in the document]
dereuromark: It would work for both just fine IMO, as it only finds the violations of specs, param/return and annotations etc.

AhmedRaafat14:
Incredible work. Just started looking into the files :)

tests/PHPUnit/PredisTestCase.php
nrk: I was being lazy here and left the comment instead of implementing this check once and for all 🙄

PS: hmmm OK maybe saying "I was being lazy" after going through two merges with more than 300 conflicts and 483 files doesn't sound too convincing, but it's true 😂

src/Command/Redis/QUIT.php
```diff
@@ -9,14 +9,16 @@
...
+
+use Predis\Command\Command as RedisCommand;
 
 /**
  * @link http://redis.io/commands/quit
  *
  * @author Daniele Alessandri <suppakilla@gmail.com>
  */
-class ConnectionQuit extends Command
+class QUIT extends RedisCommand
```
dereuromark, line 21: Why these class names?
I would recommend sticking to the naming defaults here (Quit, ...)

src/Client.php
```diff
@@ -188,23 +194,47 @@ public function getOptions()
...
+     * NOTE: This method works against any kind of underlying connection object
+     * as it uses a duck-typing approach and looks for a suitable method that
+     * matches the selector type to extract the correct connection.
+     *
+     * @param string        $selector Type of selector (`id`, `key`, `slot`, `command`)
+     * @param string        $value    Values of selector.
+     * @param callable|null $callable Optional callable object.
      *
-     * @return Client
+     * @return ClientInterface|mixed
```
dereuromark, line 212: is it really mixed or rather object at least?

nrk: I think this should really be just `ClientInterface` and I can't think of a valid reason why `mixed` is actually there.

And to be honest I can't even remember why there's an optional third parameter for this method accepting a callable object, as this case is not even covered by the tests and I can't think of a valid use case for this.

src/Command/Redis/QUIT.php
nrk: Mostly because of case-sensitivity issues with autoloading.

With the [current implementation in `Predis\Command\RedisFactory`](https://github.com/predis/predis/blob/4b2b35202acbaf79c2f543fc0be1d3520acdbad3/src/Command/RedisFactory.php#L36-L49) the command ID is uppercased in order to follow the (mostly practical) convention used by Redis and have a direct lookup in the internal `$commands` dictionary, then the same uppercased string is used as-is by `class_exists()`. While it's ture that class names are treated as case insensitive by the PHP engine, PSR-4 enforces case sensitivity for autoloading so it'd break.

That said, and considered I'm not exactly satisfied with having a 1-1 match between Redis command ID and class name due to current (and potentially more, in the future) clashes with PHP keywords, it would probably make sense to have class names named like `GetCommand` instead of `GET` (albeit the Command part seems a bit redundant). Class names are incrementally cached by the internal `$commands` dictionary so doing a bit more stuff with the string to infer the actual class name would not really have an impact with performance (I always think about cases such as hefty pipelines where thousands of instances of commands can be created).

nrk:
Just to make things clear, most of the feedback that will be provided here won't be implemented with this PR (I mean stuff like [this](https://github.com/predis/predis/pull/644#discussion_r472959183), [this](https://github.com/predis/predis/pull/644#discussion_r472865273) and [this](https://github.com/predis/predis/pull/644#discussion_r472965704)) but will be handled separately with specific pull requests with direct references to these comments.

The reason for this PR is to get a general review of what's going to be merged in `main` and eventually fix oversights mostly related to the merge of two diverging branches. Still, any kind of quick feedback or requests for clarification would be appreciated, I think it's a great opportunity even if I'm plenty aware that going through such a massive changeset is demanding.

src/Command/Redis/QUIT.php
tillkruss: That was my first thought as well `GetCommand` over `GET`, but we can change that after merging.

nrk:
Just for information, I got the test suite running locally on PHP 8.0 by using `composer install --ignore-platform-reqs` to bypass the version constraint of `phpspec/prophecy` which is still preventing phpunit from being installed on this runtime. The test suite runs just fine. I don't think it's possible to configure [our workflow](https://github.com/predis/predis/blob/main/.github/workflows/tests.yml) so that `--ignore-platform-reqs` is used only when running builds on PHP 8.0, but I'm just too new to GitHub Actions to know that for sure. Anyone?

srmklive:
> Just for information, I got the test suite running locally on PHP 8.0 by using `composer install --ignore-platform-reqs` to bypass the version constraint of `phpspec/prophecy` which is still preventing phpunit from being installed on this runtime. The test suite runs just fine. I don't think it's possible to configure [our workflow](https://github.com/predis/predis/blob/main/.github/workflows/tests.yml) so that `--ignore-platform-reqs` is used only when running builds on PHP 8.0, but I'm just too new to GitHub Actions to know that for sure. Anyone?

Running tests for PHPUnit 8 in a separate workflow file makes sense. That's one way to do it.

b1rdex:
> Just for information, I got the test suite running locally on PHP 8.0 by using `composer install --ignore-platform-reqs` to bypass the version constraint of `phpspec/prophecy` which is still preventing phpunit from being installed on this runtime. The test suite runs just fine. I don't think it's possible to configure [our workflow](https://github.com/predis/predis/blob/main/.github/workflows/tests.yml?rgh-link-date=2020-08-19T17%3A15%3A01Z) so that `--ignore-platform-reqs` is used only when running builds on PHP 8.0, but I'm just too new to GitHub Actions to know that for sure. Anyone?

@nrk please take a look on #646

nrk:
> @nrk please take a look on #646

@b1rdex Nice and clean! Thanks!

nrk:
If there's nothing to add or review, I'm going to merge this sometime tomorrow (CEST) along with #646. Separate pull requests addressing the rest of the feedback not implemented here will follow.

tillkruss:
@nrk: Fantastic!

nrk:
Guess what? :tada: And with #646 we also have PHP 8 covered.

nrk:
By the way I don't have a PR for [class names in the `Predis\Command\Redis` namespace](https://github.com/predis/predis/pull/644#discussion_r472959183) yet. They will be changed for sure, but I think it's low priority for now and it's mostly a chore. In the meanwhile I'm still deciding how to properly handle the conversion `$commandID` => FQCN in `Predis\Command\Factory::createCommand()`. Maybe I'll just open a separate issue for now and we'll see.

PR #646, opened 2020-08-20
Fix PHPUnit installation on PHP 8

b1rdex:


nrk:
Great! I think we can indeed use this approach for our test workflow on GitHub Actions until all of the dependencies of PHPUnit are updated to officially support PHP 8.0. Will merge this PR into `main` after #644 (just keeping them separate issues).

PR #648, opened 2020-08-25
Change method used to undefine commands in command factory.

nrk:
This pull request targets some issues [emerged from feedback](https://github.com/predis/predis/pull/644#discussion_r472865273) in #644.

The previous implementation for undefining commands in command factory was not good because we were exposing in the public API an internal implementation detail of the base factory class, furthermore it made `Predis\Command\Factory::defineCommand()` confusing. Having a separate method to undefine commands in the factory is self-explanatory.

We also changed `Predis\Configuration\Option\Commands` accordingly when a dictionary of `[$commandID => $classCommand]` is passed to the `commands` client option and `$classCommand` is `NULL`.

A few minor changes (mostly cosmetic or documentation) were applied too.

By the way, I was thinking about removing the "command" part from method names, so `defineCommand()` would become `define()` or `createCommand()` would become `create()`. Initially I was about to push this change here, but I decided to leave it for another PR. What do you think?

### DIFF
--- a/src/Command/Factory.php
+++ b/src/Command/Factory.php
@@ -50,11 +50,11 @@ abstract class Factory implements FactoryInterface
     }
 
     /**
-     * Returns the FQN of a class that represents the specified command ID.
+     * Returns the FQCN of a class that represents the specified command ID.
      *
      * @codeCoverageIgnore
      *
-     * @param string $commandID Command ID.
+     * @param string $commandID Command ID
      *
      * @return string|null
      */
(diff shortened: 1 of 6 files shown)


AhmedRaafat14:
It makes sense for me to remove the `Command` part from the methods names as it is a Command Factory anyway.

nrk:
> It makes sense for me to remove the `Command` part from the methods names as it is a Command Factory anyway.

Yes those method names are essentially leftovers from `Redis\Profile\RedisProfile` in v1.1, but now that command factory classes live in the `Predis\Command` namespace it's pretty obvious that the scope for such classes is limited to create instances of command objects and nothing else.

AhmedRaafat14:
> ..., but now that command factory classes live in the `Predis\Command` namespace it's pretty obvious that the scope for such classes is limited to create instances of command objects and nothing else.

So I would say go for it and remove it 👌

nrk:
Alright I'll be merging this tomorrow and then proceed with dropping the word `command` from method names. I have a few more breaking changes planned for `Predis\Connection\FactoryInterface` and `Predis\Connection\Factory`.

PR #650, opened 2020-08-25
Filter NULL and zero-length string values for parameters.

nrk:
Having `NULL` values or zero-length strings for connection parameters does not make much sense and actually it proved to be an issue with certain parameters like `password` where an empty string would trigger an `AUTH` command with an empty password (and obviously Redis was not happy with that). The main offenders were a few libraries / bundles / frameworks that kept passing empty values for parameters such as `database` and `password` even when users left them unconfigured. This fix should make things more robust and avoid such occurrences in the future.

I'm targeting v2.0 here but I wonder if it would be acceptable to backport this change to v1.1? Personally I can't think of a reason for why one would explicitly want to rely on this quirk (elements still being populated even when their values are NULL or literally empty strings) when dealing with connection parameters so I guess backporting should be fine as it would also fix the above mentioned issue. What do you think?

Related to PR #436 (rejected).

### DIFF
--- a/src/Connection/Parameters.php
+++ b/src/Connection/Parameters.php
@@ -20,9 +20,7 @@ namespace Predis\Connection;
  */
 class Parameters implements ParametersInterface
 {
-    private $parameters;
-
-    private static $defaults = array(
+    protected static $defaults = array(
         'scheme' => 'tcp',
         'host' => '127.0.0.1',
         'port' => 6379,
@@ -33,17 +31,21 @@ class Parameters implements ParametersInterface
      */
     public function __construct(array $parameters = array())
     {
-        $this->parameters = $parameters + $this->getDefaults();
+        $this->parameters = $this->filter($parameters + static::$defaults);
     }
 
     /**
-     * Returns some default parameters with their values.
+     * Filters parameters removing entries with NULL or 0-length string values.
+     *
+     * @params array $parameters Array of parameters to be filtered
      *
      * @return array
      */
-    protected function getDefaults()
+    protected function filter(array $parameters)
     {
-        return self::$defaults;
+        return array_filter($parameters, function ($value) {
+            return $value !== null && $value !== '';
+        });
     }
 
     /**
--- a/tests/Predis/Connection/ParametersTest.php
+++ b/tests/Predis/Connection/ParametersTest.php
@@ -39,23 +39,29 @@ class ParametersTest extends PredisTestCase
     {
         $parameters = new Parameters();
 
-        $this->assertTrue(isset($parameters->scheme));
-        $this->assertFalse(isset($parameters->unknown));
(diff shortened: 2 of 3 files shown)


nrk:
@predis/team does anyone have feedback about :arrow_down: 

> I'm targeting v2.0 here but I wonder if it would be acceptable to backport this change to v1.1? Personally I can't think of a reason for why one would explicitly want to rely on this quirk (elements still being populated even when their values are NULL or literally empty strings) when dealing with connection parameters so I guess backporting should be fine as it would also fix the above mentioned issue. What do you think?

jimbonator:
Certainly this makes sense for the current connection parameters (unless I'm overlooking one).  I wonder if there's a chance a connection parameter will be added in v2.x where either value would be desirable.  (I can't imagine one, though.)

The closed #436 could be fixed by using `strlen($parameters->password)` instead of `empty()`.  That would be a more modest, bug-fix-only change for v1.1 while this more ambitious change could target v2.0.

nrk:
> Certainly this makes sense for the current connection parameters (unless I'm overlooking one). I wonder if there's a chance a connection parameter will be added in v2.x where either value would be desirable. (I can't imagine one, though.)

There's no plan to rely on zero-length strings or NULL values in connection parameters for anything in Predis v2.0 and future releases. It's true that parameters are not limited to the ones being used internally by Predis and anyone can pass whatever key / value pair they want, but I wonder why **for practical usage in this context** anyone would treat something like `myvalue=NULL` or `myvalue=""` differently than just having `myvalue` as not being set at all. The scope of connection parameters is extremely limited after all.

Client options for example are a whole different story as they are meant to hold anything related to the business logic of the client so `NULL` and zero-length string values are left intact, in fact a client option explicitly set to `NULL` or `""` is still considered to be defined by the options container as defined by the user, and it will return the original value on `__get()` (unless there's a specific option handler set for an option that takes care of transforming that value).

> The closed #436 could be fixed by using `strlen($parameters->password)` instead of `empty()`. That would be a more modest, bug-fix-only change for v1.1 while this more ambitious change could target v2.0.

Hmmm yes I guess you are right, a more conservative approach is probably a better idea for v1.1.

Thank you!

tillkruss:
I tend to stick SemVer with potential breaking changes, but in this case it's hard to say.

srmklive:
I think pushing this change to 1.2 makes sense.

nrk:
> I _personally_ tend to stick SemVer with potential breaking changes, but in this case it's hard to say.

@tillkruss indeed it's hard to say but in the end I think that playing it safe with v1.1 is the best so going with `strlen()` for `password` and `database` as suggested by @jimbonator instead of changing the behaviour of `Predis\Connection\Parameters`.

> I think pushing this change to 1.2 makes sense.

@srmklive probably yes, but to be honest I'd like to avoid having a minor release for v1 and stick with patch releases in v1.1 but we'll see I'm still focused on v2.0.

PR #654, opened 2020-08-29
Various improvements to the test suite

nrk:
Another humongous set of changes, luckily vscode and a few regexp-based search & replace helped a lot.

- Added typehints for function parameters and return values everywhere.
- Added `@var` where needed to give proper hints to IDEs and avoid annoying warnings.
- Replaced the hard deprecated `TestCase::at()`, this gets rid of 99% of the warnings on PHPUnit 9 and it's better anyway.
- Replaced `MockObject::setMethods()` (soft deprecated in PHPUnit since last year) with `addMethods()` and `onlyMethods()`.
- Replaced generic `MockObject::will()` with specialized methods like `willReturn()`, `willReturnCallback()`, etc...
- Rewording of some phpdocs descriptions and annotations.
- Took the opportunity to improve some tests while converting them.

There is still one hard deprecation warning on PHPUnit 9 about `assertRegExp()`, they say to replace it with `assertMatchesRegularExpression()` but then it gave me an error on PHPUnit 8. It's just three warnings being emitted in the output of a test run so we can leave it for now, I'll look into it later.

### DIFF
--- a/bin/create-command-test
+++ b/bin/create-command-test
@@ -187,7 +187,7 @@ class $class extends PredisCommandTestCase
     /**
      * {@inheritdoc}
      */
-    protected function getExpectedCommand()
+    protected function getExpectedCommand(): string
     {
         return '$fqn';
     }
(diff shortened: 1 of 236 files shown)


tillkruss:
Amazing work 🚀

nrk:
Just rebased to track changes in `main`.

PR #657, opened 2020-09-08
Fix zrevrangebyscore pipeline annotation

Qusonann:
It looks like there is a mistake in argument order. It behaves as original redis function, https://redis.io/commands/zrevrangebyscore, annotation was incorrect.


nrk:
@Qusonann thank you!

Issue #643, opened 2020-08-19
Promises & futures for Predis pipelines


jimbonator:
One technique we've used with some success involves a miniature library I developed on top of Predis called [Deferred](https://github.com/internetarchive/deferred).  I demonstrated this library at RedisConf 2018.

Deferred formalizes a promises/futures model that use Predis pipelines (transaction, pipeline, and atomic pipeline) under-the-covers.  It's a more object-oriented approach to pipelining, and improves encapsulation both for assembling commands and processing the responses.

In particular, it allows for responses to be bound directly to PHP variables, to automatically transform values and merge them into complex data structures, and provides notifications for logging/statistics.

My proposal is for this library (or its approach) to became a standard part of Predis, since I think there's a lot of common ground between the two.  I'd be happy to help transition the existing code to Predis, or to assist in building a sub-module that follows Deferred's model.

More information can be found in Deferred's [README](https://github.com/internetarchive/deferred/blob/master/README.md).

nrk:
I'm finally testing with actual code a new approach for the pipeline abstraction's internals in v2.0 using generators (yay for dropping old PHP versions!) and in my various tests to verify if the new design is actually practical and consistent (while giving more flexibility than before) I also tried to emulate the basic idea behind Deferred and in the end I think, unless I missed something along the way, that something like Deferred could be indeed plugged into the new design easily by extending the base Pipeline class (yeah we'll have just on class) and override 2 or 3 methods. I say "emulate" because I did not use promises / futures to keep my test simple, don't have time ATM, I'll try in the next few days or maybe I'll leave it up to you as soon as I'll push a draft PR 😉

nrk:
Alright I'm tentatively adding this feature for v2.0, at least for pipelines. To be honest I can't say for plain transactions yet, ideally `Predis\Transaction\MultiExec` should be re-implemented from the ground up based on an internal design similar to what I envisioned for pipelines, that's what I was planning to do anyway so we'll see.

[...]

PR #663, opened 2020-09-20
New internals and features for pipelining abstraction.

nrk:
**NOTE** this is a work in progress! Tests are missing right now or are based on the old pipeline classes (meaning they are blatantly failing), it will take quite some time to cover all the new classes and various corner cases. Aside from that, the current implementation should be free of glaring bugs.

From a consumer point of view, changes are minor when pipelines are initialized using `Predis\Client::pipeline()`: their default behaviour is the same of Predis v1.1 and the same set of pipeline options is supported. This is just on the surface though because now we have just one pipeline class that internally relies on the new concept of **command queues** (their namespace is `Predis\Pipeline\Queue`) to encapsulate different strategies such as basic pipelining, atomic and fire-and-forget.

The scope of a command queue is to take instances of `Predis\Command\CommandInterface` and enqueue them so that they can be flushed to the desired target connection. As a general rule, unless otherwise stated, flushing the queue returns a traversable that MUST always be consumed by the caller to make sure all pending responses on the target connection are cleared to avoid protocol desynchronization issues. Command queues use PHP generators under the hood.

Queue strategies currently implemented in the `Predis\Pipeline\Queue` namespace are:

- **`Basic`**: default command queue that implements a standard queue strategy where commands are enqueued until a flush is explicitly requested over the target connection.

- **`FireAndForget`**: queued commands are sent over the target connection but responses are not read back from the server. This is accomplished by closing the underlying connection in order to quickly drop any pending response, meaning that the returned traversable can be safely ignored.

- **`Atomic`**: wraps an existing command queue instance in a `MULTI` ... `EXEC` block so that the whole pipeline is flushed in a transaction hiding all the required logic. Due to the very nature of `MULTI` ... `EXEC`, the traversable instance from the inner queue is automatically consumed so only actual responses in `EXEC` response payload are returned to the caller.

Thanks to the fact that command queues always return a `Traversable` for responses, now a pipeline can be configured to *stream* replies to the caller instead of just returning an array. Please note that actually not all command queues support streaming in a real sense, it mostly depends on their underlying implementation. Anyway, traversable responses should be handled with care in order to avoid any desynchronization issue. Also note that partial flushes are unsupported when a pipeline is configured to return a traversable response. Pipelines can be configured to return traversable responses from `Predis\Client::pipeline()` simply by setting the new **`traversable`** option to `TRUE` in the pipeline options dictionary (default is `FALSE`).

Atomic pipelines rely on the `Atomic` command queue which is a wrapper for other command queues, meaning that fire-and-forget pipelines can now be sent in a `MULTI` ... `EXEC` context simply by setting both `fire-and-forget` and `atomic` to `TRUE` in the pipeline options array. Ideally, this command queue will also be re-used for `Predis\Transaction\MultiExec`.

Furthermore, now pipelines even support iterable multibulk responses with `Predis\Connection\CompositeStreamConnection`. This is kind of an obscure feature but when combined with pipelines configured to return traversable responses it allows to effectively stream **every** element of the response at the protocol level. Atomic pipelines are supported, too. And when the pipeline is configured to return plain array responses, any iterable multibulk response is consumed automatically by the pipeline and returned as a plain array element. This feature should be used with even more care than normal traversable responses returned by pipelines.

Please note that this pull request does not directly target some cases from other issues / pull requests but they were considered during development. On a related note, implementing #643 on top of the new `Predis\Pipeline\Pipeline` class should be really easy and I already have a working proof-of-concept but it will be published separately as it is outside the scope of this pull request.

I need help for testing the new design and verify its correctness.


src/Client.php
```diff
@@ -423,40 +424,55 @@ private function sharedContextFactory($initializer, $argv = null)
...
-        if (isset($options['atomic']) && $options['atomic']) {
-            $class = 'Predis\Pipeline\Atomic';
-        } elseif (isset($options['fire-and-forget']) && $options['fire-and-forget']) {
-            $class = 'Predis\Pipeline\FireAndForget';
+        if (isset($options['traversable']) && $options['traversable']) {
+            $traversable = true;
         } else {
-            $class = 'Predis\Pipeline\Pipeline';
+            $traversable = false;
         }
```
emmanix2002, line 457: I hope this isn't a nitpick but I'd recommend we make this terse like so:

```php
$traversable = isset($options['traversable']) && $options['traversable'];

// alternatively

$traversable = false;

if (isset($options['traversable']) && $options['traversable']) {
    $traversable = true;
}
```

src/Client.php
```diff
@@ -423,40 +424,55 @@ private function sharedContextFactory($initializer, $argv = null)
...
-        $pipeline = new $class($this);
+        if (isset($options['fire-and-forget']) && $options['fire-and-forget']) {
+            $queue = new Queue\FireAndForget();
+        } else {
+            $queue = new Queue\Basic();
+        }
+
+        if (isset($options['atomic']) && $options['atomic']) {
+            $queue = new Queue\Atomic($queue);
+        }
```
emmanix2002, line 467: We could apply the alternative suggestion from above here:

```php
$queue = new Queue\Basic();

if (isset($options['fire-and-forget']) && $options['fire-and-forget']) {
    $queue = new Queue\FireAndForget();
}

if (isset($options['atomic']) && $options['atomic']) {
    $queue = new Queue\Atomic($queue);
}
```

I believe this will make it easier to read as well.

src/Pipeline/Pipeline.php
```diff
@@ -119,124 +147,252 @@ protected function getConnection()
...
+            $this->responses = new Iterator\Responses($responses);
+        } else {
+            if ($this->responses === null) {
+                $this->responses = new AppendIterator();
+            }
+
+            $this->responses->append(new ArrayIterator(
+                iterator_to_array($responses, false)
+            ));
         }
```
emmanix2002, line 297: Apologies if my comment seem like nitpicks, I'll mostly be making suggestions to improve readability

```php
// since the else isn't exactly connected to the logic here, we ca break them apart and it'll still
// be comfortable for the reader
if ($this->traversable) {
    $this->responses = new Iterator\Responses($responses);
    return;
}

if ($this->responses === null) {
    $this->responses = new AppendIterator();
}
$this->responses->append(new ArrayIterator(
    iterator_to_array($responses, false)
));

```

src/Client.php
srmklive: It can also be something like:

```
$traversable = isset($options['traversable']) && $options['traversable'] ? true : false;
```

nrk: Nitpicking is fine!

And sometimes I still forget that the null coalesce operator in PHP 7 works fine for lookups on undefined indexes:

```php
$traversable = (bool) ($options['traversable'] ?? false);
```

With psysh:

```
>>> $options = [];
=> []
>>> $traversable = (bool) ($options['traversable'] ?? false);
=> false

>>> $options = ['traversable' => false];
=> [
     "traversable" => false,
   ]
>>> $traversable = (bool) ($options['traversable'] ?? false);
=> false

>>> $options = ['traversable' => true];
=> [
     "traversable" => true,
   ]
>>> $traversable = (bool) ($options['traversable'] ?? false);
=> true
```

Can't be more terse than that :smile:

src/Client.php
nrk: In this case I'd prefer to avoid instantiating `Queue\Basic` when not needed. Maybe we can use the null coalesce operator here, too, to get something like this:

```php
// This line on top along with $traversable = ...
$fireandforget = $options['fire-and-forget'] ?? false;

// This line just before checking for atomic.
$queue = $fireandforget ? new Queue\FireAndForget() : new Queue\Basic();
```

How about it?

src/Client.php
emmanix2002: LOL!

src/Pipeline/Pipeline.php
nrk: Makes sense!

src/Client.php
emmanix2002: Yes, makes perfect sense

nrk: So in the end I guess we can rewrite `Predis\Client::createPipeline()` like this:

```php
protected function createPipeline(array $options = null, callable $callable = null)
{
    $traversable = (bool) ($options['traversable'] ?? false);
    $isFireAndForget = (bool) ($options['fire-and-forget'] ?? false);
    $isAtomic = (bool) ($options['atomic'] ?? false);
    $throwExceptions = (bool) ($options['exceptions'] ?? $this->getOptions()->exceptions);

    $queue = $isFireAndForget ? new Queue\FireAndForget() : new Queue\Basic();

    if ($isAtomic) {
        $queue = new Queue\Atomic($queue);
    }

    $pipeline = new Pipeline($this, $queue, $traversable);
    $pipeline->setThrowOnErrorResponse($throwExceptions);

    if (isset($callable)) {
        return $pipeline->execute($callable);
    }

    return $pipeline;
}
```

Null coalesce operator galore! :rocket:

emmanix2002: HAHAHA!

nrk: Fixed.

src/Pipeline/Queue/Atomic.php
```diff
@@ -0,0 +1,215 @@
...
+        $this->innerQueue = $queue;
+        $this->txCommands = new SplQueue;
+    }
+
+    /**
+     * Verifies if the inner command queue is compatibile with atomic queues.
+     *
+     * @throws InvalidArgumentException
+     */
+    protected function checkInnerQueue(CommandQueueInterface $queue): void
```
emmanix2002, line 51: We might want to rename this method so it clearly indicates intent, maybe something like `checkInnerQueueCompatibility` or `verifyInnerQueueIsCompatible`

src/Pipeline/Queue/Atomic.php
```diff
@@ -0,0 +1,215 @@
...
+    {
+        if ($queue instanceof static) {
+            throw new InvalidArgumentException('Atomic command queues cannot be nested');
+        }
```
emmanix2002, line 55: This suggestion can be skipped but we can return early on the happy path as it makes the body of the method look more useful.

```suggestion
        if (! $queue instanceof static) {
            return;
        }
        throw new InvalidArgumentException('Atomic command queues cannot be nested');
```

src/Pipeline/Queue/Atomic.php
```diff
@@ -0,0 +1,215 @@
...
+            }
+        } elseif (!is_iterable($execResponse)) {
+            $this->reset();
+            $connection->disconnect();
+
+            throw $this->onStateError($connection, sprintf(
+                'Protocol desynchronization detected on `EXEC` response (array expected, `%s` received)',
+                is_object($execResponse) ? get_class($execResponse) : gettype($execResponse)
+            ));
+        }
```
emmanix2002, line 174: This is the only section I'll need to take some time out and take a proper look at later, but I've gone through everything else and it looks good.

Good job @nrk

srmklive:
The suggested changes sounds really good to me @nrk. I can understand the reason of failing tests right now, as intent is clearly to make the underlying code work for all use cases. Is there someway we can help to achieve that goal?

src/Pipeline/Queue/Atomic.php
nrk: I like `checkInnerQueueCompatibility`, will use that.

src/Pipeline/Queue/Atomic.php
nrk: Let me know if you need any clarification.

src/Pipeline/Queue/Atomic.php
nrk: I understand the reasoning behind your suggestion. Actually I implemented the body of the method that way because I think we may introduce a few more checks later (I have a vague idea for more specialized command queues), in that case we would end up having a few cascading checks. Will leave it as committed for now, we'll see later.

nrk:
> The suggested changes sounds really good to me @nrk. I can understand the reason of failing tests right now, as intent is clearly to make the underlying code work for all use cases. Is there someway we can help to achieve that goal?

@srmklive just playing with the code and report strange or wrong behaviours with reproducible code would be enough for now, especially (but not only) when `$traversable` is set to `TRUE`. Even playing with some more ideas to build on top of the new pipelines would be cool, just like #643 (I plan to post a gist with a basic re-implementation just to play with, an actual branch will follow when this one is tested and considered stable enough).

The reason why I avoided writing tests until now is because, aside from the core idea, the whole design for the new pipelines changed quite a lot during the various coding iterations in the last 3 weeks so it would have slowed me down in the initial phase (I'm quite pragmatic and I don't believe in TDD at all costs) but now I plan on working on tests and commit the initial bits ASAP. My idea is to get up and running with the basic tests (some of them can be adapted from the ones for the current pipeline abstraction) and then start testing for corner cases which is probably the most tedious task (and this will indeed need some help).

src/Pipeline/Queue/Atomic.php
emmanix2002: Dope!

src/Pipeline/Queue/Atomic.php
emmanix2002: Ok, that makes sense.

src/Pipeline/Queue/Atomic.php
emmanix2002: Will do 😎

emmanix2002:
> > The suggested changes sounds really good to me @nrk. I can understand the reason of failing tests right now, as intent is clearly to make the underlying code work for all use cases. Is there someway we can help to achieve that goal?
> 
> @srmklive just playing with the code and report strange or wrong behaviours with reproducible samples would be enough for now, especially (but not only) when `$traversable` is set to `TRUE`. Even playing with some more ideas to build on top of the new pipelines would be cool, just like #643 (I plan to post a gist with a basic re-implementation just to play with, an actual branch will follow when this one is tested and considered stable enough).
> 
> The reason why I avoided writing tests until now is because, aside from the core idea, the whole design for the new pipelines changed quite a lot during the various coding iterations in the last 3 weeks so it would have slowed me down in the initial phase (I'm quite pragmatic and I don't believe in TDD at all costs) but now I plan on working on tests and commit the initial bits ASAP. My idea is to get up and running with the basic tests (some of them can be adapted from the ones for the current pipeline abstraction) and then start testing for corner cases which is probably the most tedious task (and this will indeed need some help).

I'll try and play with it later this evening (in the UTC+1 timezone) when I'm done with work

nrk:
> I'll try and play with it later this evening (in the UTC+1 timezone) when I'm done with work

Don't worry I'm not in a hurry to get this done and merged so we can just take some time and avoid rushing things 👍

nrk:
In the meanwhile, on a slightly related note, I posted my [proof of concept of deferred pipelines](https://gist.github.com/nrk/90f0eeedb368e384867a679656f89261) implemented on top of this PR. I'm posting the link not because I want to discuss about them here, it's not the scope of this PR, but to give you a practical examples of some design choices for the internals of pipelines.

Issue #660, opened 2020-09-15
Using local patches for dev dependencies breaks projects that use cweagans/composer-patches


Berdir:
**Describe the bug**

Follow-up of https://github.com/predis/predis/issues/632, which didn't fully address this issue 

Adding patches to composer.json for dev dependencies seems incorrect to me, for two reasons:

a) The local patch files won't be found in the module, AFAIK composer-patches won't look there
b) Even if they would be found, the project is likely to use a different version of phpunit where the patches don't apply.

**To Reproduce**
1. composer init
2. composer require cweagans/composer-patches predis/predis
3. composer config extra.enable-patching true
4. composer require --dev phpunit/phpunit

```
  - Applying patches for phpunit/phpunit
    ./tests/phpunit_php7.patch (Fix PHP 7 compatibility)
   Could not apply patch! Skipping. The error was: The "./tests/phpunit_php7.patch" file could not be downloaded: failed to open stream: No such file or directory
    ./tests/phpunit_php8.patch (Fix PHP 8 compatibility)
   Could not apply patch! Skipping. The error was: The "./tests/phpunit_php8.patch" file could not be downloaded: failed to open stream: No such file or directory
```

**Expected behavior**

**Versions (please complete the following information):**
 - Predis: 1.1.x
 - PHP 7.4
 - Redis Server irrelevant
 - OS: Ubuntu 20.10

I don't think it is possible to do this, not without cweagans/composer-patches introducing a patches-dev. My suggestion would be to switch to applying those patches with a script/travis commands (or whatever CI you use)

tillkruss:
The patches are including in the repo now, did you happen to upgrade from 1.1.1 to 1.1.x by chance?

Berdir:
that's my point, even if they would be found, they wouldn't apply to a current version of phpunit.

tillkruss:
The replicated error:

```
Could not apply patch! Skipping. The error was: The "./tests/phpunit_php7.patch" file could not be downloaded: failed to open stream: No such file or directory
```

tillkruss:
@Berdir Have you tried using `patches-ignore` to skip the Predis patches?

Berdir:
I have not, but it shouldn't be the responsibility of every project to work around this?

fgm:
@tillkruss could you suggest the recommended format of `patches-ignore` until a better solution can be found ?

tillkruss:
@fgm: That looks good.

@cweagans: Do you have any suggestions with this?

[...]

PR #673, opened 2020-11-27
Apply patches on initial install/update

roborourke:
The patches to get PHPUnit working are causing problems for other projects that use both `cweagans/composer-patches` and `predis/predis` along with PHPUnit.

This change moves the patch application to a simple PHP script run on the `post-update-cmd` composer event and removes the dependency on the patch installer package.

Addresses #660

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -31,8 +31,7 @@
         "php": ">=5.3.9"
     },
     "require-dev": {
-        "phpunit/phpunit": "~4.8",
-        "cweagans/composer-patches": "^1.6"
+        "phpunit/phpunit": "~4.8"
     },
     "suggest": {
         "ext-phpiredis": "Allows faster serialization and deserialization of the Redis protocol",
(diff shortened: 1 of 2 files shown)


nrk:
Thanks @roborourke, I'll give it a try this weekend!

roborourke:
Switched to old array syntax to support PHP 5.3 and fix the failing test there.

roborourke:
🤔 not sure why there's a failing test for PHP 5.3. Seems unrelated to this PR. Is that a known issue @nrk?

nrk:
@roborourke yes it should be unrelated to your PR, I guess it's due to the change in ca468b7 but right now I'm not sure why it fails only when running on PHP 5.3. I'll check later, I'm rebuilding my VM with all of the binaries from PHP 5.3 up to 8.0 for a final check locally, targeting a new patch release this weekend.

roborourke:
Any luck with this on your end?

roborourke:
Anything I can do to help this along?

tillkruss:
Related #679.

PR #678, opened 2021-02-02
Add some return types

nicolas-grekas:
By making these return types explicit in the annotation, we preemptively tell consumers that some next major of Predis or PHP 9 might add real return types here.

These annotations are leveraged by Symfony's `DebugClassLoader` to help one spot some missing return types in their app. That's how I spotted these. They relate only to PHP internal's base types / magic methods.


tillkruss:
LGTM

PR #679, opened 2021-02-09
Skip broken PHP 5.3 test

tillkruss:
This test fails as of ca468b7 only on PHP 5.3. I don't know why.


tillkruss:
http://blog.adamcameron.me/2020/12/phpunit-get-code-coverage-reporting.html

PR #685, opened 2021-03-17
Fix misleading argument name in SentinelReplication

BrianMwit:
In `SentinelReplication`, `$this->retryWait` is in milliseconds.
It's setter misleadingly document it as seconds. Fixing by renaming the variable name without altering the functionality.

I believe this patch would also apply to the `main` branch in https://github.com/predis/predis/blob/main/src/Connection/Replication/SentinelReplication.php#L153-L163.


tillkruss:
Thanks @BrianMwit!

PR #686, opened 2021-03-18
fix test expectation for redis 6.2

remicollet:
From Fedora CI
https://koschei.fedoraproject.org/package/php-nrk-Predis

Failing since redis 6.2
```
There were 2 failures:
1) Predis\Command\ConnectionSelectTest::testThrowsExceptionOnUnexpectedDatabaseName
Failed asserting that exception message 'ERR value is not an integer or out of range' contains 'ERR invalid DB index'.
2) Predis\Command\KeyMoveTest::testThrowsExceptionOnInvalidDatabases
Failed asserting that exception message 'ERR DB index is out of range' contains 'ERR index out of range'.
```


tillkruss:
Thanks @remicollet!

PR #688, opened 2021-04-04
Improve PHPUnit assertions

peter279k:
# Changed log

- Using the `assertCount` to assert expected is same as result.
- Using the `assertInstanceOf` to assert expected is same as result instance.


tillkruss:
Thanks @peter279k!

PR #697, opened 2021-06-02
Reuse connection scheme and SSL options from redis-sentinel parameters

ignasbernotas:
Hi! First of all, thank you for this package, it was a pleasure using and reading your code! :)

I'm using Redis 6.2 with Sentinel and self-signed TLS. I wasn't able to get it to work because the TLS details weren't being passed back before making the forwarding connections. This fixes it.

Setting the `scheme` broke some tests (tests were finding the `scheme` prop when none was expected) so I've added the `if ($parameters->scheme !== 'tcp') {` line to bypass that. I'm happy to make adjustments if needed ✌️

I'm using this package with Laravel.


tillkruss:
Thanks for this. Can you add tests as well?

tillkruss:
Please re-submit with a test case.

Issue #699, opened 2021-06-03
Please add count argument to lpop


Miteugene:
lpop command has count argument https://redis.io/commands/lpop
This feature would be very useful.
Thanks!

tillkruss:
Feel free to open a PR with the change 👍

KryDos:
Seems like predis, fortunately, don't require any changes when number of arguments changes for a command. The only thing that could be done is a test case to cover that new argument.

PR #700, opened 2021-06-08
Test optional `count` argument for LPOP

KryDos:
Just added one more test case to test LPOP's optional `count` argument.

First time here, let me know if I could do anything better. 

Related issue: #699


KryDos:
Hm... I wonder if I should update `src/ClientContextInterface` and have `@method $this lpop($key, $count = null)`
Let me know if update is needed.

PR #681, opened 2021-02-17
Fix no sentinel server available for autodiscovery on Swoole

Linc70J:
Question:
In the Swoole to close or restart sentinel may cause the node invalidation forever,
the node deleted because of in the SentinelReplication.php use array_shift($this->sentinels)

Solution:
Using $this->sentinels[$this->sentinelIndex] to replace array_shift($this->sentinels) can solve the problem, also can use  if ($this->sentinelIndex >= count($this->sentinels)) in the same time to replace  if (!$this->sentinels) prevent  undefined offset.

PR #701, opened 2021-06-16
Add sentinels back after connecting to it

guiajlopes:
Hi,

I'm sending this PR to fix an issue where all sentinels go down and you have a long lived php process it will not reconnect to the sentinels because of that and you will receive `throw new \Predis\ClientException('No sentinel server available for autodiscovery.');` all the time after.

The problem is that when we create connections we remove the sentinel from the list of available sentinels ($this->sentinels) and when we do not have any valid connection anymore (it's removed by SentinelReplication:321) it will not be able to reconnect


guiajlopes:
closing in favor #681

PR #706, opened 2021-08-04
Add ReturnTypeWillChange

derrabus:
PHP 8.1 triggers a deprecation warning if a class implements `IteratorAggregate` without specifying a return type for `getIterator()`. Doing so would break PHP 5 compatibility, so I've added the `ReturnTypeWillChange` attribute which tells PHP that the return type will be added in a future release. This way, the error is not triggered.


tillkruss:
Thanks!

PR #710, opened 2021-09-04
Updated set, setex, setexp type hint

### LINKED ISSUE #709: ClientInterface::setex() is supposed to return int but actually returns \Predis\Response\Status
**Describe the bug**
I think that some methods are wrongly described in `ClientInterface`  in `@method` annotations.
This actually cause me lot of issues with Phpstan due to bad method typing.

Can you update the interface and provide the good return type hints please ?

**To Reproduce**
Explained above

**Expected behavior**
Good typed method that actually match the one described in `ClientInterface`  in `@method` annotations.

 - Predis: latest
 - PHP 7.2+
 - Redis Server (not relevant here)
 - OS Windows 10 x64

**Code sample**

Actual code flagged by Phpstan as invalid. 
And I need the return of `setex` to prove the successful operation to the code caller.
```php
        return $this->instance->setex($item->getKey(), $ttl, $this->encode($this->driverPreWrap($item)))->getPayload() === 'OK';
```

**Additional context**
N/A

Geolim4:
This fixes #709

### DIFF
--- a/src/ClientInterface.php
+++ b/src/ClientInterface.php
@@ -15,6 +15,7 @@ use Predis\Command\CommandInterface;
 use Predis\Command\FactoryInterface;
 use Predis\Configuration\OptionsInterface;
 use Predis\Connection\ConnectionInterface;
+use Predis\Response\Status;
 
 /**
  * Interface defining a client able to execute commands against Redis.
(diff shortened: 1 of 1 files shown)


tillkruss:
Thanks!

PR #712, opened 2021-09-11
Add PHP 8.1 support

javer:


tillkruss:
Thanks for submitting this. Can you please post the errors/deprecations you're seeing when running the current release in 8.1?

src/Transaction/AbortedMultiExecException.php
```diff
@@ -29,7 +29,7 @@ class AbortedMultiExecException extends PredisException
      */
     public function __construct(MultiExec $transaction, $message, $code = null)
     {
-        parent::__construct($message, $code);
+        parent::__construct($message, $code !== null ? $code : 0);
```
javer, line 32: Reason:
```
Exception::__construct(): Passing null to parameter #2 ($code) of type int is deprecated
```

PR #718, opened 2021-11-04
Throw an exception if factory parameters are not an array

Starfox64:
We are running into an issue in production where we randomly get a `Error: Unsupported operand types` exception about this line:
https://github.com/predis/predis/blob/4c1aada5aebf0afde266b512ed069f822148c3a2/src/Connection/Factory.php#L151

The exception added by this PR will hopefully help us determine the cause of `$parameters` not being an array.

### DIFF
--- a/src/Connection/Factory.php
+++ b/src/Connection/Factory.php
@@ -146,6 +146,10 @@ class Factory implements FactoryInterface
         } else {
             $parameters = $parameters ?: array();
         }
+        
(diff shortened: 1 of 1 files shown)


tillkruss:
What version are you running in production?

Starfox64:
We are running 1.1.7

tillkruss:
Okay, please re-submit this PR to the `v1.1` branch.

Issue #722, opened 2021-11-19
Cannot use object of type Predis\Response\Error as array


kgasienica:
**Describe the bug**
```
 Error {#515
  #message: "Cannot use object of type Predis\Response\Error as array"
  #code: 0
  #file: "/var/www/mobile_api/vendor/predis/predis/src/Connection/Aggregate/SentinelReplication.php"
  #line: 533
  trace: {
    /var/www/mobile_api/vendor/predis/predis/src/Connection/Aggregate/SentinelReplication.php:533 {}
    /var/www/mobile_api/vendor/predis/predis/src/Connection/Aggregate/SentinelReplication.php:549 {}
    /var/www/mobile_api/vendor/predis/predis/src/Connection/Aggregate/SentinelReplication.php:665 {}
    /var/www/mobile_api/vendor/predis/predis/src/Connection/Aggregate/SentinelReplication.php:705 {}
    /var/www/mobile_api/vendor/predis/predis/src/Client.php:331 {}
    /var/www/mobile_api/vendor/predis/predis/src/Client.php:314 {}
    /var/www/mobile_api/src/Controller/ApiProductController.php:97 {
      MobileApi\Controller\ApiProductController->indexAction(string $x, Request $request): Response …
      › $client = new Client($sentinels, $options);
      › $client->set('test3', 'test');
      › dd($client);
      arguments: {
        $commandID: "set"
        $arguments: array:2 [ …2]
      }
    }
    /var/www/mobile_api/vendor/symfony/http-kernel/HttpKernel.php:146 {}
    /var/www/mobile_api/vendor/symfony/http-kernel/HttpKernel.php:68 {}
    /var/www/mobile_api/vendor/symfony/http-kernel/Kernel.php:201 {}
    /var/www/mobile_api/public/index.php:25 {}
  }
}
```

**To Reproduce**
1. Setup 3 sentinels, 1 master redis, 1 slave redis.
2. Make master and slave under password (requirepass XXX)
3. Try to set a KEY with Client while not providing a password, example:
```
        $sentinels = [
            'tcp://192.168.1.1:26379',
            'tcp://192.168.1.2:26379',
            'tcp://192.168.1.3:26379',
        ];
        $options = [
            'replication' => 'sentinel',
            'service' => 'redismaster',
            //  'parameters' => [
            //      'password' => 'XXX',
            // ],

        ];

        $client = new Client($sentinels, $options);
        $client->set('test3', 'test');
```

**Expected behavior**
[...]

PR #723, opened 2021-11-23
Bugfix for: Cannot use object of type Predis\Response\Error as array #722

kgasienica:
Bugfix for:
Cannot use object of type Predis\Response\Error as array #722


tillkruss:
@kgasienica This is great, can you re-submit this change to the `v1.1` branch please?

kgasienica:
Sure :)
https://github.com/predis/predis/pull/724

Issue #716, opened 2021-10-23
Deprecated: Return type of Predis\Client::getIterator() should either be compatible with IteratorAggregate::getIterator()


makhnanov:
**Describe the bug**
Deprecated: Return type of Predis\Client::getIterator() should either be compatible with IteratorAggregate::getIterator(): Traversable, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in /var/www/html/vendor/predis/predis/src/Client.php on line 518

**To Reproduce**
Create new Predis\Client() in PHP 8.1 environment and get strange message.

**Expected behavior**
Run without deprecated notice.

**Versions (please complete the following information):**
 - Predis: [e.g. 1.1.2]
 - PHP 8.1.0RC4

kdambekalns:
Oh, I should have tried the current version first. :) In fact v1.1.8 already fixes this via #706.

And as is mentioned there, adding a proper return type would break compatibility with PHP 5. If that is still a thing is a different story, but I am happy now. ;)

RobertEcker:
I use PHP 8.1.1 and the latest predis ... there is still this error for me?!

tillkruss:
Upgrade to v1.1.8.

RobertEcker:
@tillkruss 1.1.8? isn't the master already 2.0 or do you mean downgrade to 1.1.8? or what should I upgrade to 1.1.8

tillkruss:
@RobertEcker: `master` isn't stable, if you have issues with it, please submit PRs.

PR #725, opened 2021-12-01
Declare getIterator() return type (#716)

kdambekalns:
Fixes https://github.com/predis/predis/issues/716


tillkruss:
Please re-submit to the v1.1 branch.

kdambekalns:
Thanks for the pointer – but I won't do it, as it's fixed as of v1.1.8 already. Doh! 😎

PR #726, opened 2021-12-03
Session handler 8.1 deprecated shutup

pkristian:
Session handler suppress deprecated errors on PHP 8.1

i ma updating my project to PHP 8.1 and these things fixed newly risen deprecated notices.


tillkruss:
Please re-submit to v1.1 and double check the syntax.

PR #729, opened 2021-12-15
Fixes #728

### LINKED ISSUE #728: set slots is slow when there are many masters
**Is your feature request related to a problem? Please describe.**
Currently, setSlots uses getSlotsMap, which returns a new copy of
the existing slotsMap. Because setSlots is called for each master,
the array copy becomes prohibitively expensive for clusters with
a large number of masters. 

**Describe the solution you'd like**
Passing the existing array by reference would speed up the addition of slots

**Describe alternatives you've considered**

**Additional context**
We have been using this solution in our own fork successfully and I wanted to push it back in case it had wider interest

theilig:
Currently, setSlots uses getSlotsMap, which returns a new copy of
the existing slotsMap. Because setSlots is called for each master,
the array copy becomes prohibitively expensive for clusters with
a large number of masters. This commit optimizes the setSlots call
by modifying a reference to the slotsMap associative array instead
of a copy.

### DIFF
--- a/src/Connection/Aggregate/RedisCluster.php
+++ b/src/Connection/Aggregate/RedisCluster.php
@@ -289,7 +289,7 @@ class RedisCluster implements ClusterInterface, \IteratorAggregate, \Countable
      *
      * @return array
      */
-    public function getSlotsMap()
+    public function &getSlotsMap()
     {
         if (!isset($this->slotsMap)) {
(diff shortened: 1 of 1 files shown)


tillkruss:
@nrk: Any thoughts on this?

tillkruss:
Please re-submit with a test case.

Issue #727, opened 2021-12-08
PHP 8.1 support


pimjansen:
Hey,

As mentioned in other commands, last september PHP 8.1 compatibility has been added via:
https://github.com/predis/predis/pull/712

However there is still no release of the package in the 1.x stream. Since PHP 8.1 has been released a while ago when can we expect the actual tag?

**Update**
I was a bit too soon and it seems it is still not compatible:

```sh
phpfpm_1   | [Wed Dec 08 14:36:30.364115 2021] [php:notice] [pid 19] [client 172.19.0.2:47686] PHP Deprecated:  Return type of Predis\\Session\\Handler::open($save_path, $session_id) should either be compatible with SessionHandlerInterface::open(string $path, string $name): bool, or the #[\\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in /app/vendor/predis/predis/src/Session/Handler.php on line 68, referer: 
```

Best

tillkruss:
Yes there is, the fixes are included in v1.1.9.

Can you submit a fix the session handler, if you're currently experiencing that on v1.1.9?

pimjansen:
> Yes there is, the fixes are included in v1.1.9.
> 
> 
> 
> Can you submit a fix the session handler, if you're currently experiencing that on v1.1.9?

Will see if i can do that. But it seems it lacks test support in that case as it passes for 8.1

pimjansen:
@tillkruss im reviewing the issue but there is already an issue with PHPUnit which has a different interface definition for 9.x which is not compatible with the existing implementation.

Since there is a 5.3 support range available i have no idea which way you guys want to go with this?

The first thing that will popup is:
```sh
root@1e6fae7cc728:/app# ./vendor/bin/phpunit 
PHPUnit 9.4.4 by Sebastian Bergmann and contributors.

Class "PHPUnit_Framework_Constraint" not found
root@1e6fae7cc728:/app# 
```

tillkruss:
I'm not sure what you're asking. Were you able to find a PHP 8.1 related issue?

We apply a bunch of patches to PHPUnit: https://github.com/predis/predis/tree/v1.1/tests I would not try to run those local 😉

pimjansen:
> I'm not sure what you're asking. Were you able to find a PHP 8.1 related issue?
> 
> 
> 
> We apply a bunch of patches to PHPUnit: https://github.com/predis/predis/tree/v1.1/tests I would not try to run those local 😉

[...]

PR #730, opened 2021-12-16
Fix for changing return types for PHP8.1

pimjansen:
This PR fixes the return types that change for PHP 8.1

Fixes: #727. This is a followup for #712.
The fix prevents below error for showing when using the Predis Session Handler
```
phpfpm_1   | [Wed Dec 08 14:36:30.364115 2021] [php:notice] [pid 19] [client 172.19.0.2:47686] PHP Deprecated:  Return type of Predis\\Session\\Handler::open($save_path, $session_id) should either be compatible with SessionHandlerInterface::open(string $path, string $name): bool, or the #[\\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in /app/vendor/predis/predis/src/Session/Handler.php on line 68, referer: 
```

### DIFF
--- a/src/Session/Handler.php
+++ b/src/Session/Handler.php
@@ -65,6 +65,7 @@ class Handler implements \SessionHandlerInterface
     /**
      * {@inheritdoc}
      */
+    #[\ReturnTypeWillChange]
     public function open($save_path, $session_id)
     {
         // NOOP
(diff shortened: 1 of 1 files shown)


tillkruss:
Fantastic, could you post the log messages you see without the comments?

pimjansen:
Like this? Or what do you mean exactly? @tillkruss

pimjansen:
Any idea when the next patch will be?

dakorpar:
@tillkruss waiting for release so I can start using php 8.1, any plan on this soon?

pimjansen:
I see it is released with 1.10

PR #731, opened 2021-12-29
Handle rediscluster temporarily unavailable

digital2real:
This changes are trying to complete the retry logic in order to increase the tolerance to a temporary unavailability of a redis cluster.


mfn:
Something if off with the PR, like merging the wrong version into the branch. Yours is the last commit only, but it seems  30 other commits joined yours.

PR #735, opened 2022-01-07
Fix documentation of get method

### LINKED ISSUE #734: ClientInterface::get() is supposed to return string but actually can return \Predis\Response\Status
**Describe the bug**
The ClientInterface::get() has documented `string` as return type, but on some situations (detected in production) it may return `\Predis\Response\Status`.

**To Reproduce**
Steps to reproduce the behavior: unknown

**Expected behavior**
Either only string is returned or the current state is properly documented.

 - Predis: 1.1.9
 - PHP 7.4
 - Redis Server 6.2.6
 - OS Docker redis:latest

**Additional context**
Libraries such as contributte/redis depends on the string type and handle it in such way that `Status` object create an `TypeError` ([issue](https://github.com/contributte/redis/issues/34) reported, but fix here is probably also needed).

repli2dev:
Closes #734


src/ClientInterface.php
```diff
@@ -25,154 +26,154 @@
...
- * @method int         ttl($key)
- * @method mixed       type($key)
- * @method int         append($key, $value)
- * @method int         bitcount($key, $start = null, $end = null)
- * @method int         bitop($operation, $destkey, $key)
- * @method array|null  bitfield($key, $subcommand, ...$subcommandArg)
- * @method int         bitpos($key, $bit, $start = null, $end = null)
- * @method int         decr($key)
- * @method int         decrby($key, $decrement)
- * @method string|null get($key)
```
repli2dev, line 54: Status can be returned in some cases, and Error IMHO when the exception behaviour is supressed.

tillkruss:
@nrk: Any input on this?

repli2dev:
Solved via rebase & push force.

tillkruss:
Thanks. Can you please add a test case that returns `Status` when using `get()`?

repli2dev:
Fair point. I added test cases for `Status` and `Error` returns from `get()` :)

tests/Predis/Command/Redis/GET_Test.php
```diff
@@ -90,6 +90,18 @@ public function testReturnsNullOnNonExistingKeys(): void
...
+    /**
+     * @group connected
+     */
+    public function testReturnsStatusInTransaction(): void
+    {
+        $redis = $this->getClient();
+
+        $this->assertSame(0, $redis->exists('foo'));
+        $this->assertInstanceOf('Predis\Response\Status', $redis->multi());
+        $this->assertInstanceOf('Predis\Response\Status', $redis->get('foo'));
```
tillkruss, line 102: Ah, I see. You're getting a `Status` response when opening a `MULTI` transaction. Theoretically _all_ commands then can return `Error|Status`, which I don't really want to document TBH.

repli2dev: Sadly that is not true, we are not using `MULTI` anywhere and still occasionally getting the Status responses… which seems possible according to Google search when the server is under high load…

The reason I used `MULTI` here was that it is the only way known to me to reproduce the issue in a deterministic way…

On the other hand I completely understand why you don’t want to add it ewerywhere yet when relying on the code of the library and using the static analysis tools this creates a great pain:
- not aware that not documented types can be returned
- even when aware of it you need to mute static analysis warning about dead code… because it relies on the php doc comment anyway in this case…

So what is your proposal to proceed this further?

tillkruss: I don't know, but adding it only for a single command `GET` without a proper test to recreate it seems too dodgy.

I'd be open to adding it to all (most) methods to be more accurate, but this requires work and tests which I don't have the bandwidth for right now.

repli2dev: > I'd be open to adding it to all (most) methods to be more accurate, but this requires work and tests which I don't have the bandwidth for right now. If it's important to you please extend the PR 👍

Will see what I can do :-)

tillkruss:
@repli2dev: No, happy to merge once it's complete. I closed it because 4 weeks of inactivity.

PR #740, opened 2022-01-26
Fix wrong return iterator value type

fluffycondor:
There's a typehint "@return array" in the \Predis\PubSub\AbstractConsumer method current(), but the implementation \Predis\PubSub\Consumer actually returns \stdClass.
It confuses static analyzers a lot.


tillkruss:
Thanks for the PR!

Issue #742, opened 2022-02-04
Connection over TLS affects default stream context


esler:
Socket connection is opened by `stream_socket_client()` without context. PHP will assign default (global) context to the resource. Later in the connection init functions like `stream_context_set_option()` will change resource's context which's global and shared by all file handling functions. That may lead to some nasty issues including security vulnerabilities. I already prepared PR (#743) with a fix.

1. create a `tls` connection with custom certificate
2. fetch someting from valid `https` later in the code
3. see code for more details

Library should not affect any subsequent connections

 - Predis: 1.1.10 (but affects almost all versions, it's there for 7 years)
 - PHP 7.4 and 8.1 (but all supported versions are affected)
 - Redis Server 6.0 (any with TLS support)
 - OS Linux, MacOS

```php
$redis = new Client([
  'scheme' => 'tls',
  'ssl' => [
    'cafile' => 'redis.crt', // self-signed
  ],
]);

copy(
  'https://www.wikipedia.org/portal/wikipedia.org/assets/img/Wikipedia-logo-v2.png',
  '/tmp/wiki-logo.png'
);
// PHP Warning:  copy(): SSL operation failed with code 1. OpenSSL Error messages: 
// error:1416F086:SSL routines:tls_process_server_certificate:certificate verify failed in ...
```

esler:
#714 is most probably caused by this issue

tillkruss:
Fixed in #743.

Issue #714, opened 2021-09-30
stream_socket_enable_crypto(): SSL operation failed with code 1. OpenSSL Error messages: error:14090086:SSL routines:ssl3_get_server_certificate:certificate verify failed


tcagkansokmen:
Suddenly I started to get that error, 

stream_socket_enable_crypto(): SSL operation failed with code 1. OpenSSL Error messages: error:14090086:SSL routines:ssl3_get_server_certificate:certificate verify failed
Illuminate\Foundation\Bootstrap\HandleExceptions::handleError
vendor/predis/predis/src/Connection/StreamConnection.php:246

Whats wrong with It?
I can connect to project from my local. But not working at my server

tillkruss:
Are you using PHP 8.1?

tillkruss:
Fixed in #743.

odahcam:
Isn't this fix available for PHP 8.0?

PR #743, opened 2022-02-04
#742 create new context for socket connection

esler:
Connection created without context will use default context. When TLS connection is created subsequent call of `stream_context_set_option()` function may modify default (global) context.


esler:
fixes #742 and #714 (probably)

tillkruss:
Nice catch.

PR #746, opened 2022-02-17
throw exception if connection->read() returns an error object in cons…

hauptmedia:
Hi,

In Consumer.php: $client->getConnection()->read() can return an error object which should be handled accordingly. This happens for example if psubscribe failed (e.g. empty topic list was provided).

Best regard,

Julian


tillkruss:
Thanks for submitting this. Please add a test case as well.

PR #747, opened 2022-02-25
Add a more precise blpop definition for static analysers

fluffycondor:
This commit makes popular static analysers (tested with Psalm and PHPStan) much happier by letting them know one can safely assume there are keys 0 and 1 in the array returned from blpop(), while IDEs can still use an old definition.

### DIFF
--- a/src/ClientInterface.php
+++ b/src/ClientInterface.php
@@ -83,6 +83,7 @@ use Predis\Profile\ProfileInterface;
  * @method array       hvals($key)
  * @method int         hstrlen($key, $field)
  * @method array|null  blpop(array|string $keys, $timeout)
+ * @psalm-method array{0: string, 1: string}|null blpop(array<string>|string $keys, int<0, max> $timeout)
  * @method array|null  brpop(array|string $keys, $timeout)
  * @method string|null brpoplpush($source, $destination, $timeout)
  * @method string|null lindex($key, $index)


tillkruss:
Is this specific to psalm.dev?

mfn:
Yes, phpstan also parses `psalm-method`, see https://github.com/phpstan/phpstan-src/blob/0808a89469a0d28bebd0b8d7b14af1639e441215/src/PhpDoc/PhpDocNodeResolver.php#L142

My question would be: why not just go with changing the existing `@method`?

Which software does not support array shapes in the docs?

phpstorm, psalm and phpstan all do and these are the most relevant.

fluffycondor:
> My question would be: why not just go with changing the existing `@method`?
> Which software does not support array shapes in the docs?

I believe PHPStorm currently doesn't support array shapes as return value in `@method` annotations. Unfortunately, if I remove `psalm-` prefix, PHPStorm can't parse it and the IDE claims the method not found while calling it. I'm using PHPStorm 2021.3.2.
image
Also there is an int range, I doubt it's supported too.
And I'm worried about users of another IDEs.
So I think a fallback is necessary, maybe I'm wrong.

tillkruss:
Frankly, I don't want to 3rd party syntax and prefer to stick with PHPDoc.

fluffycondor:
> I don't want to 3rd party syntax and prefer to stick with PHPDoc

I'd say `psalm-`prefixed annotations actually is a de-facto standard now. E.g. doctrine/orm heavily uses it: https://github.com/doctrine/orm/search?q=psalm

tillkruss:
I don't want to maintain this.

PR #748, opened 2022-03-09
PHP 8.1 and symfony/error handler deprecations

andrew-demb:
Using predis with `symfony/error-handler:5.4+` and PHP 8.1 leads to many triggered deprecation errors (from `symfony/error-handler`) like a:

```
2022-03-09T16:04:49+00:00 [info] User Deprecated: Method "Iterator::rewind()" might add "void" as a native return type declaration in the future. Do the same in implementation "Predis\Collection\Iterator\CursorBasedIterator" now to avoid errors or add an explicit @return annotation to suppress this message.
2022-03-09T16:04:49+00:00 [info] User Deprecated: Method "Iterator::current()" might add "mixed" as a native return type declaration in the future. Do the same in implementation "Predis\Collection\Iterator\CursorBasedIterator" now to avoid errors or add an explicit @return annotation to suppress this message.
2022-03-09T16:04:49+00:00 [info] User Deprecated: Method "Iterator::key()" might add "mixed" as a native return type declaration in the future. Do the same in implementation "Predis\Collection\Iterator\CursorBasedIterator" now to avoid errors or add an explicit @return annotation to suppress this message.
2022-03-09T16:04:49+00:00 [info] User Deprecated: Method "Iterator::next()" might add "void" as a native return type declaration in the future. Do the same in implementation "Predis\Collection\Iterator\CursorBasedIterator" now to avoid errors or add an explicit @return annotation to suppress this message.
2022-03-09T16:04:49+00:00 [info] User Deprecated: Method "Iterator::valid()" might add "bool" as a native return type declaration in the future. Do the same in implementation "Predis\Collection\Iterator\CursorBasedIterator" now to avoid errors or add an explicit @return annotation to suppress this message.
```

They may be fixed by using utility script from `symfony/error-handler` - https://github.com/symfony/symfony/blob/v5.4.6/src/Symfony/Component/ErrorHandler/Resources/bin/patch-type-declarations

I applied fixed from this script and added some manual fixes over changes:
- ensure return type of `\Iterator::key()` implementations as a `int` (instead of possible `int|null`)
- narrow down return type in some methods from `mixed` to more specific (example - `ProcessorInterface|null`)
- dropped return types from methods `__sleep` as there is no interface for this method in PHP with future signature change (they were automatically added by script - `@return array`)

There are one open issue with proper return type - https://github.com/predis/predis/blob/ab0c46332c1f5956e93693586123017dbf62198e/src/PubSub/AbstractConsumer.php#L181-L192
PHP will change return type for `\Iterator::rewind()` to `: void` in PHP 9, so this code won't be valid with next PHP major update.
I left this without changes for initiate a discussion - what I need to do about this method signature - possibly break BC with changing return (drop return a value) or just specify "invalid" return type in phpdoc (`@return void`) to suppress error from `symfony/error-handler`

### DIFF
--- a/src/Collection/Iterator/CursorBasedIterator.php
+++ b/src/Collection/Iterator/CursorBasedIterator.php
@@ -138,6 +138,7 @@ abstract class CursorBasedIterator implements \Iterator
 
     /**
      * {@inheritdoc}
+     * @return void
      */
     #[\ReturnTypeWillChange]
     public function rewind()
@@ -148,6 +149,7 @@ abstract class CursorBasedIterator implements \Iterator
 
     /**
      * {@inheritdoc}
+     * @return mixed
      */
     #[\ReturnTypeWillChange]
     public function current()
@@ -157,6 +159,7 @@ abstract class CursorBasedIterator implements \Iterator
 
     /**
      * {@inheritdoc}
+     * @return int
      */
     #[\ReturnTypeWillChange]
     public function key()
@@ -166,6 +169,7 @@ abstract class CursorBasedIterator implements \Iterator
 
     /**
      * {@inheritdoc}
+     * @return void
      */
     #[\ReturnTypeWillChange]
     public function next()
@@ -187,6 +191,7 @@ abstract class CursorBasedIterator implements \Iterator
 
     /**
      * {@inheritdoc}
+     * @return bool
      */
     #[\ReturnTypeWillChange]
     public function valid()
--- a/src/Collection/Iterator/ListKey.php
+++ b/src/Collection/Iterator/ListKey.php
@@ -127,6 +127,7 @@ class ListKey implements \Iterator
 
     /**
      * {@inheritdoc}
+     * @return void
      */
     #[\ReturnTypeWillChange]
     public function rewind()
@@ -137,6 +138,7 @@ class ListKey implements \Iterator
 
     /**
      * {@inheritdoc}
+     * @return mixed
      */
     #[\ReturnTypeWillChange]
     public function current()
@@ -146,6 +148,7 @@ class ListKey implements \Iterator
 
     /**
      * {@inheritdoc}
+     * @return int
      */
     #[\ReturnTypeWillChange]
     public function key()
@@ -155,6 +158,7 @@ class ListKey implements \Iterator
 
     /**
      * {@inheritdoc}
+     * @return void
      */
     #[\ReturnTypeWillChange]
     public function next()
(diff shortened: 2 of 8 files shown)


tillkruss:
I'm not familiar with `symfony/error-handler`. Why is it complaining about docblocks in another library?

andrew-demb:
@tillkruss here is some discussions about `symfony/error-handler` decisions about return types:
- https://github.com/symfony/symfony/issues/44802
- https://github.com/sebastianbergmann/php-file-iterator/pull/74#issuecomment-984546983

TLDR: Symfony "forces" the PHP ecosystem to make typing more explicit and allows to be sure that library/application will be ready for future PHP changes (like a changing return type). My TLDR may be a bit wrong, but here it is as I understood it before.

tillkruss:
Aha, thanks. Does that library allow ignoring certain packages from the analysis?

andrew-demb:
No, probably. I didn't find a way to do this.

On the other hand these additions makes more clear methods return types and plan to upgrade compatibility with PHP 9 (convert phpdoc return types to the native ones)

tillkruss:
I find this quite annoying. I don't want to maintain docblocks to please some library doing future proofing.

- Can `{@inheritdoc}` be combined with `@return void`? Is this documented somewhere?
- Do all common code editors support this syntax?

andrew-demb:
> Can {@inheritdoc} be combined with @return void? Is this documented somewhere?

My observations indicate that yes, it is. As of https://phpdoc.org, `@inheritdoc` are intended just to "copy" **description** [1] and other information should be inherited (example with information about methods phpdoc inheritance [2])

So combination `@inheritdoc` with any other tags are meant to override some inherited information (like a return type via `@return`)

I can't find any other more clear documentation about this.

> Do all common code editors support this syntax?

Checked with PhpStorm, VSCode. 
Unfortunately, I can't find a way to make Atom work for me (how to configure it properly for PHP with the preview quick doc feature).

<details>
  <summary>Screenshots</summary>

  phpstorm_after
  vscode_after

</details>
_____

[1] https://docs.phpdoc.org/guide/guides/inheritance.html#the-inheritdoc-tag  ("The inline tag {@inheritDoc} is used in a description to import the description of a parent element, even if the child element already has a description.")

[2] https://docs.phpdoc.org/guide/guides/inheritance.html#methods ("... When a superclass of the current class contains a method with the same name (hence, this method is re-defined) then the following information is inherited from that overridden method: ... The following tags: ... param, return")

tillkruss:
@andrew-demb Excellent work 🏅

PR #750, opened 2022-04-05
Update .gitattributes

angeljqv:
Make the vendor 56kb smaller,
is it possible or is there any reason not to do it?

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -11,3 +11,5 @@
 /.travis.yml export-ignore
 /phpunit.xml.dist export-ignore
 /phpunit.xml.travisci export-ignore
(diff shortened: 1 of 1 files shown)


tillkruss:
We can certainly do that.

mfn:
👎🏼 on removing CHANGELOG from the installed vendor files, it's very useful to be able to quickly check the latest changes.

tillkruss:
@mfn: If you want to open a PR to split up the changeling like Laravel does, that's cool.

The `v1.1` branch should only contains `v1.1.x` release notes.

angeljqv:
split is a great option too 👍

PR #752, opened 2022-04-26
Fix `getDefault` return type-hint

BafS:
`getDefault` returns ProfileInterface|null


tillkruss:
Please post the return values of `getDefault()` for all possible options.

BafS:
Not sure what you mean but it seems that it can be more than ProfileInterface, depending the handler. It's probably safer to let it like this because there is not a common interface shared among handlers defaults.

PR #753, opened 2022-04-28
Use env var to get changed FLUSHDB command

newelement:
It's common to change certain commands in Redis for security reasons. When using Laravel and running artisan command cache:clear the command will throw an error because FLUSHDB command was changed. My change uses an optional env variable to set the changed FLUSHDB command. I'm open to doing this for other critical command if necessary.

### DIFF
--- a/src/Command/ServerFlushDatabase.php
+++ b/src/Command/ServerFlushDatabase.php
@@ -23,6 +23,6 @@ class ServerFlushDatabase extends Command
      */
     public function getId()
     {
-        return 'FLUSHDB';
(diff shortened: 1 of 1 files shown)


tillkruss:
I'd suggest wrapping Predis and supplying your own `PredisConnection` with Laravel.

newelement:
I'm not sure how this is a Laravel only issue, but ok...

tillkruss:
It's not a Laravel-only issue, I'm just letting you know how to solve it on your end.

Using `getenv()` is not acceptable, Predis isn't an app. If you want to submit a PR that adds support for passing in various aliases when instantiating the class, please do.

Issue #754, opened 2022-05-02
predis incompatible with Sentinel in Redis 7


rhl-jfm:
In a Redis Sentinel cluster, predis still uses the SENTINEL SLAVES command to find replicas. This command was deprecated in Redis 6 (replaced by SENTINEL REPLICAS command). The SENTINEL SLAVES command was removed entirely in Redis 7.

In order to work with Redis 7+ Sentinel clusters, predis should be updated to use the SENTINEL REPLICAS command instead.

tillkruss:
Right on. Can you submit a PR with a possible solution?

tillkruss:
https://github.com/redis/redis/pull/10723

PR #756, opened 2022-05-02
redis 7 sentinel support #754 

axeltomasson:


src/Connection/Aggregate/SentinelReplication.php
```diff
@@ -381,7 +381,7 @@ protected function querySentinelForSlaves(NodeConnectionInterface $sentinel, $se
         $slaves = array();
 
         $payload = $sentinel->executeCommand(
-            RawCommand::create('SENTINEL', 'slaves', $service)
+            RawCommand::create('SENTINEL', 'replicas', $service)
```
tillkruss, line 384: This will break compatibility with older versions.

tillkruss (changes requested):

src/Connection/Aggregate/SentinelReplication.php
axeltomasson: Yeah sorry of course did not think of that.  Was fixated on that everything older then 7.0.x, 6.0.x and 6.2.x was EOL.

tillkruss: I wish 🙃

axeltomasson: Misuse profile to signal 7.0.0 version from config time?

~~~
        if (version_compare($sentinel->getParameters()->profile, '7.0') >= 0) {
            $payload = $sentinel->executeCommand(
                RawCommand::create('SENTINEL', 'replicas', $service)
            );
        } else {
            $payload = $sentinel->executeCommand(
                RawCommand::create('SENTINEL', 'slaves', $service)
            );
        }
~~~

rhl-jfm: > Yeah sorry of course did not think of that. Was fixated on that everything older then 7.0.x, 6.0.x and 6.2.x was EOL.

From a brief review of the Redis sources, SENTINEL REPLICAS is supported from Redis 5.0.

PR #759, opened 2022-05-05
Use SENTINEL REPLICAS instead of SENTINEL SLAVES, closes #754

### LINKED ISSUE #754: predis incompatible with Sentinel in Redis 7

rhl-jfm:
Use SENTINEL REPLICAS instead of SENTINEL SLAVES, which was removed in Redis 7.

Trivial implementation. This breaks compatibility with Redis Sentinel setups based on Redis < 5.0. If this is not acceptable, the code will need some refactoring to support all Redis versions.

### DIFF
--- a/src/Command/ServerSentinel.php
+++ b/src/Command/ServerSentinel.php
@@ -45,7 +45,7 @@ class ServerSentinel extends Command
     }
 
     /**
-     * Returns a processed response to SENTINEL MASTERS or SENTINEL SLAVES.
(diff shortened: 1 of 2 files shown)


rhl-jfm:
Duplicate of PR #756.

Issue #512, opened 2018-07-19
@method annotations in ClientInterface.php are not exact


arnaud-lb:
For example, the get() method is typed like this: `@method string get($key)`.

This is not exact, because, the method returns null if the key does not exist.

This can lead some IDEs or static analysers (like phpstan) to think that its return value is never null.

rannmann:
It's not just the return values either.  The params are wrong in a lot of cases too.

```
 * @method int    rpush($key, array $values)
```

But when you look at the tests, it doesn't take an array at all.

```
    public function testReturnsElementAtIndex()
    {
        $redis = $this->getClient();

        $redis->rpush('letters', 'a', 'b', 'c', 'd', 'e');

        $this->assertSame('a', $redis->lindex('letters', 0));
        $this->assertSame('c', $redis->lindex('letters', 2));
        $this->assertNull($redis->lindex('letters', 100));
    }
```

Every single instance of calling rpush is an error in my IDE.  :(

nrk:
@arnaud-lb indeed, thanks for the suggestion.

@rannmann it [**does** take an array](https://github.com/nrk/predis/blob/v1.1.1/tests/Predis/Command/ListPushTailTest.php#L50-L62) too, just like many other methods for variadic redis commands. In the past we had an [extremely long discussion](https://github.com/nrk/predis/issues/89) about how to support method annotations, a long-time request at that time, in Predis (which by the way supports different method signatures for Redis commands: one that tries to stick closely with the Redis docs and one that tries to use a more PHP-friendly approach) and this is [the relevant comment](https://github.com/nrk/predis/issues/89#issuecomment-31068030) that summarise the final decision on the matter. Is there a better solution? Not sure, at least not until I see a different proposal or a PR.

rannmann:
@nrk I don't mean to hijack this ticket, but that linked comment about PHPStorm supporting the syntax is no longer true, so it might be worth revisiting.

I was able to make the warnings go away by changing the definition to:

```
 * @method int    rpush(string $key, array|string... $values)
```

But this isn't quite right since it would error if you passed multiple arrays.  Maybe `array|string` would be a better middle ground since it would still error if you passed multiple arrays, but the common usage of `rpush($key, 'value');` would no longer trigger warnings.

Seldaek:
[...]

PR #762, opened 2022-05-11
Update @method tags

Seldaek:
Made keys strings, added some variadics in the form of `string[]|string $keyOrKeys, string ...$keys = null` instead of simple `array $keys` in a few places.

Probably tons more could be done, but it's a long list and I rather not mess up, so this is a start :)

Refs #512


tillkruss:
Thanks for submitting this, can you send it to `main`? I'm gonna put the v1.1 branch on maintenance soon.

Seldaek:
Ok sure, will rebase tomorrow

Seldaek:
OK rebased and added a bunch more types

PR #764, opened 2022-05-12
Testing code against Redis 7

zaherg:
This PR make sure that tests will pass against Redis 7.

PS: this PR will just make tests pass against redis 7, it does not add any new feature or changes which was implemented in redis 7 like the changes in the replication naming.

### DIFF
--- a/tests/Predis/Command/Redis/CONFIG_Test.php
+++ b/tests/Predis/Command/Redis/CONFIG_Test.php
@@ -143,7 +143,7 @@ class CONFIG_Test extends PredisCommandTestCase
     public function testThrowsExceptionWhenSettingUnknownConfiguration(): void
     {
         $this->expectException('Predis\Response\ServerException');
-        $this->expectExceptionMessage('ERR Unsupported CONFIG parameter: foo');
+        $this->expectExceptionMessageMatches("/ERR|CONFIG|foo|arguments|parameter/");
 
         $redis = $this->getClient();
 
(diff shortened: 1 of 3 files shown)


tests/Predis/Command/Redis/CONFIG_Test.php
```diff
@@ -143,7 +143,7 @@ public function testReturnsTrueOnSuccessfulConfiguration(): void
     public function testThrowsExceptionWhenSettingUnknownConfiguration(): void
     {
         $this->expectException('Predis\Response\ServerException');
-        $this->expectExceptionMessage('ERR Unsupported CONFIG parameter: foo');
+        $this->expectExceptionMessageMatches("/ERR|CONFIG|foo|arguments|parameter/");
```
tillkruss, line 146: Should we maybe have two full sentences here? That will make it easier to read.

tillkruss (changes requested):

tests/Predis/Command/Redis/CONFIG_Test.php
zaherg: The sentence is different between 7 and anything lower so instead I was aiming to check that the key words are there which will work on all versions.

If you prefer to have a full sentence, I can do a small version check and test its content based on redis version

tillkruss: Yeah, I think `(old error message|new error msg)` will be less difficult to read.

tillkruss: Even better!

PR #766, opened 2022-05-15
[Fix] Upgrade style code

zaherg:
- This PR upgrade the code style check (PHP-CS-Fixer) from `v2` to `v3`, while also adding a new workflow to check the code style.

- It is based on the original style rules.

- Since `php-cs-fixer v3` requires `php-7.4` or higher I have not added it to `composer.json`.

Also, I ran the check to auto fix all the files (which explains the huge edited files).


zaherg:
It looks like `git` has failed while checking out the code, if we can rerun the workflow that would be great.

zaherg:
While the code style check works at my repo https://github.com/zaherg/predis/runs/6442288106?check_suite_focus=true it fails here 🤷‍♂️

PR #768, opened 2022-05-27
Allow PubSub / MultiExec with Replication

josiasmontag:
This is a rebase of #514 for the v2 branch. It allows to use PUB/SUB and MULTI/EXEC with a Redis Master/Slave replication setup. 

Currently, Predis throws an exception when using these commands with any _aggregate_ connection. This exception should only be thrown for _cluster_ connections but not for _replication_ connections.

I have been using this patch for several years in a Redis master/slave setup and it is working fine.

### DIFF
--- a/src/Monitor/Consumer.php
+++ b/src/Monitor/Consumer.php
@@ -12,7 +12,7 @@
 namespace Predis\Monitor;
 
 use Predis\ClientInterface;
-use Predis\Connection\AggregateConnectionInterface;
+use Predis\Connection\Cluster\ClusterInterface;
 use Predis\NotSupportedException;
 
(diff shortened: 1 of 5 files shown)


josiasmontag:
```php
        $parameters = ['tcp://127.0.0.1?role=master', 'tcp://127.0.0.1'];
        $options    = ['replication' => 'predis'];

        $client = new Client($parameters, $options);

        $client->transaction(function (MultiExec $tx) {
            $tx->set('foo', 'bar');
        });

        $this->assertSame('bar', $client->get('foo'));
```

Without this patch it throws an exception, with this patch it will just work fine.

Issue #779, opened 2022-06-27
srem method should accept $members as array


brightq:
**Describe the bug**
In v2.0, srem has been defined as following in ClientInterface.php
```
 * @method int         srem(string $key, string $member)
```

This should be 
```
 * @method int         srem(string $key, array|string $member)
```

**To Reproduce**
Pass an array arguments to srem method will cause PHP type error `Parameter #2 $member of method Predis\ClientInterface::srem() expects string, array given.`

**Expected behavior**
Array should be accepted as per https://redis.io/commands/srem/
it says Starting with Redis version 2.4.0: Accepts multiple member arguments.

**Versions (please complete the following information):**
 - Predis: [2.0]

**Code sample**

**Additional context**

PR #780, opened 2022-06-27
bugfix for srem  type to accept array #779

brightq:


tillkruss:
Can you add a unit test for this as well, please?

PR #784, opened 2022-07-12
Fix sentinel reconnect issue on long-running PHP processes

Linc70J:
I added a test case to make sure that sentinel connections can be reconnected after restarting sentinel on a long-running PHP process.
Also, I fixed in #681 that there are no sentinel servers available for autodiscover on Swoole.


tillkruss:
LGTM 👍

dennishoffmann-edu:
Sry for asking, but is a new release planned in the next days?  This fix would be important, but that maybe only concerns me :D

dennishoffmann-edu:
Thanks for the quick response. Yeah, that would have been my second choice :D.

Issue #783, opened 2022-07-08
Sentinel without AUTH and replicas with AUTH


1thew:
**Describe the bug**
I'm trying to connect sentinel
Sentinel does not have authentication, but replicas do have a password.
Can't create client that uses password only for Redis replica

**To Reproduce**
```
        $sentinels = ["tcp://my_sentinel:26379"];

        $client = new Client($sentinels, [
            "replication" => "sentinel",
            "service" => "test",
            "parameters" => [
                "password" => "mypassword",
                "database" => "0",
            ],
        ]);

        $client->connect();
```
If config have  "password"  -- get an error "No sentinel server available for autodiscovery". Client send AUTH to sentinel connections.
If config dont have "password" :
`Predis\Connection\ConnectionException : `SELECT` failed: NOAUTH Authentication required. [tcp://172.16.101.3:6379]
`
Client dont use AUTH for replica.
.

**Versions (please complete the following information):**
 - Predis: 2.
 - PHP 8.1
 - Redis Server  6.2.
 - Ubuntu

Maybe I'm doing something wrong?

tillkruss:
It's a known issue. Pull Requests welcome. See #782.

dakorpar:
If there's no password on sentinels then you need to set password in url as empty string, otherwise it will use default redis password for sentinels and fail. This is breaking change from 1.1.10 to 2.0.0, try something like this:
```
$sentinels = array(
    'tcp://sentinel1:26379?password=',
    'tcp://sentinel2:26379?password=',
    'tcp://sentinel3:26379?password=',
);

$client = new Predis\Client($sentinels, array(
    'replication' => 'sentinel',
    'service' => 'mymaster',
    'parameters' => ['database'=> 0, 'password' => 'redis-password']
));
```

tillkruss:
I don't have the bandwidth to test Sentinel setups. I'm open to PRs with test cases and detailed explanations.

waqas-badar:
I was facing this issue. My issue got resolved after increasing the timeout.

I face another issue, while investigating, it is found that in SentinelReplication.php we are using SENTINEL slaves command which is not supported in new version IMHO. I tried with sentinel help command in redis-cli and unable to find it. When i switch the command to SENTINELS then it work. Not sure either it was the exact fix or not.

tillkruss:
The missing command was re-added in 7.0.1: https://github.com/redis/redis/pull/10723

mta59066:
[...]

PR #787, opened 2022-07-20
Do not use password for sentinel

dakorpar:
After update to 2.0 it's not possible to connect to sentinel server that worked fine with 1.1.10. I went over 

#783 and #782 fixed unit tests there and I'm opened up if something more needs to be created (maybe adding some new option rather to use sentinel auth or no)

### DIFF
--- a/src/Connection/Replication/SentinelReplication.php
+++ b/src/Connection/Replication/SentinelReplication.php
@@ -268,6 +268,7 @@ class SentinelReplication implements ReplicationInterface
             // in a later release.
             $parameters['database'] = null;
             $parameters['username'] = null;
+            $parameters['password'] = null;
 
             if (!isset($parameters['timeout'])) {
                 $parameters['timeout'] = $this->sentinelTimeout;
--- a/tests/Predis/Connection/Replication/SentinelReplicationTest.php
+++ b/tests/Predis/Connection/Replication/SentinelReplicationTest.php
@@ -38,7 +38,7 @@ class SentinelReplicationTest extends PredisTestCase
     /**
      * @group disconnected
      */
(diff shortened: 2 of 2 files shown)


tillkruss:
We need a test that shows Sentinel works with and without password authentication, no?

dakorpar:
So I would need t sentinel on github actions with different auth and run tests against config with password set on sentinel and without password? Do U have already something that can save me some time at least? Did I understand this well?

tillkruss:
Or just Mock it. Right now you're changing an existing test case, which is too risky.

dakorpar:
My problem is that test that I've changed doesn't really proove anything, just that password is still there and that doesn't actually work on real sentinel setup, so test itself is probably wrong, maybe it works on different version/configuration of sentinel....

dakorpar:
I'm closing this since with this options it works... maybe it would be cool to add that to docs because is not really clear...
If there's no password on sentinels then you need to set password in url as empty string, otherwise it will use default redis password for sentinels and fail. This is breaking change from 1.1.10 to 2.0.0
```
$sentinels = array(
    'tcp://sentinel1:26379?password=',
    'tcp://sentinel2:26379?password=',
    'tcp://sentinel3:26379?password=',
);

$client = new Predis\Client($sentinels, array(
    'replication' => 'sentinel',
    'service' => 'mymaster',
    'parameters' => ['database'=> 0, 'password' => 'redis-password']
));
```

rmertins:
We are running around 4 sentinel replication cluster in production for over three years now.
And we had never ever bevor to auth with the sentinel. Version 2.0.0 of Predis is breaking everything and this workaround doesn't feel good to mee. Upvote for a merge.

tillkruss:
@rmertins I'll gladly accept any PR with tests.

PR #791, opened 2022-08-05
Make sure to close connection before finalizing

zexa:


tillkruss:
No tests, no explanation. Send it to v2.

zexa:
Sorry, didn't mean to push this request to your branch yet. I'm still testing things out. 

We recently found out that one of our servers has 512 dangling redis connections (same as our fpm count) and we thought that this might be the reason.

tillkruss:
Oh yikes. Feel free to resend to v2.0 with a test.

zexa:
Hello again,

Just wanted to share the results of our tests. Indeed, after closing the connections manually via the code provided in this PR we did see a significant decrease in our connection count with our Redis servers, however, our CPU usage increased on our fpm servers significantly as well.

We're guessing that by not sending a close connection request the operating system manages to optimize getting the connection somehow (we're not knowledgeable enough to tell whether the os actually does this).

Since we're currently optimizing for CPU consumption, this PR is not good enough for us as it is, and I don't know what I want to do with it. I guess this could be made into a config (i.e. the user could specify if they wish to close the connection on finalization or not). Though if we were to make this into a config we should also decide on default functionality (close by default or leave it?)

Let me know what you think. If you think this is something that's useful in Predis I might still provide a PR with tests and such.

PR #797, opened 2022-09-06
improve PHP 8.2 support

drealecs:
This builds upon #796 and removes the custom logic implemented there in `process()`, by fixing the commands string as suggested in https://wiki.php.net/rfc/deprecate_partially_supported_callables#backward_incompatible_changes


tillkruss:
Just out of curiosity, why not keep it the way @driesvints did it?

drealecs:
> Just out of curiosity, why not keep it the way @driesvints did it?

Not really important but just because that contains some extra logic that can be avoided, specifically the call to `strpos`, `explode` that kind of always runs.

In construct we build strings with `'static::<method>'` just so that we will replace them with `[$this, '<method>']` before execution. I mean, we could build them directly like that.

Also, as those are static methods I believe it would have been a bit more accurate to have them as `[static::class, '<method>']`.

src/Command/Processor/KeyPrefixProcessor.php
```diff
@@ -31,142 +31,154 @@ class KeyPrefixProcessor implements ProcessorInterface
     public function __construct($prefix)
     {
         $this->prefix = $prefix;
-        $this->commands = array(
+
+        $prefixFirst = static::class . '::first';
```
drealecs, line 35: We could also have here now the callables as:
```php
$prefixFirst = static function (CommandInterface $command, $prefix) {
    static::first($command, $prefix);
};
```
that can be upgraded in PHP 8.1+ to:
```php
$prefixFirst = static::first(...);
```

driesvints:
I actually didn't change that commands array because I thought it could be a breaking change if the contents were changed. But if the authors here are okay with it then go for it 👍

drealecs:
> I actually didn't change that commands array because I thought it could be a breaking change if the contents were changed. But if the authors here are okay with it then go for it 👍

Thanks for explaining.
However, `KeyPrefixProcessor::$commands` is private and there are no getters so I can't see how that would be a breaking change.

PR #799, opened 2022-09-16
check resource before writing

### LINKED ISSUE #793: supplied resource is not a valid stream resource
**Describe the bug**
predis/predis/src/Connection/StreamConnection.php:291 is throwing an exception.
Except this, predis seems to be working as expected.
```
fwrite(): supplied resource is not a valid stream resource {"exception":"[object] (TypeError(code: 0): fwrite(): supplied resource is not a valid stream resource at /predis/predis/src/Connection/StreamConnection.php:291)
```

**To Reproduce**
Use predis with PHP 8.1.6

**Expected behavior**
It should not throw exception even if the resource is already closed.

 - Predis: 1.1.9
 - PHP 8.1.6

**Additional context**
I think it is throwing an exception because the fwrite behavior was changed in PHP8.
https://github.com/php/doc-en/issues/1581

stebogit:
Fixes #793 

@tillkruss I was not able to add any test, actually I was not even able to run any tests because of this error. I must be missing something or doing something really stupid, but I could not find any instructions on how to set up the environment for running tests. I simply installed Redis (via brew) and the composer dependencies, but this is the really unexpected result... 🤔 

Screen Shot 2022-09-16 at 10 09 18 AM

Any tips? I found unusual not seeing instructions on how to set up the dev environment, I'd be happy to contribute to add to the docs and then add tests to this PR. Also, it seems like #793 is going to [occur on v2 as well](https://github.com/predis/predis/blob/main/src/Connection/StreamConnection.php#L263), so Id be happy to push a PR for that as well if you want.

### DIFF
--- a/src/Connection/StreamConnection.php
+++ b/src/Connection/StreamConnection.php
@@ -289,7 +289,7 @@ class StreamConnection extends AbstractConnection
         $socket = $this->getResource();
 
         while (($length = strlen($buffer)) > 0) {
-            $written = @fwrite($socket, $buffer);
+            $written = is_resource($socket) ? @fwrite($socket, $buffer) : false;
 
             if ($length === $written) {
                 return;


src/Connection/StreamConnection.php
```diff
@@ -289,7 +289,7 @@ protected function write($buffer)
         $socket = $this->getResource();
```
stebogit, line 289: The fix could probably be done here skipping the loop entirely, but I wasn't sure since I could not run tests:
```php
$socket = $this->getResource();
if (!is_resource($socket)) {
    return;
}
```

tillkruss:
Thanks! Can you send this to the v2.0 `main` branch. We don't maintain v1.1 anymore.

stebogit:
@tillkruss will do in a few.
Though I'm a bit surprised because Laravel currently still [relies on `predis/predis ~1.0` ](https://laravel.com/docs/master/queues#other-driver-prerequisites) (that's where I have issues).
Any chance you might merge this as well?

tillkruss:
I can merge it, but I don't know if we'll have any more releases.

@driesvints What Laravel version is switching to Predis v2.0 and will support PHP 8.2?

driesvints:
Laravel v9 & v10 will support PHP 8.2. We could maybe check into making v2 the minimum version for Laravel v10. Is there anything that needs to be changed except dropping the constraint?

driesvints:
https://github.com/laravel/framework/pull/44209

tillkruss:
I'll merge this as well. Maybe we need one more v1.1 tag just to keep Laravel 9 happy.

stebogit:
me happy, thanks 😄

stebogit:
@tillkruss though wouldn't it be best if the new release also included the same [fix for `fread`](https://github.com/predis/predis/pull/802) as well?

tillkruss:
> though wouldn't it be best if the new release also included the same [fix for `fread`](https://github.com/predis/predis/pull/802) as well?

Yes, want to open a PR?

stebogit:
sure, will do

PR #800, opened 2022-09-16
check resource before writing, w/ test

### LINKED ISSUE #793: supplied resource is not a valid stream resource

stebogit:
Fixes #793 for v2 (ref #799)

@tillkruss I added a test and ran them all with both PHP v7.4 and 8.0, all passed

### DIFF
--- a/src/Connection/StreamConnection.php
+++ b/src/Connection/StreamConnection.php
@@ -260,7 +260,7 @@ class StreamConnection extends AbstractConnection
         $socket = $this->getResource();
 
         while (($length = strlen($buffer)) > 0) {
-            $written = @fwrite($socket, $buffer);
+            $written = is_resource($socket) ? @fwrite($socket, $buffer) : false;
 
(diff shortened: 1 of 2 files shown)


tillkruss:
Thank you!

PR #802, opened 2022-09-18
Check if socket is resource before reading

streamingsystems:
Per your request, here is the PR for checking if the socket is a resource prior to reading.

### DIFF
--- a/src/Connection/StreamConnection.php
+++ b/src/Connection/StreamConnection.php
@@ -304,7 +304,7 @@ class StreamConnection extends AbstractConnection
                 $bytesLeft = ($size += 2);
 
                 do {
-                    $chunk = fread($socket, min($bytesLeft, 4096));
+                    $chunk = is_resource($socket) ? fread($socket, min($bytesLeft, 4096)) : false;
 
                     if ($chunk === false || $chunk === '') {
                         $this->onConnectionError('Error while reading bytes from the server.');


tillkruss:
This can't hurt and won't impact performance that much.

streamingsystems:
This is a dumb question but I already have 2.02 in my composer.lock and I can't seem to get this file change down. Do you need to bump to 2.03 so I will get the change?

mfn:
> Do you need to bump to 2.03 so I will get the change?

Yes, a new release needs to be done for it to appear.

Note: you can try out releases before they are done by [using `2.0-dev` as version](https://github.com/predis/predis/blob/42c33a406f1fc8f63dc91a2cc32a5fd832659ec6/composer.json#L47) (at your own risk)

tillkruss:
Wouldn't it be `dev-main`?

streamingsystems:
Thanks. Do you know when your next release is planned? We would like to get this fix into production as soon as we can. We have other packages that reply on predis that we moved to “^2.0” in composer from 1.1.

From: Markus Podar ***@***.***>
Date: Monday, September 19, 2022 at 1:48 AM
To: predis/predis ***@***.***>
Cc: streamingsystems ***@***.***>, Author ***@***.***>
Subject: Re: [predis/predis] Check if socket is resource before reading (PR #802)

Do you need to bump to 2.03 so I will get the change?

Yes, a new release needs to be done for it to appear.

Note: you can try out releases before they are done by using 2.0-dev as version<https://github.com/predis/predis/blob/42c33a406f1fc8f63dc91a2cc32a5fd832659ec6/composer.json#L47> (at your own risk)

tillkruss:
@streamingsystems: Tagged v2.0.3 just now.

PR #803, opened 2022-09-19
Check resource before reading

stebogit:
Ref https://github.com/predis/predis/pull/799#issuecomment-1251506706


stebogit:
thank @tillkruss, are you going to push a new `v1.1.11` release?

stebogit:
ping @tillkruss

tillkruss:
@stebogit Yeah at some point for sure, I'll see if anything else comes in with 8.2 that has to be fixed in v1.1.

PR #804, opened 2022-09-27
Add partial support for Stream commands

TobiasBengtsson:
Adds support for the following commands not related to groups/consumers: XADD, XDEL, XLEN, XRANGE, XREVRANGE, XTRIM.


tillkruss:
Nice, can you fix the tests?

TobiasBengtsson:
Updated, apart from a couple of typos it was related to `stream-node-max-entries` config that we need to control for deterministic tests. I moved it from `setUp` and `tearDown` to the individual tests to just run them on supported versions.

The potential problem here is if you have one failure the config will never be reset for the test run (even worse locally: never). I could add try-catch blocks around the test, but if you can think of any better way to handle it I'm all ears

tillkruss:
I've relaxed the permissions, so workflows always run 🥲

tillkruss:
Thanks for the test fixes, I assume a re-run would resolve this. If they are too unable we can tweak them later. Thanks for the contribution!

PR #807, opened 2022-10-02
sentinel default to no password when replica authentication is used

mta59066:
In order to work with replica authentication but no sentinel authentication, empty password (`password=`) needs to be added to the sentinel uri but this is not documented and possibly not expected.

This pr removes the requirement for empty password and defaults to no password if it is not provided.

### DIFF
--- a/src/Connection/Replication/SentinelReplication.php
+++ b/src/Connection/Replication/SentinelReplication.php
@@ -268,6 +268,9 @@ class SentinelReplication implements ReplicationInterface
             // in a later release.
             $parameters['database'] = null;
             $parameters['username'] = null;
+            if (!isset($parameters['password'])) {
+                $parameters['password'] = null;
+            }
 
             if (!isset($parameters['timeout'])) {
                 $parameters['timeout'] = $this->sentinelTimeout;
--- a/src/Connection/Replication/SentinelReplication.php
+++ b/src/Connection/Replication/SentinelReplication.php
@@ -268,7 +268,10 @@ class SentinelReplication implements ReplicationInterface
             // in a later release.
             $parameters['database'] = null;
             $parameters['username'] = null;
-            if (!isset($parameters['password'])) {
+
+            // don't leak password from between configurations
+            // https://github.com/predis/predis/pull/807/#discussion_r985764770
+            if (! isset($parameters['password'])) {
                 $parameters['password'] = null;
             }
 
--- a/tests/Predis/Connection/Replication/SentinelReplicationTest.php
+++ b/tests/Predis/Connection/Replication/SentinelReplicationTest.php
@@ -62,6 +62,7 @@ class SentinelReplicationTest extends PredisTestCase
 
         $this->assertArrayNotHasKey('database', $parameters, 'Parameter `database` was expected to not exist in connection parameters');
         $this->assertArrayNotHasKey('username', $parameters, 'Parameter `username` was expected to not exist in connection parameters');
+        $this->assertArrayNotHasKey('password', $parameters, 'Parameter `password` was expected to not exist in connection parameters');
     }
 
     /**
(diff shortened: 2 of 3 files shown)


src/Connection/Replication/SentinelReplication.php
```diff
@@ -268,6 +268,9 @@ protected function createSentinelConnection($parameters)
             // in a later release.
             $parameters['database'] = null;
             $parameters['username'] = null;
+            if (!isset($parameters['password'])) {
+                $parameters['password'] = null;
```
tillkruss, line 272: Can you explain this particular change in-depth? Why? What does it do? Can you think of anything it would break?

mta59066: Predis uses the same code for both Sentinel and Redis authentication (or parameter parsing). Which leads to redis authentication settings leaking to Sentinel authentication settings [here](https://github.com/predis/predis/blob/v2.0.2/src/Connection/Factory.php#L166).

What is happening
- Predis user upgrades predis to a version that supports sentinel authentication from a version that did not support sentinel authentication
    - no authentication on redis or sentinel - everything works fine
    - authentication on redis but no authentication on sentinel - upgrade breaks and user needs to edit code and provide `password=` to sentinel connection string to prevent it from using the redis password on sentinel

Possible scenarios with proposed changes
- User experienced above problem and updated code to include `password=`, the proposed change will not cause a problem
- User had redis and sentinel authentication configured and provided the passwords in parameters given to predis, the proposed change will not cause a problem
- User unintentionally used the same password for both redis and sentinel and did not provide the password in sentinel connection parameters and relied on the leak of redis password to sentinel parameters, the proposed change will create a problem as it assumes no authentication if no password is given for sentinel

Possible actions to take
- The current re-use of redis password for sentinel password behavior and the need to explicitly set sentinel password to empty if there is no sentinel authentication can be documented and the code left as is
- The change proposed here will set password parameter for sentinel to null if it is not given
    - rationale is similar to the `database` and `username` parameters which are also set to null because there is no `database` parameter in sentinel and there was no `username` parameter (this will need to be reworked if user acl support is added to predis in the future) 
    - `timeout` which is a valid parameter for sentinel is also handled the same way as the proposed change to `password`

tillkruss: Thanks for the thorough explanation. Seems reasonable.

>User unintentionally used the same password for both redis and sentinel and did not provide the password in sentinel connection parameters and relied on the leak of redis password to sentinel parameters, the proposed change will create a problem as it assumes no authentication if no password is given for sentinel

Can you give me a code sample how the config would look right now and how it should look _after_ this merge? That way we can include quickly help out users experiencing this issue after the update.

mta59066: Currently the following code works (tested with redis-7.0.5 and predis-2.0.2)

- redis `requirepass "secret"`
- sentinel `requirepass "secret"`

```php
$sentinels = array(
    'tcp://127.0.0.1:26379',
);

$options = [
    'replication' => 'sentinel',
    'service' => 'mymaster',
    'parameters' => ['database'=> 0, 'password' => 'secret']
];

$client = new Predis\Client($sentinels, $options);
```

After this proposed change, the sentinel connection explicitly needs its own password
```php
$sentinels = array(
    'tcp://127.0.0.1:26379?password=secret',
);
```

Note: Because current tests are only doing redis or sentinel and not both there is not a representative test for this case, and the tests I added are not doing much. I'll try to improve them and send a commit tomorrow.

tillkruss: Thank you!

tillkruss: @mta59066 Do you want to push a test or two, so we can merge?

fengtan: > User unintentionally used the same password for both redis and sentinel and did not provide the password in sentinel connection parameters and relied on the leak of redis password to sentinel parameters, the proposed change will create a problem as it assumes no authentication if no password is given for sentinel

We have just upgraded predis from 2.0.2 to 2.0.3 and ran into this issue (things broke).

In our case, we *intentionally* used the same password for both redis and sentinel and did not provide a password in sentinel connection parameters to make the config easier to maintain (this allows to set the password in a single place).

We fixed the problem by appending `?password=xxx` to the sentinel connection strings are documented above.

Since this problem is expected, perhaps it could be documented in the [release notes](https://github.com/predis/predis/releases/tag/v2.0.3) along with instructions to resolve it, in case other people run into this issue.

PR #811, opened 2022-11-13
Adding the release drafter

chayim:
This pull request adds support for the release drafter - an automated tool to help **improve** (and not replace!) writing release notes. It takes the name of a given PR, and inserts it as an entry, into a proposed "release", to be managed within the releases tab of the repository.

The drafter makes use of the following labels - which would need to be configured in this repository. A good, practical output of the drafter assisting with releases can be viewed at [this version of redis-py](https://github.com/redis/redis-py/releases/tag/v4.4.0rc3).

- bug - If a PR is marked as a bug, it will show up as a bug fixed, in the release notes
- feature - When a PR is marked thus, it shows up as a feature
- breakingchange - While we never want them, of course they're unavoidable. This is a call out that something is a breaking change
- maintenance - For CI, documentation, and other things that people sued to call "chores", but are super valuable!
- experimental - Anything marked thus, is also called out so the community can know that it's both new, and not necessarily reliable.

@tillkruss WDYT?

### DIFF
--- a/.github/release-drafter-config.yml
+++ b/.github/release-drafter-config.yml
@@ -1,0 +1,48 @@
+name-template: '$NEXT_MINOR_VERSION'
+tag-template: 'v$NEXT_MINOR_VERSION'
+autolabeler:
+  - label: 'maintenance'
+    files:
+      - '*.md'
+      - '.github/*'
+  - label: 'bug'
+    branch:
+      - '/bug-.+'
+  - label: 'maintenance'
+    branch:
+      - '/maintenance-.+'
+  - label: 'feature'
+    branch:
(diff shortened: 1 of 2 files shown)


tillkruss:
I guess if this repo gets busy, sure. I can add those issue labels.

chayim:
Given our chat... I went ahead and added the labels, using this PR. WDYT?

chayim:
@tillkruss added you as a reviewer - and want to make sure we have the same process down. Assuming you approve, would you rather I squash+merge or you?  I don't want to step on toes.

tillkruss:
I've never used this thing, let's try it 👍

Issue #815, opened 2022-11-16
Add support for missing redis command ZDIFF


chayim:
predis is missing support for some of the sorted set functions, such as [zdiff](https://redis.io/commands/zdiff/). This command has been available since Redis 6.2.0

PR #826, opened 2022-11-23
Extend Sorted Set support by implementing ZDIFF command

vladvildanov:
#815 Added support for ZDIFF command to redis client

@dvora-h @sazzad16

### DIFF
--- a/src/ClientInterface.php
+++ b/src/ClientInterface.php
@@ -125,6 +125,7 @@ use Predis\Response\Status;
  * @method int               zadd(string $key, array $membersAndScoresDictionary)
  * @method int               zcard(string $key)
  * @method string            zcount(string $key, int|string $min, int|string $max)
+ * @method array             zdiff(int $numkeys, array $keys, bool $withScores = false)
  * @method string            zincrby(string $key, int $increment, string $member)
  * @method int               zinterstore(string $destination, array|string $keys, array $options = null)
  * @method array             zmscore(string $key, string ...$member)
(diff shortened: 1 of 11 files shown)


shacharPash (commented):
LGTM :+1:

src/ClientInterface.php
```diff
@@ -125,6 +125,7 @@
  * @method int               zadd(string $key, array $membersAndScoresDictionary)
  * @method int               zcard(string $key)
  * @method string            zcount(string $key, int|string $min, int|string $max)
+ * @method array             zdiff(int $numkeys, array $keys, bool $withScores = false)
```
sazzad16, line 128: Perhaps we can remove `numkeys` param from definition and calculate it from length of `keys` when necessary.

tillkruss: Yeah, good idea. Let's do that.

tillkruss (changes requested):

src/Command/Redis/ZDIFF.php
```diff
@@ -0,0 +1,38 @@
...
+ * This command is similar to ZDIFFSTORE, but instead of
+ * storing the resulting sorted set, it is returned to the client.
+ */
+class ZDIFF extends RedisCommand
+{
+    use Numkeys {
+        Numkeys::setArguments as setNumkeys;
+    }
+    use WithScores {
+        WithScores::setArguments as setWithScore;
```
tillkruss, line 22: Are these explicit overwrites needed?

vladvildanov: Yeah, both this traits contains overridden setArguments method so it can be used without any additional code changes in separate, but if we wan't to combine them we need to add explicit aliases so we have different aliases for different setArguments approach and call it within a main class setArgument. This way this traits becomes more flexible and reusable in my opinion

vladvildanov: And yeah, they're required, my IDE highlighted it

PR #830, opened 2022-11-30
Added response parser for WithScore trait

vladvildanov:
I notice that I missed response parsing in case if WITHSCORE modifier will be applied to command. According to other commands, it needs to be converted into key-value pairs.


vladvildanov:
@dvora-h @sazzad16

Issue #818, opened 2022-11-16
Add support for missing redis command BZMPOP


chayim:
predis is missing support for some of the sorted set functions, such as [bzmpop](https://redis.io/commands/bzmpop/). This command has been available since Redis 6.2.0

PR #833, opened 2022-12-05
Added support for BZMPOP command

vladvildanov:
#818 Added support to redis client for BZMPOP command

@dvora-h @sazzad16


chayim (approved):
Solid. Glad you used the MinMaxModifier for validation.

PR #858, opened 2022-12-08
Refactored ZUNIONSTORE, ZINTERSTORE, removed Numkeys trait, improved Withscores response parsing

vladvildanov:
Refactored following commands to make them easy to extend (ZUNIONSTORE, ZUNION, ZINTERSTORE, ZINTER is the same family of commands, ZUNIONSTORE implemented first and should be a base for other commands from family).

Removed Numkeys trait and moved functionality into Keys, because they're never exists in separate, whenever we pass keys arguments as array we anyway need to calculate numkeys count

Added check for empty array within Withscores trait, to prevent from exception if empty response was returned


tillkruss (approved):
LGTM 👍

Issue #836, opened 2022-12-07
Add support for missing redis command ZINTER


chayim:
predis is missing support for the sorted set function [ZINTER](https://redis.io/commands/ZINTER)

PR #859, opened 2022-12-12
Extended Sorted Set support by adding ZINTER command

vladvildanov:
#836

@dvora-h @sazzad16


src/ClientContextInterface.php
```diff
@@ -114,6 +114,7 @@
  * @method $this zdiffstore(string $destination, array $keys)
  * @method $this zincrby($key, $increment, $member)
  * @method $this zinterstore(string $destination, array $keys, int[] $weights = [], string $aggregate = 'sum')
+ * @method $this zinter(array $keys, int[] $weights = [], string $aggregate = 'sum', bool $withScores = false)
```
tillkruss, line 117: Should the keys and weights be a single array?

vladvildanov: The same as ZUNION. Keys should be array because of numkeys calculation and weights as array just easier to maintain

Issue #839, opened 2022-12-07
Add support for missing redis command ZUNION


chayim:
predis is missing support for the sorted set function [ZUNION](https://redis.io/commands/ZUNION)

PR #860, opened 2022-12-12
Extended Sorted Set support by adding ZUNION command

vladvildanov:
#839 

@dvora-h @sazzad16


src/ClientContextInterface.php
```diff
@@ -127,6 +127,7 @@
  * @method $this zrevrange($key, $start, $stop, array $options = null)
  * @method $this zrevrangebyscore($key, $max, $min, array $options = null)
  * @method $this zrevrank($key, $member)
+ * @method $this zunion(array $keys, int[] $weights = [], string $aggregate = 'sum', bool $withScores = false)
```
tillkruss, line 130: Should keys and weights be a single array?

vladvildanov: Not sure If I get you. If you mean argument types, then I think both of them should be arrays.

For keys it's definitely should be, since we need to resolve numkeys argument and the easiest way is to calculate array count and we already have a trait for it. For weights it's optional, but in my opinion it's always easy to deal with single element the same way we're dealing with collection, in this case we don't need to maintain different implementations for different argument types

Issue #841, opened 2022-12-07
Add support for missing redis command BZPOPMAX


chayim:
predis is missing support for the sorted set function [BZPOPMAX](https://redis.io/commands/BZPOPMAX)

PR #864, opened 2022-12-13
Extend Sorted Set support by implementing BZPOPMAX command

vladvildanov:
#841 

@dvora-h @sazzad16


src/Command/Redis/BZPOPMAX.php
```diff
@@ -0,0 +1,21 @@
...
+ * @link https://redis.io/commands/bzpopmax/
+ *
+ * BZPOPMAX is the blocking variant of the sorted set ZPOPMAX primitive.
+ *
+ * It is the blocking version because it blocks the connection when there are
+ * no members to pop from any of the given sorted sets.
+ * A member with the highest score is popped from first sorted set that is non-empty,
+ * with the given keys being checked in the order that they are given.
+ */
+class BZPOPMAX extends BZPOPMIN
```
sazzad16, line 15: This doesn't seem right. Either we can have a BZPOPBase which would be extended by both BZPOPMAX and BZPOPMIN or these two would be independent.

sazzad16 (changes requested):

src/Command/Redis/BZPOPMAX.php
vladvildanov: Good point

vladvildanov: @sazzad16 Fixed

sazzad16 (approved):

sazzad16 (approved):

Issue #845, opened 2022-12-07
Add support for missing redis command COPY


chayim:
predis currently lacks support for the function [COPY](https://redis.io/commands/COPY)

PR #866, opened 2022-12-13
Extended Key Space support by implementing COPY command

vladvildanov:
#845 

@dvora-h @sazzad16

### DIFF
--- a/tests/PHPUnit/PredisCommandTestCase.php
+++ b/tests/PHPUnit/PredisCommandTestCase.php
@@ -51,10 +51,10 @@ abstract class PredisCommandTestCase extends PredisTestCase
      * Returns a new client instance.
      *
      * @param bool $flushdb Flush selected database before returning the client
-     *
+     * @param array|null $parameters
      * @return Client
      */
-    public function getClient(bool $flushdb = true): Client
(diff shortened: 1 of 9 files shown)


tests/PHPUnit/PredisCommandTestCase.php
```diff
@@ -51,10 +51,10 @@ public function getCommand(): Command\CommandInterface
      * Returns a new client instance.
      *
      * @param bool $flushdb Flush selected database before returning the client
-     *
+     * @param array|null $parameters
      * @return Client
      */
-    public function getClient(bool $flushdb = true): Client
+    public function getClient(bool $flushdb = true, ?array $parameters = null): Client
```
tillkruss, line 57: I quite dislike this addition, can you call `createClient()` directly, since you're not using the `$flushdb` argument anyhow?

tillkruss (changes requested):

tests/PHPUnit/PredisCommandTestCase.php
vladvildanov: Makes sense, didn't notice that `createClient()` is `protected` not a `private`

Issue #848, opened 2022-12-07
Add support for missing redis command GEOSEARCH


chayim:
predis currently lacks support for the function [GEOSEARCH](https://redis.io/commands/GEOSEARCH)

PR #867, opened 2022-12-15
Extend Geospatial support by implementing GEOSEARCH command

vladvildanov:
#848


vladvildanov:
@tillkruss Hey. Till! I want you to check this PR and let me know what you're thinking about it

In this PR I introduce an idea to represent complex command arguments as objects to make it easy handle them and provide more useful interface for developers to use such commands.

By "complex command arguments" I mean a conditional arguments that can accept different values (For example FROM and BY arguments in [GEOSEARCH](https://redis.io/commands/geosearch/) command). To avoid adding all of them as optional arguments in PHP and avoid problems with processing them, I introduce an interfaces and objects that allows to simplify usage for clients and simplify argument processing for developers as well (In my opinion).

Let me know what you think and is it a good idea to use objects to represent command arguments

tillkruss:
@vladvildanov I'll take look in the new year. It's a rather big change for a patch release, this should probably be a minor release at least.

chayim:
@vladvildanov Why'd you move this back to draft?  Command-wise and implementation wise, this looks to properly cover the GEO* options and command set, thank you. Like @tillkruss said.. looks like we're getting closer to releasing with a bump.

vladvildanov:
@chayim Just because it's not fully covered with tests (f.e Argument objects). According to functionality itself, it's ready

PR #868, opened 2022-12-20
Added support for Redis JSON module commands, added JSON.SET and JSON.GET commands

vladvildanov:
New functionality implemented:
1. Changed CI - Added one more workflow to run tests over redis-stack instances. Redis JSON as many other modules available in redis-stack which is based on redis-server >= 6.2.0. So now all redis-server instances <= 6.2.0 runs within "Redis server tests" workflow and >= 6.2.0 runs over "Redis stack tests" workflow.
2. Added new configuration object (for now can be used as configuration layer), that keeps information about modules.
3. Added new layer for command resolving, which is responsible for resolving Redis command objects from virtual methods of client. More logic to command resolving required as new module support was introduced, so it's make sense to keep it at separate layer.
4. Added new directory structure for module commands. Since, new module support was introduced it's make sense to keep module commands in separate directories (f.e .../Redis/Json/...), so it have more structured format. New module command directories should have the same name as module (keeps in configuration) and command ID should starts with corresponding module prefix, same as in Redis (f.e Json - JSON, Search - FT etc.)
5. New decorator. Added new requiresRedisJsonVersion annotation, so we can mark tests to run only on instances with corresponding version of Redis JSON.
6. Added support for base commands JSON.GET and JSON.SET

@dvora-h @sazzad16

### DIFF
--- a/.github/workflows/redis-server-tests.yml
+++ b/.github/workflows/redis-server-tests.yml
@@ -1,4 +1,4 @@
-name: Tests
+name: Redis server tests
 
 on:
   - push
@@ -12,7 +12,7 @@ jobs:
 
     services:
       redis:
-        image: redis-stack-server:${{ matrix.redis-versions }}
+        image: redis:${{ matrix.redis-versions }}
         ports:
         - 6379:6379
         options: --health-cmd="redis-cli ping" --health-interval=10s --health-timeout=5s --health-retries=3
@@ -21,7 +21,7 @@ jobs:
       fail-fast: false
       matrix:
         php-versions: ['7.2', '7.3', '7.4', '8.0', '8.1', '8.2']
-        redis-versions: ['3', '4', '5', '6', '7']
+        redis-versions: ['3', '4', '5']
 
     steps:
     - name: Checkout
--- a/.github/workflows/redis-stack-tests.yml
+++ b/.github/workflows/redis-stack-tests.yml
@@ -7,12 +7,12 @@ on:
 jobs:
 
   predis:
-    name: PHP ${{ matrix.php-versions }} (Redis ${{ matrix.redis-versions }})
+    name: PHP ${{ matrix.php-versions }} (Redis stack ${{ matrix.redis-stack-versions }})
     runs-on: ubuntu-latest
 
     services:
       redis:
-        image: redis-stack-server:${{ matrix.redis-stack-versions }}
+        image: redis/redis-stack-server:${{ matrix.redis-stack-versions }}
         ports:
         - 6379:6379
         options: --health-cmd="redis-cli ping" --health-interval=10s --health-timeout=5s --health-retries=3
--- a/.github/workflows/redis-stack-tests.yml
+++ b/.github/workflows/redis-stack-tests.yml
@@ -1,0 +1,52 @@
+name: Redis stack tests
+
+on:
+  - push
+  - pull_request
+
+jobs:
+
+  predis:
+    name: PHP ${{ matrix.php-versions }} (Redis ${{ matrix.redis-versions }})
+    runs-on: ubuntu-latest
+
+    services:
+      redis:
+        image: redis-stack-server:${{ matrix.redis-stack-versions }}
+        ports:
+        - 6379:6379
+        options: --health-cmd="redis-cli ping" --health-interval=10s --health-timeout=5s --health-retries=3
+
+    strategy:
+      fail-fast: false
+      matrix:
+        php-versions: ['7.2', '7.3', '7.4', '8.0', '8.1', '8.2']
+        redis-stack-versions: ['6.2.6-v0', '7.0.6-RC2']
+
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v3
+
+    - name: Setup PHP with Composer and extensions
+      with:
+        php-version: ${{ matrix.php-versions }}
+      uses: shivammathur/setup-php@v2
+
+    - name: Get Composer cache directory
+      id: composercache
+      run: echo "::set-output name=dir::$(composer config cache-files-dir)"
+
+    - name: Cache Composer dependencies
+      uses: actions/cache@v2
+      with:
+        path: ${{ steps.composercache.outputs.dir }}
+        key: ${{ runner.os }}-composer-${{ hashFiles('**/composer.lock') }}
+        restore-keys: ${{ runner.os }}-composer-
+
+    - name: Install Composer dependencies
+      env:
+        PHP_VERSION: ${{ matrix.php-versions }}
+      run: composer install --no-progress --prefer-dist --optimize-autoloader $(if [ "$PHP_VERSION" == "8.0" ]; then echo "--ignore-platform-reqs"; fi;)
+
+    - name: Test with PHPUnit
+      run: vendor/bin/phpunit
--- a/src/ClientConfiguration.php
+++ b/src/ClientConfiguration.php
@@ -1,0 +1,28 @@
+<?php
+
+namespace Predis;
+
+class ClientConfiguration
+{
+    /**
+     * @var array{modules: array}|string[][]
+     */
+    private static $config = [
+        'modules' => [
+            ['name' => 'Json', 'commandPrefix' => 'JSON'],
+            ['name' => 'Graph', 'commandPrefix' => 'GRAPH'],
+            ['name' => 'Search', 'commandPrefix' => 'FT'],
+            ['name' => 'TimeSeries', 'commandPrefix' => 'TS'],
+        ]
+    ];
+
+    /**
+     * Returns available modules with configuration.
+     *
+     * @return array|string[][]
+     */
+    public static function getModules(): array
+    {
+        return self::$config['modules'];
+    }
+}
--- a/src/Command/RedisFactory.php
+++ b/src/Command/RedisFactory.php
@@ -52,7 +52,14 @@ class RedisFactory extends Factory
             return $this->commands[$commandID];
         }
 
-        return $this->commandResolver->resolve($commandID);
+        $commandClass = $this->commandResolver->resolve($commandID);
+
+        if (null === $commandClass) {
+            return null;
+        }
+
+        $this->commands[$commandID] = $commandClass;
+        return $commandClass;
     }
 
     /**
--- a/src/Command/RedisFactory.php
+++ b/src/Command/RedisFactory.php
@@ -11,6 +11,8 @@
 
 namespace Predis\Command;
 
+use Predis\Command\Resolver\CommandResolverInterface;
+
 /**
  * Command factory for mainline Redis servers.
  *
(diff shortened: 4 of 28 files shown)


.github/workflows/redis-stack-tests.yml
```diff
@@ -0,0 +1,52 @@
...
+        image: redis/redis-stack-server:${{ matrix.redis-stack-versions }}
+        ports:
+        - 6379:6379
+        options: --health-cmd="redis-cli ping" --health-interval=10s --health-timeout=5s --health-retries=3
+
+    strategy:
+      fail-fast: false
+      matrix:
+        php-versions: ['7.2', '7.3', '7.4', '8.0', '8.1', '8.2']
+        redis-stack-versions: ['6.2.6-v0', '7.0.6-RC2']
```
chayim, line 24: IMHO you shouldn't have to do this. 'latest' should be fine - especially since the redis version isn't he important change, but the underlying module versions.

redis-stack-server releases monthly, so I'd err on the side of latest here - like other clients.

.github/workflows/redis-stack-tests.yml
```diff
@@ -0,0 +1,52 @@
...
+      with:
+        php-version: ${{ matrix.php-versions }}
+      uses: shivammathur/setup-php@v2
+
+    - name: Get Composer cache directory
+      id: composercache
+      run: echo "::set-output name=dir::$(composer config cache-files-dir)"
+
+    - name: Cache Composer dependencies
+      uses: actions/cache@v2
```
chayim, line 40: Good time to move to cache@v3...

src/ClientConfiguration.php
```diff
@@ -0,0 +1,28 @@
...
+
+class ClientConfiguration
+{
+    /**
+     * @var array{modules: array}|string[][]
+     */
+    private static $config = [
+        'modules' => [
+            ['name' => 'Json', 'commandPrefix' => 'JSON'],
+            ['name' => 'Graph', 'commandPrefix' => 'GRAPH'],
```
chayim, line 13: Should these be here if this is json only? I realize we need them for the future.. but IMHO for now, they don't necessarily belong in this PR.

tests/Predis/Command/Redis/Json/JSONSET_Test.php
```diff
@@ -0,0 +1,176 @@
...
+            'override failed with XX subcommand' => [
+                'key',
+                '{"key1":"value1","key2":"value2"}',
+                '"value3"',
+                '$.key3',
+                'xx',
+                null,
+                '{"key1":"value1","key2":"value2"}'
+            ],
+            'append failed with NX subcommand' => [
```
chayim, line 165: IMHO we should do hierarchy (path based) tests as well.

tests/Predis/Command/Resolver/CommandResolverTest.php
```diff
@@ -0,0 +1,33 @@
...
+
+        $this->assertSame($expectedCommandClass, $resolver->resolve($commandID));
+    }
+
+    public function commandsProvider(): array
+    {
+        return [
+            'core command exists' => ['SET', SET::class],
+            'module not exist' => ['FOOBAR', null],
+            'module exists, module command not exists' => ['JSONFOO', null],
```
chayim, line 30: :heart:

src/Command/Traits/Json/NxXxSubcommand.php
```diff
@@ -0,0 +1,52 @@
...
+
+namespace Predis\Command\Traits\Json;
+
+use Predis\Command\Command;
+use UnexpectedValueException;
+
+/**
+ * @mixin Command
+ */
+trait NxXxSubcommand
```
sazzad16, line 11: NX/XX can be called option or argument or, maybe, parameter; but not sub-command.

sazzad16 (changes requested):

.github/workflows/redis-stack-tests.yml
vladvildanov: Okay, got it. I just followed by approach of current workflow that runs tests over each major version of Redis server

.github/workflows/redis-stack-tests.yml
vladvildanov: Good point

src/ClientConfiguration.php
vladvildanov: Yeah, I agree. Not necessary at least now

src/Command/Traits/Json/NxXxSubcommand.php
vladvildanov: According to Redis specification for JSON.SET:

_Optional subcommands NX and XX modify this behavior for both new RedisJSON data type keys as well as the JSON Object keys in them._

tests/Predis/Command/Redis/Json/JSONSET_Test.php
vladvildanov: Not sure If I get you

chayim: redisjson allows us to not just set things at the top level. We can do path based setting like JSON. SET $ <some doc> JSON.SET $..foo <some doc>

See the redis-py tests (test_json.py) as an example.

vladvildanov: @chayim There's a bunch of test cases with path based JSON setting:
https://github.com/predis/predis/pull/868/files/92de6634d9ae24d7a2f6d8d6c0d6c9fba02a2980#diff-c5a3fb39f6bdb4210ce15d464750651776475b2635b1ba1be15741ffecfcf91cR120

https://github.com/predis/predis/pull/868/files/92de6634d9ae24d7a2f6d8d6c0d6c9fba02a2980#diff-c5a3fb39f6bdb4210ce15d464750651776475b2635b1ba1be15741ffecfcf91cR129

https://github.com/predis/predis/pull/868/files/92de6634d9ae24d7a2f6d8d6c0d6c9fba02a2980#diff-c5a3fb39f6bdb4210ce15d464750651776475b2635b1ba1be15741ffecfcf91cR138

https://github.com/predis/predis/pull/868/files/92de6634d9ae24d7a2f6d8d6c0d6c9fba02a2980#diff-c5a3fb39f6bdb4210ce15d464750651776475b2635b1ba1be15741ffecfcf91cR147

chayim (approved):
Functionality met :D

src/Command/Traits/Json/NxXxSubcommand.php
sazzad16: The documentation must be wrong. I have opened https://github.com/RedisJSON/RedisJSON/pull/900. Please at least wait for that PR to be resolved.

@chayim

tillkruss:
Let's put this into a minor release in the future.

src/Command/Traits/Json/NxXxSubcommand.php
sazzad16, line 11: > ... Optional arguments NX and XX modify this behavior ...

sazzad16 (changes requested):

src/Command/Traits/Json/NxXxSubcommand.php
vladvildanov: @sazzad16 Fixed

chayim:
@tillkruss what do you think of this direction? I'm happy to review the other PRs and push through - but wanted your confirmation on this. 

I'm thinking post 2.1 (2.2?) we can start to roll all of these in. WDYT?

tillkruss:
> @tillkruss what do you think of this direction?

What do you mean by direction?

> I'm thinking post 2.1 (2.2?) we can start to roll all of these in. WDYT?

I rather keep 2.1 smaller, without these larger changes.

.github/workflows/redis-server-tests.yml
```diff
@@ -24,8 +24,6 @@ jobs:
           - 3
           - 4
           - 5
-          - 6
```
vladvildanov, line 27: @tillkruss We don't need this versions in this workflow, we're running tests through `latest` version of server within our `redis-stack` workflow, so we make sure that we have all modules available and we can run tests for module commands

.github/workflows/redis-server-tests.yml
```diff
@@ -1,4 +1,4 @@
-name: Tests
+name: Redis_server_tests
```
tillkruss, line 1: No need to change this.

.github/workflows/redis-stack-tests.yml
```diff
@@ -0,0 +1,52 @@
+name: Redis_stack_tests
```
tillkruss, line 1: This file isn't needed anymore.

README.md
```diff
@@ -491,7 +491,7 @@ The code for Predis is distributed under the terms of the MIT license (see [LICE
 [ico-version-stable]: https://img.shields.io/packagist/v/predis/predis.svg?style=flat-square
 [ico-version-dev]: https://img.shields.io/packagist/vpre/predis/predis.svg?style=flat-square
 [ico-downloads-monthly]: https://img.shields.io/packagist/dm/predis/predis.svg?style=flat-square
-[ico-build]: https://img.shields.io/github/actions/workflow/status/predis/predis/tests.yml?branch=main&style=flat-square
+[ico-build]: https://img.shields.io/github/actions/workflow/status/predis/predis/redis-server-tests.yml?branch=main&style=flat-square
```
tillkruss, line 494: Let's not rename the file, it messes with GHA.

tillkruss (changes requested):

.github/workflows/redis-stack-tests.yml
vladvildanov: So we're changing original workflow to run tests over `redis-stack` instead of `redis-server`?

src/Command/RedisFactory.php
```diff
@@ -25,15 +27,18 @@
 class RedisFactory extends Factory
 {
     /**
-     *
+     * @var CommandResolverInterface
      */
-    public function __construct()
+    private $commandResolver;
+
+    public function __construct(CommandResolverInterface $commandResolver)
```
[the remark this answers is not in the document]
vladvildanov: Logic of command resolving becomes more complex since we adding support for modules, so In my opinion to avoid violation of SRP it makes sense to separate logic of object creation (factory) and resolving object from command ID (resolver)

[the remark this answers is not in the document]
tillkruss: We can't break the interface in a point release, we can make this a `3.0` and release when it's feature complete with Redis Stack support?

.github/workflows/redis-stack-tests.yml
tillkruss: Keep `tests.yml` and call this `stack.yml`.

src/Command/RedisFactory.php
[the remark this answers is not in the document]
vladvildanov: Makes sense to me. @chayim What do you think?

.github/workflows/redis-stack-tests.yml
vladvildanov: `Redis stack` only available since `Redis` server version >= 6.x, so it means we can only run tests against v6.x, 7.x

PR #893, opened 2023-01-04
Extended RedisBloom support by implementing BF.CARD command

vladvildanov:
Added support for new BloomFIlter command BF.CARD

@dvora-h @sazzad16


chayim:
@vladvildanov the command is solid.
@tillkruss we have a bunch of these starting to stack up - that are starting to be based on each other.

I'm happy with the command support in each. Anything else you'd like to see here?

chayim:
no longer relevent - abandoning

PR #810, opened 2022-11-02
added annotations to suppress PHP 8.1 return type deprecation warning…

trickeyone:
… for `IteratorAggregate::toIterator()`
fixes #736

PR #1017, opened 2023-01-17
Fix Client template

franmomu:
https://github.com/predis/predis/pull/810 added generics to `Client`, that means that when using `Client` you should specify both parameters, getting this kind of message using static analysis tools:

```
Method App\My\Service::__construct() has parameter $redisClient with generic class Predis\Client but does not specify its types:
         TKey, TValue
```

Looking at what `Client::getIterator()` does:

https://github.com/predis/predis/blob/fbf270156f529a9920551c6f499c52a11ca4a1c3/src/Client.php#L523-L539

it returns an `\ArrayIterator` having as keys strings and as values instances of the class itself, so I think this class shouldn't be generic.

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -36,10 +36,7 @@ use Predis\Transaction\MultiExec as MultiExecTransaction;
  * abstractions are built. Internally it aggregates various other classes each
  * one with its own responsibility and scope.
  *
- * {@inheritdoc}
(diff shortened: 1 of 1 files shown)


tillkruss:
Are you running PHPStan? How do you feel about setting up basic level 1 or 2 testing for this repo?

PR #1021, opened 2023-01-18
Replace goto operators

szepeviktor:
@tillkruss I'm no developer.
This was _donated_ to me.
Would you give it a go? 🟢

Please **DO NOT** merge this. This PR is only a question.

### DIFF
--- a/src/Connection/Cluster/RedisCluster.php
+++ b/src/Connection/Cluster/RedisCluster.php
@@ -238,7 +238,7 @@ class RedisCluster implements ClusterInterface, IteratorAggregate, Countable
         $retryAfter = $this->retryInterval;
         $command = RawCommand::create('CLUSTER', 'SLOTS');
 
-        while ($retries < $this->retryLimit) {
+        while ($retries <= $this->retryLimit) {
             try {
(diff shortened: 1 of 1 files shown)


tillkruss:
It would never throw the exception this way, which is needed.

szepeviktor:
> It would never throw the exception this way, which is needed.

Is this a go or a no-go?

tillkruss:
No-go, since it doesn't throw exceptions anymore.

tillkruss:
I see your 2nd commit fixed the condition.

Why do you want to change the logic?

szepeviktor:
> Why do you want to change the logic?

Because goto is the worst part of PHP.
There bad parts but goto it the worst. No software should have goto hidden inside.

szepeviktor:
Fixing up to PHPStan level 2 is going on ...

tillkruss:
> Because goto is the worst part of PHP. There bad parts but goto is the worst. No software should have goto hidden inside.

All operators are beautiful, we don't hate on `goto` around here.

szepeviktor:
Emoootional overload!

PR #1024, opened 2023-01-18
Add CS job

franmomu:


.github/workflows/coding-standards.yml
```diff
@@ -0,0 +1,28 @@
+name: Coding Standards
```
chayim, line 1: ```name: Linters```

.github/workflows/coding-standards.yml
```diff
@@ -0,0 +1,28 @@
...
+
+on:
+  push:
+    branches:
+      - main
+      - v2.**
```
chayim, line 7:
```
- v[0-9].[0-9]
```

.github/workflows/coding-standards.yml
```diff
@@ -0,0 +1,28 @@
...
+  pull_request:
+
+jobs:
+
+  php-cs-fixer:
+    name: PHP-CS-fixer
```
chayim, line 13: ```name: linter```

chayim:
@franmomu Big fan of linting... thank you. WDYT of the small changes?

.github/workflows/coding-standards.yml
franmomu: I've used the same as in https://github.com/predis/predis/blob/113145f3065b8f34da2e09359f403e4db213b136/.github/workflows/tests.yml#L7

should I change it also there?

tillkruss: This is fine 👍

.github/workflows/coding-standards.yml
```diff
@@ -0,0 +1,28 @@
...
+
+    steps:
+
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Setup PHP with Composer and extensions
+        with:
+          php-version: 8.1
+          tools: php-cs-fixer
```
tillkruss, line 24: Love this without a dependency!

.php-cs-fixer.dist.php
```diff
@@ -13,6 +13,7 @@
 return (new PhpCsFixer\Config)
     ->setUsingCache(false)
     ->setRules([
+        '@PHP71Migration' => true,
```
tillkruss, line 16: I didn't know this rule, very nice.

tillkruss (approved):

PR #1026, opened 2023-01-18
Fix LICENSE file

szepeviktor:
GitHub has problems with it
`curl "https://api.github.com/repos/predis/predis" -H "Accept: application/vnd.github.drax-preview+json" -s | jq '.license'`


szepeviktor:
```
https://img.shields.io/github/license/szepeviktor/predis.svg?style=flat-square
```
fixed

PR #1112, opened 2023-02-08
Enable CI on v3 branch

szepeviktor:
Just that.
I wish you many ✔️✔️✔️

Simply add `[skip ci]` to commit message when it should not run.


szepeviktor:
Will there be no v3? 🤯

tillkruss:
Once we tag it, we can make all the `v2.x => v3.x` adjustments.

PR #1143, opened 2023-02-10
Added support for RediSearch, implemented FT.CREATE command

### LINKED ISSUE #1121: Add support for redis search command FT.CREATE
As documented at https://redis.io/commands/FT.CREATE

vladvildanov:
Closes #1121 

@shacharPash


phpunit.xml.dist
```diff
@@ -48,7 +48,7 @@
         <!-- Redis -->
         <const name="REDIS_SERVER_HOST" value="127.0.0.1" />
         <const name="REDIS_SERVER_PORT" value="6379" />
-        <const name="REDIS_SERVER_DBNUM" value="15" />
+        <const name="REDIS_SERVER_DBNUM" value="0" />
```
chayim, line 51: search only goes against index 0 - gerat.

src/Command/Argument/Search/CreateArguments.php
```diff
@@ -0,0 +1,248 @@
...
+
+    /**
+     * Specify data type for given index. To index JSON you must have the RedisJSON module to be installed.
+     *
+     * @param  string $modifier
+     * @return $this
+     */
+    public function on(string $modifier = 'hash'): self
+    {
+        if (strtoupper($modifier) === 'HASH' || strtoupper($modifier) === 'JSON') {
```
chayim, line 33: I think ```in_array``` is better here. One day search should support more types (streams?) and this would be limiting. WDYT?

src/Command/Argument/Search/CreateArguments.php
```diff
@@ -0,0 +1,248 @@
+<?php
+
+/*
+ * This file is part of the Predis package.
```
chayim, line 4: IMHO new code, stop adding the copyright...

src/Command/Argument/Search/CreateArguments.php
```diff
@@ -0,0 +1,248 @@
...
+        return $this;
+    }
+
+    /**
+     * Adds default language for documents within an index.
+     *
+     * @param  string $defaultLanguage
+     * @return $this
+     */
+    public function language(string $defaultLanguage): self
```
chayim, line 78: Can we have default assignments?

src/Command/Argument/Search/CreateArguments.php
vladvildanov: Yeah, totally agree. P.S Lucking Enums so much 😩

src/Command/Argument/Search/CreateArguments.php
vladvildanov: Since, one of latest @tillkruss PR, copyright becomes mandatory linter option, it's automatically added by IDE according to linter settings, so it will break CI from now

src/Command/Argument/Search/CreateArguments.php
vladvildanov: Yeah, we can. Good suggestion 👌

src/Command/Argument/Search/CreateArguments.php
chayim: @tillkruss WDYT?

tillkruss: Let's use a dedicated copyright for all files, not individual headers for various files.

PR #1145, opened 2023-02-13
Extended RediSearch support by implementing FT.SEARCH command

### LINKED ISSUE #1132: Add support for redis search command FT.SEARCH
As documented at https://redis.io/commands/FT.SEARCH

vladvildanov:
Closes #1132 

@shacharPash
@sazzad16


src/Command/Argument/Search/CreateArguments.php
```diff
@@ -0,0 +1,656 @@
...
+        'desc' => 'DESC',
+    ];
+
+    /**
+     * Specify data type for given index. To index JSON you must have the RedisJSON module to be installed.
+     *
+     * @param  string $modifier
+     * @return $this
+     */
+    public function on(string $modifier = 'hash'): self
```
shacharPash, line 47: the `$modifier` can be  `supportedDataTypesEnum` type instead of `string`?
if its possible I think it will be better.

src/Command/Argument/Search/CreateArguments.php
```diff
@@ -0,0 +1,656 @@
...
+        return $this;
+    }
+
+    /**
+     * Adds default language for documents within an index.
+     *
+     * @param  string $defaultLanguage
+     * @return $this
+     */
+    public function language(string $defaultLanguage = 'english'): self
```
shacharPash, line 95: If someone calls the function and puts the default value into it, it's better not to call the function because it only adds actions that don't change anything and extends arguments[].
If possible, I think it should be written in the description of the function what the default value of that field is, and the user will call the function only if he wants to change this value.

src/Command/Argument/Search/CreateArguments.php
```diff
@@ -0,0 +1,656 @@
...
+
+    /**
+     * Formats occurrences of matched text.
+     *
+     * @param  array  $fields
+     * @param  string $openTag
+     * @param  string $closeTag
+     * @return $this
+     */
+    public function highlight(array $fields = [], string $openTag = '', string $closeTag = ''): self
```
shacharPash, line 463: Are you intentionally not separating the arguments of FT.CREATE and FT.SEARCH into two separate classes?

src/Command/Argument/Search/CreateArguments.php
vladvildanov: That how it was before, @chayim requested this changes. @chayim Should we keep it?

src/Command/Argument/Search/CreateArguments.php
shacharPash: But according to what I see in the form it is now, you can insert highlight into the FT.CERATE command even though it does not have this parameter. Am I missing something?

vladvildanov: This PR is about FT.SEARCH, but it's based on FT.CREATE branch, that's why there's both changes here. This parameter for FT.SEARCH.

src/Command/Argument/Search/Schema.php
```diff
@@ -0,0 +1,191 @@
...
+     * @param  bool   $sortable
+     * @param  bool   $unf
+     * @param  bool   $noIndex
+     * @return $this
+     */
+    private function addField(
+        string $type,
+        string $fieldName,
+        string $alias = '',
+        bool $sortable = false,
```
leibale, line 139: 1. Not all fields are `SORTABLE`, only `NUMERIC`, `TAG`, `TEXT` and `GEO` are.
2. `$sortable = false, $unf = true` is invalid... I think you should use an `enum` (or any equivalent alternative):
```php
enum Sortable
{
    case False; // rename to `Off`/`False`?
    case True; // rename to `On`/`True`?
    case UNF;
}

private function addField(
    string $type,
    string $fieldName,
    string $alias = '',
    bool $sortable = Sortable.False,
    bool $noIndex = false
)
```

src/ClientInterface.php
```diff
@@ -89,6 +91,8 @@
  * @method int               decr(string $key)
  * @method int               decrby(string $key, int $decrement)
  * @method Status            failover(?To $to = null, bool $abort = false, int $timeout = -1)
+ * @method Status            ftcreate(string $index, CreateArguments $arguments, Schema $schema)
```
leibale, line 94: `CreateArguments $arguments` should be optional

src/ClientInterface.php
```diff
@@ -89,6 +91,8 @@
...
+ * @method array             ftsearch(string $index, string $query, CreateArguments $arguments)
```
leibale, line 95: `CreateArguments $arguments`?!

src/Command/Argument/Search/Schema.php
```diff
@@ -0,0 +1,191 @@
...
+     * @return $this
+     */
+    public function addVectorField(
+        string $fieldName,
+        string $alias = '',
+        bool $sortable = false,
+        bool $unf = false,
+        bool $noIndex = false
+    ): self {
+        return $this->addField('VECTOR', $fieldName, $alias, $sortable, $unf, $noIndex);
```
leibale, line 121: `VECTOR` does not support `SORTABLE` and have more options, see [here](https://redis.io/docs/stack/search/reference/vectors/)

src/Command/Argument/Search/CreateArguments.php
leibale: Why `modifier` is optional with a default value? If you want to use the default just don't call this function...

leibale, line 47: Why `modifier` is optional with a default value? If you want to use the default just don't call this function...

src/Command/Argument/Search/Schema.php
vladvildanov: @leibale 
1. Okay, thanks for letting me know 👌
2. Enums in PHP supported only from >= 8.1, but in Predis we should support >= 7.2

src/ClientInterface.php
vladvildanov: Oh, you're right 👌

src/ClientInterface.php
vladvildanov: It was `Arguments` first, but then @sazzad16 suggest to rename it to `CreateArguments`, for me it more feels like `SearchArgument`, this object is all about creating an array of arguments

src/Command/Argument/Search/CreateArguments.php
vladvildanov: It was like this before, then @chayim suggest to add it, but I agree we don't need default value if it's anyway will be applied

vladvildanov: Fixed

src/Command/Argument/Search/Schema.php
vladvildanov: Fixed

src/ClientInterface.php
vladvildanov: Fixed

src/Command/Argument/Search/Schema.php
vladvildanov: Fixed

src/ClientInterface.php
leibale: to me `CreateArguments` suggests it's the arguments for `FT.CREATE`... I would rename it `SearchArguments`

src/Command/Argument/Search/Schema.php
leibale: @vladvildanov you can use constants instead of an enum

src/ClientInterface.php
vladvildanov: @leibale Fixed

src/Command/Argument/Search/Schema.php
vladvildanov: @leibale Fixed

leibale (approved):

PR #1159, opened 2023-02-17
Extended RediSearch support by implementing FT.EXPLAIN command

### LINKED ISSUE #1128: Add support for redis search command FT.EXPLAIN
As documented at https://redis.io/commands/FT.EXPLAIN

vladvildanov:
Closes #1128 

@shacharPash


src/Command/Argument/Search/ExplainArguments.php
```diff
@@ -0,0 +1,17 @@
...
+ * (c) 2009-2020 Daniele Alessandri
+ * (c) 2021-2023 Till Krüss
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Predis\Command\Argument\Search;
+
+class ExplainArguments extends CommonArguments
```
shacharPash, line 15: why this is needed if its empty?

shacharPash (approved):

src/Command/Argument/Search/ExplainArguments.php
vladvildanov: @shacharPash We agreed to provide separate interface for command arguments if it's needed, well `CommonArguments` is an interface that contain arguments to be used across multiple commands, however this command can accept only arguments already specified within `CommonArguments`, so this object was created just to separate interfaces. In future if it could accept more specific arguments, it will be specified here

shacharPash (approved):

.editorconfig
```diff
@@ -22,3 +22,6 @@ max_line_length = unset
 
 [{src/ClientInterface.php,src/ClientContextInterface.php}]
 max_line_length = unset
+
+[tests/Predis/Command/Redis/Search/FTEXPLAIN_Test.php]
+indent_size = unset
```
tillkruss, line 27: Let's not add exceptions for test files.

tillkruss (changes requested):

.editorconfig
vladvildanov: @tillkruss Well, in this case CI breaks, to make sure that expected response exactly matches actual response there should be no indents in it

tillkruss: `eclint` is so inflexible 😓

tillkruss (approved):

.editorconfig
szepeviktor: We may have 1 additional exception per year!!!

szepeviktor:
Note: many of these commits are made with a non-GitHub email address

.editorconfig
tillkruss: @vladvildanov Can you try adding a 2nd `.editorconfig` to the `/tests` folder that contains the `FTEXPLAIN_Test.php` exception only?

tillkruss (changes requested):

.editorconfig
vladvildanov: @tillkruss Works!

tillkruss (approved):

Issue #1144, opened 2023-02-13
LOADING Redis is loading the dataset in memory


turbo124:
**Describe the bug**
When using redis-sentinel, if the slave network cannot maintain throughput to stay within the client-output-buffer limits, the slave will disconnect with the following

```
Predis\Response\ServerException
LOADING Redis is loading the dataset in memory
```

This issue has been in the library since 1.x
https://github.com/predis/predis/pull/475

The above PR provided a graceful fallback to the master in these scenarios where the slaves were temporarily unable to return read requests.  When evaluating 2.x of this library, we are still able to reproduce these scenarios resulting in exceptions rather than graceful fallback to master

**Expected behavior**
There are a couple of options for this scenario.

- Provide a configuration option to only serve requests to the master. Load balancing is not the primary goal of using a Redis-sentinel.
- Patch using a similar approach where in the catch, if the exception matches, we serve the request from the master

```php

    private function retryCommandOnFailure(CommandInterface $command, $method)
    {
        $retries = 0;

        while ($retries <= $this->retryLimit) {
            try {
                $response = $this->getConnectionByCommand($command)->$method($command);
                break;
            } catch(\Predis\Response\ServerException $exception){

                if($exception->getMessage() == 'LOADING Redis is loading the dataset in memory'){
                    $response = $this->getMaster()->$method($command);
                }

            } catch (CommunicationException $exception) {

                
                $this->wipeServerList();
                $exception->getConnection()->disconnect();

                if ($retries === $this->retryLimit) {
                    throw $exception;
                }

                usleep($this->retryWait * 1000);

                ++$retries;
            }
        }

        return $response;
    }

```

Would a solution like this be considered? Or is there another approach recommended so that we can get this issue resolved and merged into 2.x ?

**Versions (please complete the following information):**
- Predis: [e.g. 1.x 2.x]
- PHP [e.g. 8.1.14]
- Redis Server [e.g. 7.*]
- OS [e.g. Ubuntu 20.04]

tillkruss:
What do the Redis Sentinel specs say about failing commands during `LOADING`? I always treated it as the server being down.

Ping @chayim.

tillkruss:
I'd say in that case let's handle in better on our end 👍

@vladvildanov Any thoughts?

vladvildanov:
[...]

PR #1163, opened 2023-02-20
Optionally disable load balancing reads to slaves when using a replication topology.

turbo124:
Following on from #1144 here is a solution for users who do not wish to load balance reads to slaves. This is via the configuration of the replication strategy in the configuration file.

```php
'sentinel-cache' => array_merge(
    array_map(
        function ($a, $b) {
            return ['host' => $a, 'port' => $b];
        },
        explode(',', env('REDIS_HOST', 'localhost')),
        explode(',', env('REDIS_PORT', 26379))
    ),
    ['options' => [
        'replication' => function () {
                              $strategy = new Predis\Replication\ReplicationStrategy();
                              $strategy->disableLoadBalancing();
                              return new Predis\Connection\Replication\SentinelReplication($strategy);
                          },
        'service' =>  env('REDIS_SENTINEL_SERVICE', 'mymaster'),
        'sentinel_timeout' => 3.0,
        'parameters' => [
            'password' => env('REDIS_PASSWORD', null),
            'database' => env('REDIS_CACHE_DB', 1),
        ],
    ]]
),
```

This will ensure all read/writes are delivered to the master redis server.

### DIFF
--- a/src/Replication/ReplicationStrategy.php
+++ b/src/Replication/ReplicationStrategy.php
@@ -23,6 +23,7 @@ class ReplicationStrategy
     protected $disallowed;
     protected $readonly;
     protected $readonlySHA1;
+    protected $loadBalancing = true;
 
     public function __construct()
     {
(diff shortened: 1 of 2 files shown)


tillkruss:
@vladvildanov Any thoughts?

PR #1179, opened 2023-02-24
Deprecate webdis and phpiredis

tillkruss:
Will be replaced with [Relay](https://github.com/cachewerk/relay) in v3.0.

See #1084.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@
 - Fix `@template` in `Predis\Client` (#1017)
 - Fix support options array in `ZINTERSTORE` and `ZUNIONSTORE` (#1018)
 
+## Deprecated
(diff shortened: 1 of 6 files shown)


vladvildanov:
@tillkruss What are the purpose of this deprecation? This clients aren't useful and makes no sense to support?

tillkruss:
The phpiredis client hasn't been updated in years no and the author has stepped away from coding: https://github.com/nrk/phpiredis

PAXANDDOS:
Is there any difference w/ and w/o phpiredis right now (before v3)? And just curious, what is expected to change with the Relay?

tillkruss:
You gotta benchmark that yourself if it makes sense for you workloads. 

Relay will drastically reduce the read IO when used with Predis.

PR #1180, opened 2023-02-26
Fix a few issues (group relay-todo)

zaherg:
This is an attempt to fix a few bugs, currently the errors went down from 55 to 19 using Redis 7 and to 3 using Redis 6)

```
(Redis 7)
./vendor/bin/phpunit -c phpunit.relay.xml --group relay-todo

ERRORS!
Tests: 140, Assertions: 195, Errors: 19, Failures: 13.
```

```
(Redis 6)
./vendor/bin/phpunit -c phpunit.relay.xml --group relay-todo

Tests: 140, Assertions: 174, Errors: 3, Failures: 9, Skipped: 30.
```

@tillkruss most of what is left related to relay like `Wrong number of arguments for 'MOVE' command (argc: 2, arity: 2) (RELAY_ERR_OTHER; commands.c:7344)` so any feedback on how to deal with that would be great.

if you think the PR is ready, feel free to merge.

### DIFF
--- a/tests/PHPUnit/OneOfConstraint.php
+++ b/tests/PHPUnit/OneOfConstraint.php
@@ -33,6 +33,8 @@ class OneOfConstraint extends Constraint
     protected function matches($other): bool
     {
         if (is_array($other)) {
+            $other = is_array($other[0]) ? array_merge(...$other) : $other;
+
             return !empty(array_intersect($other, $this->array));
         }
 
--- a/tests/PHPUnit/OneOfConstraint.php
+++ b/tests/PHPUnit/OneOfConstraint.php
@@ -33,8 +33,6 @@ class OneOfConstraint extends Constraint
     protected function matches($other): bool
     {
         if (is_array($other)) {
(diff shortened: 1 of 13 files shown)


tests/PHPUnit/OneOfConstraint.php
```diff
@@ -33,6 +33,8 @@ public function __construct(array $array)
     protected function matches($other): bool
     {
         if (is_array($other)) {
+            $other = is_array($other[0]) ? array_merge(...$other) : $other;
```
tillkruss, line 36: I don't know what this does. Removing for now.

zaherg: Basically, relay is sending the data in the form of an array inside an array, this check if the first element of the array is array then it flatten the array into one dimensional array

tillkruss: I think the better solution was this: 5d729787f0f153feac03c503f91be67c831de72d

PR #1181, opened 2023-02-27
Added CODEOWNERS file

vladvildanov:


tillkruss (approved):

.github/CODEOWNERS
```diff
@@ -0,0 +1,5 @@
+# Default
+*    @tillkruss
+
+# Redis commands (Stack included)
+src/Command/Redis/*              @vladvildanov
```
chayim, line 5: We should use @predis/team here instead of you @vladvildanov

vladvildanov: @chayim Right, fixed!

PR #1184, opened 2023-02-28
Working on relay-tx group

zaherg:
This PR should be a start to work on `relay-tx` group

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -436,10 +436,11 @@ class Client implements ClientInterface, IteratorAggregate
     /**
      * Actual pipeline context initializer method.
      *
-     * @param array $options  Options for the context.
+     * @param array|null $options Options for the context.
      * @param mixed $callable Optional callable used to execute the context.
      *
      * @return Pipeline|array
+     * @throws \Relay\Exception
      */
     protected function createPipeline(array $options = null, $callable = null)
     {
@@ -453,7 +454,7 @@ class Client implements ClientInterface, IteratorAggregate
 
         if( $this->connection instanceof RelayConnection) {
             $class = 'Predis\Pipeline\Relay';
-            
+
             if (isset($options['fire-and-forget']) && $options['fire-and-forget']) {
                 throw new \Relay\Exception('Relay does not support fire-and-forget');
             }
--- a/src/Client.php
+++ b/src/Client.php
@@ -25,6 +25,7 @@ use Predis\Configuration\OptionsInterface;
 use Predis\Connection\ConnectionInterface;
 use Predis\Connection\Parameters;
 use Predis\Connection\ParametersInterface;
+use Predis\Connection\RelayConnection;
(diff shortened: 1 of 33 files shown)


src/Pipeline/Relay.php
```diff
@@ -0,0 +1,37 @@
...
+{
+    /**
+     * {@inheritdoc}
+     */
+    protected function executePipeline(ConnectionInterface $connection, SplQueue $commands)
+    {
+        $pipeline = $connection->getClient()->pipeline();
+
+        foreach ($commands as $command) {
+           $pipeline->rawCommand($command->getId(), ...$command->getArguments());
```
tillkruss, line 32: This needs the same exceptions like `RelayConnection` has, select, auth, type.

tests/Predis/Pipeline/PipelineTest.php
```diff
@@ -492,7 +493,13 @@ public function testIntegrationWithServerExceptionInCallableBlock(): void
             // NOOP
         }
 
-        $this->assertInstanceOf('Predis\Response\ServerException', $exception);
+        // TODO: check and change the name if needed
+        if($client->getConnection() instanceof RelayConnection) {
+            $this->assertInstanceOf('Relay\Exception', $exception);
```
tillkruss, line 498: The `executePipeline()` needs to catch and convert exceptions. This test must remain unchanged.

tests/Predis/Pipeline/PipelineTest.php
```diff
@@ -504,6 +511,10 @@ public function testIntegrationWithServerErrorInCallableBlock(): void
     {
         $client = $this->getClient([], ['exceptions' => false]);
 
+        if($client->getConnection() instanceof RelayConnection) {
+            $this->markTestSkipped('Relay will always throw a connection');
```
tillkruss, line 515: Use the `relay-incompatible` group, instead of `markTestSkipped()`.

src/Client.php
```diff
@@ -450,6 +451,10 @@ protected function createPipeline(array $options = null, $callable = null)
             $class = 'Predis\Pipeline\Pipeline';
         }
 
+        if( $this->connection instanceof RelayConnection) {
+            $class = 'Predis\Pipeline\Relay';
+        }
+
```
tillkruss, line 457: Relay does not support `fire-and-forget`, we need to throw an exception here.

tillkruss (changes requested):

src/Pipeline/Relay.php
```diff
@@ -0,0 +1,60 @@
...
+ */
+class Relay extends Pipeline
+{
+    /**
+     * These commands must be called on the
+     * client, not using `rawCommand()`.
+     *
+     * @var string[]
+     */
+    private $atypicalCommands = [
```
tillkruss, line 31: Since we already have the connection, let's not define these twice. We could turn it into a public constant on the connection.

PR #1187, opened 2023-03-02
Low-level implementation for FT.CREATE command

vladvildanov:
@shacharPash


src/ClientInterface.php
```diff
@@ -108,7 +108,7 @@
  * @method Status            ftaliasdel(string $alias)
  * @method Status            ftaliasupdate(string $alias, string $index)
  * @method Status            ftalter(string $index, Schema $schema, ?AlterArguments $arguments = null)
- * @method Status            ftcreate(string $index, Schema $schema, ?CreateArguments $arguments = null)
+ * @method Status            ftcreate(string $index, FieldInterface[] $schema, string $on = '', array $prefixes = [], string $filter = '', string $language = '', string $languageField = '', float $score = 0, string $scoreField = '', string $payloadField = '', bool $maxTextFields = false, int $temporary = 0, bool $noOffsets = false, bool $noHl = false, bool $noFields = false, bool $noFreqs = false, array $stopWords = [], bool $skipInitialScan = false)
```
chayim, line 111: FieldInterface - why does this come before on?
The command order should follow FT.CREATE, even though most of the (post index) order is moveable.

chayim, line 111: The [default](https://redis.io/commands/ft.create/) score should be 1.0 not 0

chayim, line 111: Maybe temporary should become temporarySeconds to make it more user friendly - even though the arg name is _temporary_

chayim: Also I think we should check the two hash field types - even picking a default (JSON?)

chayim (changes requested):
WDYT?

src/ClientInterface.php
vladvildanov: In PHP pre-defined arguments cannot goes before required one, so since `$schema` is required it should goes before any pre-defined.

It checks field types within a builder, but it makes no sense to give it default value, because in this case we will always have `ON HASH` in argument set which makes no sense because it will be HASH by default anyway

vladvildanov: It have 2 perspectives:

From Redis server perspective if `default_score` parameter isn't set it will have default value `1.0`. 
From client perspective, if you want to keep the default value you should pass nothing as argument in this case `0`, which means that this value won't be included in argument set so Redis server will use `1.0` as default.

Isn't it?

vladvildanov: Okay 👍

vladvildanov: @chayim https://github.com/vladvildanov/predis/blob/vv-ftcreate-low-level/src/Command/Argument/Search/CreateArguments.php#L35

vladvildanov: Fixed

chayim: No, I disagree. I think we should have the client, echo the default value of the server (i.e use 1.0). Otherwise, we end up with possible confusion. IMHO the experience is better this way  (aligned with Redis), especially when a user reads the command docs.

vladvildanov: Okay 👍

vladvildanov: @chayim Fixed

tillkruss (approved):

chayim (approved):

PR #1192, opened 2023-03-07
Extended core support by changing arguments interface for SET command

### LINKED ISSUE #992: Support new Redis 6.2 arguments for SET
Add support for the enw arguments in Redis 6.2 for SET as specified at https://redis.io/commands/SET.

Detail of update: Allowed the `NX` and `GET` options to be used together.

This is part of #820.
Note: Due to the autogenerated nature of this, some of this may be already implemented.

vladvildanov:
Closes #992

### DIFF
--- a/src/ClientContextInterface.php
+++ b/src/ClientContextInterface.php
@@ -70,7 +70,7 @@ use Predis\Command\Redis\Container\FunctionContainer;
  * @method $this mset(array $dictionary)
  * @method $this msetnx(array $dictionary)
  * @method $this psetex($key, $milliseconds, $value)
- * @method $this set($key, $value, $expireResolution = null, $expireTTL = null, $flag = null)
+ * @method $this set($key, $value, bool $getOld = false, $expireResolution = null, $expireTTL = null, $flag = null)
  * @method $this setbit($key, $offset, $value)
  * @method $this setex($key, $seconds, $value)
  * @method $this setnx($key, $value)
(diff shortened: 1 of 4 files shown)


tillkruss:
This would be a breaking change and should go into v3.x, eh?

tillkruss:
> @tillkruss Looks like, there's a lot of command interface changes since Redis 6.2, so I assume it should be 3.x for all of them

Yes, we can't merge breaking changes into v2.

chayim:
Folks, I don't think we're having the same conversation here. If an interface is extended, meaning existing user usage doesn't change - I'd argue we can continue. If we **break**  an interface, then yes.

@tillkruss This is similar to many other projects so WDYT? Again the proposal I'm making is that when extending the interface - meaning no user facing change or behavioural change unless actively used, we continue within this code path.

tillkruss:
> Folks, I don't think we're having the same conversation here. If an interface is extended, meaning existing user usage doesn't change - I'd argue we can continue. If we **break** an interface, then yes.

Correct. I do see several changes to existing test cases. If we can leave the current ones in place and add new ones that prove the old functionality works, it's not a BC.

src/ClientContextInterface.php
```diff
@@ -70,7 +70,7 @@
  * @method $this mset(array $dictionary)
  * @method $this msetnx(array $dictionary)
  * @method $this psetex($key, $milliseconds, $value)
- * @method $this set($key, $value, $expireResolution = null, $expireTTL = null, $flag = null)
+ * @method $this set($key, $value, bool $getOld = false, $expireResolution = null, $expireTTL = null, $flag = null)
```
chayim, line 73: given the location. this looks breaking.
how about we punt this @vladvildanov

PR #1226, opened 2023-03-27
Extended RediSearch support by implementing TS.CREATERULE command

### LINKED ISSUE #1203: Implement TimeSeries command TS.CREATERULE
As documented at https://redis.io/commands/TS.CREATERULE

vladvildanov:
Closes #1203 

@shacharPash


src/Command/Redis/TimeSeries/TSCREATERULE.php
```diff
@@ -0,0 +1,40 @@
...
+{
+    public function getId()
+    {
+        return 'TS.CREATERULE';
+    }
+
+    public function setArguments(array $arguments)
+    {
+        [$sourceKey, $destKey, $aggregator, $bucketDuration] = $arguments;
+        $processedArguments = [$sourceKey, $destKey, 'AGGREGATION', $aggregator, $bucketDuration];
```
chayim, line 32: Maybe we should add a convenience wrapper on aggregations - so that a user can get the list of potential aggregations, and also validate against them...

tillkruss (approved):

src/Command/Redis/TimeSeries/TSCREATERULE.php
chayim: I think so. Other clients have added (to json) the use case where we enable setting from a file (so json.set_file) and a directory full of files.

IMHO the clients should be low level.. but in some cases (we can definitely talk about which) helper functions become useful. This is a think in go-redis already .. basically all over the place.

chayim: If you feel that's out of scope of this PR, then all good - it can be in other issues, clearly.

PR #1228, opened 2023-03-27
Extended TimeSeries support by implementing TS.INFO command

### LINKED ISSUE #1209: Implement TimeSeries command TS.INFO
As documented at https://redis.io/commands/TS.INFO

vladvildanov:
Closes #1209 

@shacharPash


examples/Commands/TimeSeries/ts_info.php
```diff
@@ -0,0 +1,32 @@
...
+
+require __DIR__ . '/../../shared.php';
+
+// Example of TS.CREATE command usage:
+
+// 1. Create time series
+$client = new Client();
+
+$arguments = (new CreateArguments())
+    ->retention(60000)
```
chayim, line 25: IMHO we called it in python retention_msecs as a variable. The reason - making it clear that its in milliseconds. WDYT?

vladvildanov: @chayim Sure, let's make it more consistent

tillkruss (approved):

PR #1235, opened 2023-03-30
Added stack commands to cluster and prefix available

vladvildanov:


tillkruss (changes requested):
Tests are failing.

vladvildanov:
@tillkruss Yup, I'm fixing it 👌

tillkruss:
@vladvildanov What does this exactly do and is it needed with #1253?

vladvildanov:
@tillkruss Yeah, this one is needed since we need to specify each command within `ClusterStrategy` to be able to use it in cluster mode, as well as we need to specify each command within `KeyPrefixProcessor` to be able to use it in case if `prefix` will be provided in configuration (but the last one won't need if #1262 will be merged)

tillkruss:
Alright, #1262 is merged.

radu-neacsu:
Up!

PR #1236, opened 2023-03-31
Added support for tests running against redis cluster

vladvildanov:
Tests marked with `@group cluster` annotation will be ran against redis cluster


phpunit.xml.dist
```diff
@@ -53,5 +53,8 @@
         <!-- Webdis -->
         <const name="WEBDIS_SERVER_HOST" value="127.0.0.1" />
         <const name="WEBDIS_SERVER_PORT" value="7379" />
+
+        <!-- Redis Cluster -->
+        <const name="REDIS_CLUSTER_ENDPOINTS" value="127.0.0.1:6372,127.0.0.1:6373,127.0.0.1:6374" />
```
chayim, line 58: Maybe we should add to the comments that these are the master nodes?

chayim (changes requested):

phpunit.xml.dist
vladvildanov: @chayim Absolutely 👍

tillkruss (changes requested):
Let's get this set up?

vladvildanov:
@tillkruss What exactly?

tillkruss:
I reckon duplicate the `test.yml` and run it against cluster.

.github/workflows/tests.yml
```diff
@@ -41,6 +41,16 @@ jobs:
         ports:
           - 6379:6379
 
+      redis-cluster:
+        image: redisfab/redis-py-cluster:6.2.6
```
tillkruss, line 45: This seems to ridgit. We should support 4, 5, 6 and 7 no?

chayim: Only 5,6,7 are supported. But cluster support in 6.x covers most cases - if anything we should use the redisenv clusters @vladvildanov

vladvildanov: @tillkruss @chayim Let's stick to other client practice, are you testing agains all cluster version or latest one?

chayim: Against that spreadsheet; 7.x is fine for cluster

.github/workflows/linters.yml
```diff
@@ -133,7 +133,7 @@ jobs:
 
       - name: Check exported files
         run: |
-          EXPECTED="LICENSE,README.md,autoload.php,composer.json"
+          EXPECTED="LICENSE,README.md,autoload.php,composer.json,docker/,docker/unstable_cluster/,docker/unstable_cluster/Dockerfile,docker/unstable_cluster/create_cluster.sh,docker/unstable_cluster/docker-compose.yml,docker/unstable_cluster/redis.conf"
```
chayim, line 136: can this be asterisked - ie docker/* or docker/**? Does that work here?

chayim (approved):

.github/workflows/linters.yml
vladvildanov: I think we can add docker folder as exclusion, would be even better

chayim (dismissed):

.github/workflows/tests-cluster.yml
```diff
@@ -0,0 +1,58 @@
...
+  pull_request:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+
+  predis:
+    name: PHP ${{ matrix.php }} Redis cluster:unstable
```
tillkruss, line 17:
```suggestion
    name: PHP ${{ matrix.php }} (Redis latest)
```

tillkruss (dismissed):

tillkruss (commented):
Should we merge `tests-cluster.yml` into `tests.yml`?

vladvildanov:
@tillkruss Yeah, would be better

src/Cluster/ClusterStrategy.php
```diff
@@ -53,6 +53,7 @@ protected function getDefaultCommands()
             'SORT' => [$this, 'getKeyFromSortCommand'],
             'DUMP' => $getKeyFromFirstArgument,
             'RESTORE' => $getKeyFromFirstArgument,
+            'FLUSHDB' => [$this, 'getFakeKey'],
```
tillkruss, line 56: Can you explain this hack?

vladvildanov: I need this hack because of the bad design of cluster connection.

Before sending command to the cluster, we need to determine the exact node that should accept this command. So, we're using the same hash algorithm that Redis does on the server side (CRC16), but we're pre-calculating it on the client-side, so we definitely knew the correct node before we're sending command. Hash calculation based on the key when we're running specific command (`SET key value`).

The issue is that there's a lot of commands that can be executed at any node and doesn't operate against specific key (like `FLUSHDB`), so we don't care on which node it will be executed. But the interface always expects key to calculate a hash and define correct node of cluster. So without changing the existing interface the hack is to provide literally any value so the hash can be calculated and we can run command, since we don't care about the node where it will be executed.

So, we need to change an interface or temporary use this "fake" value for the command that doesn't operate against storage keys

tillkruss: 😭 As long as Predis still complains that `FLUSHDB` isn't supported in cluster mode we're good 👍

tillkruss (approved):

src/Cluster/ClusterStrategy.php
vladvildanov: Here's a PR that introduce Clusterable interface and represents the command that can be executed against cluster. IMHO, it should be a part of command implementation and we don't need to touch any other code whenever new command appears.

For now this PR only touch a part of commands, more like a POC. Let me know what you think about this approach

https://github.com/predis/predis/pull/1318

chayim (approved):

PR #1258, opened 2023-04-04
Feature/fcall prefix processor and tests

mehrud:
Since adding FCALL [functions] to Redis 7.0, the related key prefix processor method is missed in the package, it's been added, and related test also has been added for this functionality

### DIFF
--- a/src/Command/Processor/KeyPrefixProcessor.php
+++ b/src/Command/Processor/KeyPrefixProcessor.php
@@ -39,6 +39,7 @@ class KeyPrefixProcessor implements ProcessorInterface
         $prefixSkipLast = static::class . '::skipLast';
         $prefixSort = static::class . '::sort';
         $prefixEvalKeys = static::class . '::evalKeys';
+        $prefixFCallKeys = static::class . '::fcallKeys';
         $prefixZsetStore = static::class . '::zsetStore';
         $prefixMigrate = static::class . '::migrate';
         $prefixGeoradius = static::class . '::georadius';
@@ -186,6 +187,8 @@ class KeyPrefixProcessor implements ProcessorInterface
             'XLEN' => $prefixFirst,
             'XACK' => $prefixFirst,
             'XTRIM' => $prefixFirst,
+            /* ---------------- Redis 7.0 ---------------- */
+            'FCALL' => $prefixFCallKeys,
         ];
     }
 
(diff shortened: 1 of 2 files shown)


mehrud:
@tillkruss hi, the `Coding Standards` checks seems not to be related to my changes. Could you please check

tillkruss (approved):

mehrud:
@tillkruss may I have any updates on this PR?

vladvildanov (approved):

src/Command/Processor/KeyPrefixProcessor.php
```diff
@@ -406,6 +409,23 @@ public static function evalKeys(CommandInterface $command, $prefix)
         }
     }
 
+    /**
+     * Applies the specified prefix to the keys of an FCall-based command.
+     *
+     * @param CommandInterface $command Command instance.
+     * @param string           $prefix  Prefix string.
+     */
+    public static function fCallKeys(CommandInterface $command, $prefix)
```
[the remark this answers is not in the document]
vladvildanov: @mehrud Need to add a check for keys argument. In `FCALL` interface only `$function` argument is required, so if no keys or arguments will be provided it will throw `Undefined idnex: $argument[1]`

vladvildanov (changes requested):

tillkruss:
@vladvildanov: Do you prefer #1258 over this PR?

src/Command/Processor/KeyPrefixProcessor.php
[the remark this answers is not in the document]
mehrud: @vladvildanov It's OK, since in FCALL command, like in EVAL, function name is always the first argument, and key numbers is the second one, so if there's no argument provided by the user, the `$command->getArguments()` will result in ['function_name', 0]

vladvildanov:
@tillkruss Do you mean #1262 ??

src/Command/Processor/KeyPrefixProcessor.php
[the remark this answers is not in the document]
vladvildanov: @mehrud Right, I missed that this one executes after `setArguments()`

vladvildanov (approved):

vladvildanov (approved):

tillkruss:
@vladvildanov: In the v2.x branch when a new prefix is added, we also need to add it to `ClusterStrategy.php`, correct?

vladvildanov:
@tillkruss No, `ClusterStrategy` is about availability of specific command against clustered Redis server, if you wan't that command would be available in cluster mode, you need to specify it within `ClusterStrategy`. Implementation quite same as prefix, but it's a totally different functionality

src/Command/Processor/KeyPrefixProcessor.php
tillkruss, line 418: This code is identical to `evalKeys()`. Can you use that function, or at least call it from `fCallKeys()`?

tillkruss (changes requested):

src/Command/Processor/KeyPrefixProcessor.php
mehrud: Yes I know, since it's name is `evalKeys`, I just thought to add a new function name, if you prefer to use `evalKeys` for both `EVAL` and `FCALL`, then I will change it.

mehrud: Yes I know, since it's name is `evalKeys`, I just thought to add a new function name, if you prefer to use `evalKeys` for both `EVAL` and `FCALL`, then I will change it.

tillkruss: Yes, either reuse the function, or call it from within it, instead of duplicate code.

PR #1262, opened 2023-04-10
Added support for PrefixableCommandInterface

vladvildanov:
This changes brings more structure and decoupling between commands and `KeyPrefixProcessor`. It's makes easy to maintain commands code and introduce new commands, so it's enough for command to implement `PrefixableCommandInterface` to make it passes through `KeyPrefixProcessor`


tillkruss:
How is this related to #1235?

vladvildanov:
@tillkruss With this changes, we don't need anymore to specify commands within `KeyPrefixProcessor` and setup the way keys should be prefixed, after this will be merged command itself would be responsible for applying prefix provided by processor to it's own keys.

PR #1265, opened 2023-04-19
Add missing `@return` annotations

nicolas-grekas:
Declaring `#[ReturnTypeWillChange]` is not precise enough, that's why they should always have their corresponding `@return`.

This change is required to enable using Predis v2 with Symfony, see https://github.com/symfony/symfony/pull/49801

### DIFF
--- a/src/Command/Processor/ProcessorChain.php
+++ b/src/Command/Processor/ProcessorChain.php
@@ -93,7 +93,9 @@ class ProcessorChain implements ArrayAccess, ProcessorInterface
     }
 
     /**
-     * {@inheritdoc}
+     * @param int $index
+     *
+     * @return bool
      */
     #[ReturnTypeWillChange]
     public function offsetExists($index)
@@ -102,7 +104,9 @@ class ProcessorChain implements ArrayAccess, ProcessorInterface
     }
 
     /**
-     * {@inheritdoc}
+     * @param int $index
+     *
+     * @return ProcessorInterface
      */
     #[ReturnTypeWillChange]
     public function offsetGet($index)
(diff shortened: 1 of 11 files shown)


src/Command/Processor/ProcessorChain.php
```diff
@@ -93,7 +93,7 @@ public function count()
     }
 
     /**
-     * {@inheritdoc}
+     * @return bool
```
tillkruss, line 96: These new blocks are missing `@param`.

src/Session/Handler.php
```diff
@@ -53,7 +53,7 @@ public function register()
     }
 
     /**
-     * {@inheritdoc}
+     * @return bool
```
tillkruss, line 56: These new blocks are missing `@param`.

tillkruss (changes requested):
Thanks for the PR, makses sense. Just some minor changes needed.

src/Command/Processor/ProcessorChain.php
nicolas-grekas: well, ArrayAccess only allows mixed for them, how do you want to document this? `@param mixed`?

tillkruss: Yes, we can use mixed for everything for now.

src/Session/Handler.php
nicolas-grekas: updated

src/Command/Processor/ProcessorChain.php
nicolas-grekas: Actually I went with `@param int` as done in Cluster/SlotMap

PR #1266, opened 2023-04-26
When being moved to a different connection, reuse the parameters (apa…

### LINKED ISSUE #1259: Connection scheme (rediss) seems to be ignored when retrieving (new) clients from the IteratorAggregate
**Describe the bug**
Predis/Client configured in cluster mode does not behave correctly when using the iterator to get individual node connections. The scheme used when configuring the cluster wide client, seems to be forgotten when a node connection is requested. Eg: When rediss (aka TLS over TCP) is requested a node connection will fall back to just using TCP.


Reason for using the iterator: KEYS/SCAN operator seems to be undefined when using the cluster mode, thus we assume it would be possible to iterate over all the nodes, and issue a KEYS/SCAN operation on them.

**Background**
MemoryDB for redis seems to be an Amazon specific, but Redis protocol compliant implementation of Redis. The setup we have is a single shard, multiple replica Redis cluster.

**Expected behavior**
A. Be able to successfully issue a KEYS/SCAN command when running in (p)redis-cluster mode
or
B. When creating new connections using the Client iterator, use the scheme provided in the original config.

- Predis: 2.1.2
- PHP: 7.3
- Redis: Amazon MemoryDB for Redis
- OS: Ubuntu Bionic (docker image "ubuntu@sha256:c1d0baf2425ecef88a2f0c3543ec43690dc16cc80d3c4e593bb95e4f45390e45")

**Code sample**

Dependency injection:
```
class: Predis\Client
arguments:
    $parameters:
    - { scheme: '%redis_shared.scheme%', host: '%redis_shared.host%', port: '%redis_shared.port%', username: '%redis_shared.username%', password: '%redis_shared.password%' }
    $options: { prefix: '%redis.database.geo_prefix%', cluster: 'redis' }
```
where `redis_shared.scheme` = "rediss" and the `redis_shared.host` is of following format: `clustercfg.<project-name>-eu-central-1.htpkhd.memorydb.eu-central-1.amazonaws.com`. More specifically, this DNS record resolves to the 3 replica IP's.

```
$contexts = [];
$pattern = '<pattern>.*';
foreach ($this->redisClient as $nodeClient)
{
     foreach (new Iterator\Keyspace($nodeClient, $pattern) as $key)
     {
       list(/*prefix*/, $context) = explode('.', $key, 2);
       $contexts[] = $context;
      }
}
return $contexts;
```
`$this->redisClient` is the cluster configured Predis/Client, `$nodeClient` is, what I assume a client that is able to issue commands to a single instance

Result stacktrace:
```
In AbstractConnection.php line 131:                                                                                                                                                                                                            │
│                                                                                                                                                                                                                                                │
│   [Predis\Connection\ConnectionException]                                                                                                                                                                                                      │
│   Error while reading line from the server. [tcp://<node-uri>-aws-eu-central-1.htpkhd.memorydb.eu-central-1.amazonaws.com:6379]                                     
```

**Additional context**
Other actions such as SET or append to list are working when using the cluster client.

MrKickkiller:
…rt from host & port) from a random previously established connection.

Closes #1259

### DIFF
--- a/src/Connection/Cluster/RedisCluster.php
+++ b/src/Connection/Cluster/RedisCluster.php
@@ -25,6 +25,8 @@ use Predis\Command\RawCommand;
 use Predis\Connection\ConnectionException;
 use Predis\Connection\FactoryInterface;
 use Predis\Connection\NodeConnectionInterface;
+use Predis\Connection\Parameters;
+use Predis\Connection\ParametersInterface;
 use Predis\NotSupportedException;
 use Predis\Response\Error as ErrorResponse;
 use Predis\Response\ErrorInterface as ErrorResponseInterface;
@@ -333,14 +335,21 @@ class RedisCluster implements ClusterInterface, IteratorAggregate, Countable
      *
      * @return NodeConnectionInterface
      */
-    protected function createConnection($connectionID)
+    protected function createConnection($connectionID, ParametersInterface $src = null)
     {
         $separator = strrpos($connectionID, ':');
+        $host = substr($connectionID, 0, $separator);
+        $port = substr($connectionID, $separator + 1);
 
-        return $this->connections->create([
-            'host' => substr($connectionID, 0, $separator),
-            'port' => substr($connectionID, $separator + 1),
-        ]);
+        if ($src != null) {
+            $params = Parameters::create($src->toArray());
+        } else {
+            $params = new Parameters();
+        }
+        $params->host = $host;
+        $params->port = $port;
+
+        return $this->connections->create($params);
     }
 
     /**
@@ -384,7 +393,8 @@ class RedisCluster implements ClusterInterface, IteratorAggregate, Countable
         $connectionID = $this->guessNode($slot);
 
         if (!$connection = $this->getConnectionById($connectionID)) {
-            $connection = $this->createConnection($connectionID);
+            $randomConnection = $this->getRandomConnection();
+            $connection = $this->createConnection($connectionID,$randomConnection != null ? $randomConnection->getParameters(): null);
             $this->pool[$connectionID] = $connection;
         }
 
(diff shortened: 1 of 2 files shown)


vladvildanov:
@tillkruss @MrKickkiller This PR looks like a workaround. You shouldn't be able to communicate with "undefined" or "hidden" nodes, if you need to then it's a problem of Redis infrastructure. I would recommend to close this PR and start investigation on MemoryDB Redis infrastructure. I left a comment within #1259 that describes it better.

You should never have MOVED or ASK response from Redis server if you communicate via proxy, because it's a proxy responsibility to redirect it to the correct node, not a library.

vladvildanov:
@MrKickkiller UPD: This PR could make sense if other client also have a "hack" for MemoryDB connection, let me ensure.

Also, it's a breaking change, isn't it? @tillkruss

vladvildanov (changes requested):

vladvildanov:
@tillkruss @MrKickkiller I'm okay to approve, but it's looks like a breaking change?

tillkruss:
It is, so I'd would need to go into v3.0 which does not have a release date. Maybe towards the end of the year.

src/Connection/Cluster/RedisCluster.php
```diff
@@ -329,18 +331,29 @@ protected function guessNode($slot)
...
      * Creates a new connection instance from the given connection ID.
      *
-     * @param string $connectionID Identifier for the connection.
+     * @param string                   $connectionID Identifier for the connection.
+     * @param ParametersInterface|null $src          Previously established connection to base the connection setup off.
      *
      * @return NodeConnectionInterface
      */
-    protected function createConnection($connectionID)
+    protected function createConnection(string $connectionID, ParametersInterface $src = null): NodeConnectionInterface
```
tillkruss, line 339: Please rebase your PR to `main`. We cannot accept breaking changes in `v2.x`.

tillkruss (changes requested):

PR #1267, opened 2023-05-03
Added RESP3 protocol option, establish RESP3 connection

vladvildanov:
- Added new default option `protocol` with default value `2` (which corresponds to RESP2)
- Added possibility to establish different protocol version connection if given version `>2`


tillkruss:
Should we develop RESP3 in `main` after releasing v2.2 with Redis Stack and merging making sure the branches are in sync?

vladvildanov:
@tillkruss I prefer to develop in in separate branch as it's a big feature, but this branch definitely forked from `main`

PR #1269, opened 2023-05-03
Added timeout to prevent search before create

vladvildanov:
Fixes bug with randomly failing test, because search performed on index that wasn't created yet


tillkruss (approved):
love it

PR #1285, opened 2023-05-04
Added support for few commands in cluster strategy

vladvildanov:
Extended `ClusterStrategy` with commands, added support for commands with no `keys` arguments to run against cluster


tillkruss:
Can this be closed?

PR #1300, opened 2023-05-19
RESP3 commands integration tests and response handling

vladvildanov:


src/Command/Redis/HSCAN.php
```diff
@@ -84,6 +84,15 @@ public function parseResponse($data)
         return $data;
     }
 
+    /**
+     * @param $data
+     * @return array|mixed|string|null
+     */
+    public function parseResp3Response($data)
```
chayim, line 91: IMHO these should be private

src/Protocol/Parser/Strategy/Resp3Strategy.php
```diff
@@ -95,6 +100,7 @@ protected function parseVerbatimString(string $string): array
         return [
             'type' => 'verbatimString',
```
chayim, line 101: Pull the return type strings to a const that can be refered to

src/Connection/StreamConnection.php
```diff
@@ -313,11 +313,15 @@ public function read()
                 return $data;
 
             case 'bulkString':
```
chayim, line 315: const string comment from Resp3Strategy applies here. Then, you can case on the const

src/Command/Redis/HSCAN.php
vladvildanov: Then it's makes no sense. This method is called from outside in case if RESP3 is active and you need additional response processing for specific command. It works the same as already existing `parseResponse()` method

src/Protocol/Parser/Strategy/Resp3Strategy.php
vladvildanov: Done!

src/Connection/StreamConnection.php
vladvildanov: Done!

PR #1303, opened 2023-05-24
Added support for sharded pub/sub consumer

### LINKED ISSUE #971: Implement Redis 7 command SPUBLISH
Implement the Redis 7 command SPUBLISH as specified at https://redis.io/commands/SPUBLISH.

### LINKED ISSUE #972: Implement Redis 7 command SSUBSCRIBE
Implement the Redis 7 command SSUBSCRIBE as specified at https://redis.io/commands/SSUBSCRIBE.

### LINKED ISSUE #973: Implement Redis 7 command SUNSUBSCRIBE
Implement the Redis 7 command SUNSUBSCRIBE as specified at https://redis.io/commands/SUNSUBSCRIBE.

vladvildanov:
1. Added support for `SSUBSCRIBE`, `SUNSUBSCRIBE`, `SPUBLISH` commands.
2. Updated pub/sub `Consumer` class, to allow run over cluster connection in sharded pub/sub context.
3. Updated `RedisCluster` connection to be able to loop over sockets and read incoming messages from Redis server.
4. Updated `DispatcherLoop` to run pub/sub commands depends on subscription context.
5. Added example for sharded pub/sub consumer usage and dispatcher loop.

Closes #971 
Closes #972 
Closes #973

### DIFF
--- a/examples/sharded_dispatcher_loop.php
+++ b/examples/sharded_dispatcher_loop.php
@@ -11,7 +11,6 @@
  */
 
 use Predis\Client;
-use Predis\Consumer\PubSub\SubscriptionContext;
 
 require __DIR__ . '/shared.php';
 


src/Cluster/ClusterStrategy.php
```diff
@@ -388,6 +393,24 @@ protected function getKeyFromZsetAggregationCommands(CommandInterface $command)
...
+     * @param  CommandInterface $command
+     * @return string
+     */
+    protected function getKeyFromSUnsubscribeCommand(CommandInterface $command): ?string
+    {
+        $arguments = $command->getArguments();
+
+        // SUNSUBSCRIBE command could be called without arguments, so it doesn't matter on each node it will be called.
+        if (empty($arguments)) {
+            return 'fake';
```
leibale, line 408: "SUNSUBSCRIBE command could be called without arguments, so it doesn't matter on ~each~ *which* (?) node it will be called."

why `return 'fake';` and not `return null;`?

src/Connection/Cluster/RedisCluster.php
```diff
@@ -691,4 +705,51 @@ public function getParameters(): ?ParametersInterface
...
+                }
+
+                return $connection->read();
+            }
+
+            if (!isset($this->sockets[++$i])) {
+                $i = 0;
+            }
+
+            usleep(1000);
```
leibale, line 740: make that configurable?

src/Cluster/ClusterStrategy.php
leibale: also, why this function is here and not in the `SUNSUBSCRIBE` class?

src/Connection/Cluster/RedisCluster.php
```diff
@@ -691,4 +705,51 @@ public function getParameters(): ?ParametersInterface
...
+    {
+        $this->setSocketsNonBlockingMode();
+
+        $i = 0;
+        while (true) {
+            $socket = $this->sockets[$i]['socket'];
+            $socketArray = [$socket];
+            $write = null;
+            $except = null;
+            $num = stream_select($socketArray, $write, $except, 0);
```
leibale, line 724: either use `stream_select` and pass it all the sockets in one call, or use `socket_select` if there is a reason to read them one by one

src/Cluster/ClusterStrategy.php
vladvildanov: 1. In Predis we're using pre-calculation (same algorithm that redis do) of slots to determine node that should accept command, it works well for commands that runs against specific key/keys, but if there's no key you still need to pre-calculate node so it's just random string returns to allow this action.
2. It's just because of bad design, I'm totally agree it should be a part of command object and do not violate SRP, but fixing it is beyond the scope of RESP3

src/Connection/Cluster/RedisCluster.php
vladvildanov: Yeah, will be more flexible

src/Connection/Cluster/RedisCluster.php
```diff
@@ -64,6 +69,11 @@ class RedisCluster implements ClusterInterface, IteratorAggregate, Countable
     private $retryLimit = 5;
     private $retryInterval = 10;
 
+    /**
+     * @var resource[]
+     */
+    private $sockets;
```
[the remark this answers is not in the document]
vladvildanov: Okay, will do something more specific

src/Connection/Cluster/RedisCluster.php
vladvildanov: It's some kind of event loop so it's constantly loop over streams if there's some data to read, so we can't process all of them in the same time. I know there's a confusion between streams and sockets here, the reason is that it's actually streams but in sockets context, that why it called sockets across the code, so I decided to keep this naming so it won't   confuse.

src/Cluster/ClusterStrategy.php
leibale: 1. but that means that the node that corresponds to `'fake'` (slot 14787) will be used more than other nodes..

src/Connection/Cluster/RedisCluster.php
leibale: @vladvildanov can't you do something like [this](https://www.php.net/manual/en/function.stream-select.php#122839) (run `stream_select` for all the open sockets, then if it returns a number > 0 loop on the sockets)

src/Cluster/ClusterStrategy.php
vladvildanov: For `SUNSUBSCRIBE` command only which is okay??, since it's not something you're often sends to server, but I can use randomly generated string here instead

src/Connection/Cluster/RedisCluster.php
vladvildanov: @leibale I can, but still only 1 response per read request can be returned (in case if all sockets have data to read). PHP is sync script language, so only one response per object lifecycle

vladvildanov: @leibale So I can `stream_select` all of them, but in case if `$num > 0` it's still only one could be processed

chayim (approved):

PR #1310, opened 2023-06-06
Added condition to prevent count error

### LINKED ISSUE #1309: count(): Argument #1 ($value) must be of type Countable|array, int given in HGETALL
**Describe the bug**
Error thrown randomly for HGETALL command.

**To Reproduce**
We are unable to reproduce it at will, but this appears in our logs every few days. 

**Expected behavior**
We expect this error to not be thrown.

- Predis: 2.1.2
- PHP 8.2.6 
- Redis Server 6.2.12
- OS CentOS (Rocky)

**Code sample**
N/A

**Additional context**
 
count(): Argument #1 ($value) must be of type Countable|array, int given {"exception":"[object] (TypeError(code: 0): count(): Argument #1 ($value) must be of type Countable|array, int given at /home/ss/application/vendor/predis/predis/src/Command/Redis/HGETALL.php:37)
[stacktrace]
#0 /home/ss/application/vendor/predis/predis/src/Command/Redis/HGETALL.php(37): count(0)
#1 /home/ss/application/vendor/predis/predis/src/Client.php(359): Predis\\Command\\Redis\\HGETALL->parseResponse(0)
#2 /home/ss/application/vendor/predis/predis/src/Client.php(303): Predis\\Client->executeCommand(Object(Predis\\Command\\Redis\\HGETALL))

It look like in this function is called at random times for some random reason with $data is coming in as an int which is blowing up as count() is not expecting an int. 

public function parseResponse($data)
    {
        $result = [];

        for ($i = 0; $i < count($data); ++$i) {
            $result[$data[$i]] = $data[++$i];
        }

        return $result;
    }

I think changing it this would be a safe code change and solve this issue (since the code say that $data is array|null|string): 

public function parseResponse($data)
    {
        $result = [];

        if (is_countable($data)) {
           for ($i = 0; $i < count($data); ++$i) {
              $result[$data[$i]] = $data[++$i];
           }
        }

        return $result;
    }

Thanks!

vladvildanov:
Closes #1309

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -346,7 +346,7 @@ class Client implements ClientInterface, IteratorAggregate
     }
 
     /**
-     * @param $name
+     * @param                     $name
      * @return ContainerInterface
      */
     public function __get($name)
@@ -355,8 +355,8 @@ class Client implements ClientInterface, IteratorAggregate
     }
 
     /**
-     * @param $name
-     * @param $value
+     * @param        $name
+     * @param        $value
      * @return mixed
      */
     public function __set($name, $value)
@@ -365,7 +365,7 @@ class Client implements ClientInterface, IteratorAggregate
     }
 
     /**
-     * @param $name
+     * @param        $name
      * @return mixed
      */
     public function __isset($name)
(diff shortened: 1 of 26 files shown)


tillkruss:
Let's not include various style changes in a bug fix PR.

PR #1316, opened 2023-06-12
Push notifications support

### LINKED ISSUE #1308: RESP3: Push notifications
1. Add new consumer for all upcoming messages.
2. Add support for callback bindings depends on push data type.
3. Add test coverage.
4. Add example.

vladvildanov:
Closes #1308


examples/push_notifications.php
```diff
@@ -0,0 +1,49 @@
...
+//
+// PUBLISH channel message1
+// PUBLISH channel message2
+// PUBLISH channel message3
+// PUBLISH control terminate
+// Data types should be changed in near future. Instead of Message data type it should be one of kind data types.
+
+foreach ($push as $notification) {
+    if ((null !== $notification) && $notification->getDataType() === PushResponseInterface::MESSAGE_DATA_TYPE) {
+        if ($notification[1] === 'control' && $notification[2] === 'terminate') {
```
chayim, line 39: How about moving control and terminate into things we can import + test against (const-y)

examples/push_notifications_dispatcher.php
```diff
@@ -0,0 +1,68 @@
...
+// Send following commands via redis-cli to test:
+//
+// PUBLISH channel message1
+// PUBLISH channel message2
+// PUBLISH channel message3
+// PUBLISH control terminate
+// Data types should be changed in near future. Instead of Message data type it should be one of kind data types.
+
+$dispatcher->attachCallback(
+    PushResponseInterface::MESSAGE_DATA_TYPE,
```
chayim, line 46: YES!

src/Connection/StreamConnection.php
```diff
@@ -303,6 +306,13 @@ public function read()
 
         switch ($parsedData['type']) {
             case 'push':
+                $data = [];
+
+                for ($i = 0; $i < $parsedData['value']; ++$i) {
```
chayim, line 311: forach with arr.push?

tests/Predis/ClientTest.php
```diff
@@ -1048,7 +1048,7 @@ public function testPubSubLoopWithoutArgumentsReturnsPubSubConsumer(): void
     {
         $client = new Client();
 
-        $this->assertInstanceOf('Predis\PubSub\Consumer', $client->pubSubLoop());
+        $this->assertInstanceOf('Predis\Consumer\PubSub\Consumer', $client->pubSubLoop());
```
chayim, line 1051: TBH I preferred the old pathing. It looks prettier. Any way to approach it?

vladvildanov: @chayim It was done to keep different entities within the same domain. There are a lot of "Consumers": PubSub, Sharded PubSub, Push Notifications. All of them "consume" messages from server. IMHO, it's better to create a separate place where we can keep all entities related to this domain, then spreading them across whole library.

src/Connection/StreamConnection.php
vladvildanov: This one defines a behaviour of reader in case if it's Push notification. Since, we're read response line by line, we need to run foreach depends of array.length that we get from server.

chayim (approved):

Issue #1305, opened 2023-05-30
Key prefix is not used for GETDEL command


Pazns:
**Describe the bug**
When issuing a GETDEL command,
I expect to see Redis doing the GETDEL command on the requested key prefixed with the key prefix configured in the Redis client.

**To Reproduce**
Steps to reproduce the behaviour:
1. Initialize `Predis\Client` class with an option array containing they key-value `prefix: _myprefix_`
2. Issue the GETDEL command with `$client->getdel('mykey')`

**Expected behavior**
Redis must receive a GETDEL command for the key `_myprefix_mykey`.

**Observed behaviour**
Redis wrongly receives a GETDEL command for the key `mykey`.

**Versions**
- Predis 2.1.2
- PHP 8.1.16
- Redis 7.0.9
- Debian 11 

**Additional context**
I believe it's the same issue as #1286.
The KeyPrefixProcessor class seems to be lacking the GETDEL command.

PR #1318, opened 2023-06-16
Clusterable interface for redis commands

vladvildanov:
The goal of this PR is to get rid off current implementation of filtering for commands that could be executed against OSS cluster. For now, each command should be specified within a dictionary in `ClusterStrategy`. So whenever we implement a new command the code should be changed in two different places, so it violates some of SOLID principles and make it hard to maintain.

The enhancement based on another case that similar to this one:
https://github.com/predis/predis/pull/1262

Here's a proposal of new implementation:
1. Introduce `Clusterable` interface that represents command that can be executed against cluster.
2. `ClusterStrategy` filtering now based on `Clusterable` interface.
3. `ClusterableContract` trait contains API that allows to complete `Clusterable` interface contract.
4. Commands that currently supported execution against cluster will be marked with `Clusterable` interface so from user perspective all functionality will be kept.
5. Introduce `getKeys()` method within `CommandInterface`, so now each command itself should provide information about the keys, instead of detecting them within `ClusterStrategy`.

Enhancements:
1. `ClusterStrategy` object wont be changed in case if new command will be introduced, or old one will allowed to be executed against cluster (no more bugs like this: #1368, #1352, #1305). Removed violation of SOLID OCP principle.
2. `ClusterStrategy` object now isn't responsible for detecting command keys, moved on a command object level. It's more natural when the information about command keys stores in the command itself, since it's a part of command domain. Removed violation of SOLID SRP principle.
3. We keep only one responsibility for `ClusterStrategy` - calculate a hash slot depends on given cluster context. In future, we should get rid off strategies at all and move hash slot calculation into command object as well (according to Information Expert pattern), but this require massive changes that I won't add to this PR.


vladvildanov:
@tillkruss @chayim I want you to have a look on the draft of PR and lmk what you think about this changes. Don't want to spend so much time for it if it makes no sense 👌

PR #1327, opened 2023-06-20
Added support for XREADGROUP command

### LINKED ISSUE #1326: Add support for missing redis command XREADGROUP

vladvildanov:
Closes #1326


tillkruss (approved):

chayim (changes requested):
Let's get more tests in place. XREADGROUP accepts key [key ...] id [id...] , so I think we're missing at least one test case (multiple if you think of ID as a list).

vladvildanov:
@chayim Done!

chayim (approved):

PR #1339, opened 2023-07-20
Fix prefix error if prefix equals a php command

mmodler:
Fix php error if prefix equals a php command like "chr" for example.

Replaced is_callable with instanceof Closure, which should preserve the desired behavior.

### DIFF
--- a/src/Configuration/Option/Prefix.php
+++ b/src/Configuration/Option/Prefix.php
@@ -12,6 +12,7 @@
 
 namespace Predis\Configuration\Option;
 
+use Closure;
 use Predis\Command\Processor\KeyPrefixProcessor;
 use Predis\Command\Processor\ProcessorInterface;
 use Predis\Configuration\OptionInterface;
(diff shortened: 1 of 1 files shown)


mmodler:
I'm not sure if its security related and justifies a hotfix release?

tillkruss:
This would remove some functionality, please explain what this fixes with a test case. I don't understand what this resolves.

mmodler:
Set the prefix to a php command like `chr`

`$client = new Predis\Client($parameters, ['prefix' => 'chr']);`

Then is_callable() is true and the command is executed with  call_user_func().

tillkruss (changes requested):
Tests are failing.

mmodler:
Did you get the point?

tillkruss:
No, we need unit tests to explain this change, since it **narrowing** the accepted type from `callable` to `Closure`.

mmodler:
Hm, do you understand the problem and someone needs to fix the tests? Or do you ask me WHY this should be narrowed?

tillkruss:
Yes, I understand the issue, but we don't merge PRs without tests.

PR #1344, opened 2023-07-25
Added client name on connection to server

### LINKED ISSUE #1338: Clients should identify themselves on connect
[CLIENT SETINFO](https://redis.io/commands/client-setinfo/) is a new fire-and-forget command that is available as part of redis 7.2. Clients should send the lowercased version of the client name (i.e redis-py, jedis, go-redis) as part of the initial connect, after authenticating to the server.

vladvildanov:
Closes #1338


src/Connection/Factory.php
```diff
@@ -173,6 +173,10 @@ protected function prepareConnection(NodeConnectionInterface $connection)
             );
         }
 
+        $connection->addConnectCommand(
+            new RawCommand('CLIENT', ['SETINFO', 'LIB-NAME', 'predis'])
```
tillkruss, line 177: Should (or can) we also set the version?

vladvildanov:
@tillkruss Relay tests are failing because `executeCommand()` throws exception instead of returning `ErrorInterface`. There's design problems with `RelayConnection` it calls `StreamConnection::connect()` that calls `RelayConnection::executeCommand()` instead of `AbstractConnection::executeCommand()`. Would you mind to fix it?

tillkruss (approved):

tillkruss:
> @tillkruss Relay tests are failing because `executeCommand()` throws exception instead of returning `ErrorInterface`. There's design problems with `RelayConnection` it calls `StreamConnection::connect()` that calls `RelayConnection::executeCommand()` instead of `AbstractConnection::executeCommand()`, so the child overridden method is called instead of the parent. Can you take a look on it?

I don't have bandwidth this week for it, can you open a PR?

PR #1347, opened 2023-07-27
Client set name and version on connection

### LINKED ISSUE #1338: Clients should identify themselves on connect

vladvildanov:
Closes #1338

### DIFF
--- a/src/Connection/Factory.php
+++ b/src/Connection/Factory.php
@@ -13,6 +13,7 @@
 namespace Predis\Connection;
 
 use InvalidArgumentException;
+use Predis\Client;
 use Predis\Command\RawCommand;
 use ReflectionClass;
 use UnexpectedValueException;
@@ -173,6 +174,14 @@ class Factory implements FactoryInterface
                 new RawCommand('AUTH', $cmdAuthArgs)
             );
         }
+
+        $connection->addConnectCommand(
+            new RawCommand('CLIENT', ['SETINFO', 'LIB-NAME', 'predis'])
+        );
+
+        $connection->addConnectCommand(
+            new RawCommand('CLIENT', ['SETINFO', 'LIB-VER', Client::VERSION])
+        );
 
         if (isset($parameters->database) && strlen($parameters->database)) {
             $connection->addConnectCommand(
--- a/src/Connection/Factory.php
+++ b/src/Connection/Factory.php
@@ -175,13 +175,15 @@ class Factory implements FactoryInterface
             );
         }
 
-        $connection->addConnectCommand(
-            new RawCommand('CLIENT', ['SETINFO', 'LIB-NAME', 'predis'])
-        );
+        if (!$connection instanceof RelayConnection) {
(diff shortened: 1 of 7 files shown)


vladvildanov:
@tillkruss Adding a fix for Relay connection, leads to an error in CI

https://github.com/predis/predis/actions/runs/5678238422/job/15388057449?pr=1347

Trying to run Relay tests locally, but have a problems with relay installation over homebrew.

`Error: An exception occurred within a child process:
  MachO::DylibUnknownError: No such dylib name: /usr/local/lib/libhiredis_ssl.dylib.1.1.0`
  
 This file actually exists and I have correct version of `hiredis`

tillkruss:
Aah, maybe Relay is already setting `CLIENT SET-NAME` is causing the issue.

>Trying to run Relay tests locally, but have a problems with relay installation over homebrew.
>Error: An exception occurred within a child process: MachO::DylibUnknownError: No such dylib name: /usr/local/lib/libhiredis_ssl.dylib.1.1.0

Relay via Homebrew uses OpenSSL v3, can you try `brew uninstall relay` and reinstall it?

src/Connection/Factory.php
```diff
@@ -174,6 +175,16 @@ protected function prepareConnection(NodeConnectionInterface $connection)
             );
         }
 
+        if (!$connection instanceof RelayConnection) {
```
chayim, line 178: Could these all be pipelined? Speed up the connection a bit?

chayim (approved):

tillkruss (approved):

BafS:
1. In this PR add 2 commands on initialization, that's a lot of IO. It should at least be an option IMO. We saw an important load increase in production due to this PR.
2. Why was it bundled in a patch version?

tillkruss:
>We saw an important load increase in production due to this PR.

Can you elaborate on this and provide numbers?

BafS:
We didn't have "client" call before

image

And here you can see the substantial change in CPU in the cluster:

image

it took us a while to find the culprit because we didn't expect to have a change in behavior in a patch version

vladvildanov:
@tillkruss I talked about this with other client-developers and they confirm that this should be an optional. 

We can't fix it other way since it's a Redis API and we must call it twice each time. Let's merge #1370, but before we need to merge #1369 that fixes CI after Redis 7.2 update

BafS:
Thank you for your feedback and thanks for helping fixing it.

Don't you think that this PR should be reverted for the next patch release (to revert the issue introduced) and the new feature (with the new `set_client_info` defaulting to `false` to not have BC) should be in a new minor version?

vladvildanov:
@BafS Totally works for me, but @tillkruss is the releases responsible person

PR #1354, opened 2023-08-06
[3.x] Spellcheck as part of CI

chayim:
As part of extending CI and providing a better experience working with the repository and codebase itself, I'm starting by adding some scaffolding, spelling CI.


vladvildanov:
@chayim we already have a spellcheck in our CI

https://github.com/predis/predis/blob/v2.x/.github/workflows/linters.yml#L146

.github/workflows/tests.yml
```diff
@@ -5,7 +5,6 @@ on:
     branches:
       - main
       - v2.**
-      - v3.**
```
vladvildanov, line 8: We don't need to run tests on 3.x?

PR #1355, opened 2023-08-06
[main] Spellcheck as part of CI

chayim:
As part of extending CI and providing a better experience working with the repository and codebase itself, I'm starting by adding some scaffolding, spelling CI.


.github/workflows/spellcheck.yml
```diff
@@ -0,0 +1,14 @@
...
+  check-spelling:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Check Spelling
+        uses: rojopolis/spellcheck-github-actions@0.33.1
+        with:
+          config_path: .github/spellcheck-settings.yml
+          task_name: Markdown
```
tillkruss, line 14: Does it have to be in it's on file? Why not keep it in `linters.yml`?

tillkruss (changes requested):

.github/workflows/spellcheck.yml
chayim: maintainability - it still reports back to GH.

PR #1356, opened 2023-08-08
Extended core support by changing arguments interface for SET command

### LINKED ISSUE #992: Support new Redis 6.2 arguments for SET

vladvildanov:
Closes #992

### DIFF
--- a/src/ClientContextInterface.php
+++ b/src/ClientContextInterface.php
@@ -147,7 +147,7 @@ use Predis\Command\Container\XGROUP;
  * @method $this mset(array $dictionary)
  * @method $this msetnx(array $dictionary)
  * @method $this psetex($key, $milliseconds, $value)
- * @method $this set($key, $value, bool $getOld = false, $expireResolution = null, $expireTTL = null, $flag = null)
+ * @method $this set($key, $value, $expireResolution = null, $expireTTL = null, $flag = null, bool $getOld = false)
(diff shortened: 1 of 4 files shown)


src/ClientContextInterface.php
```diff
@@ -147,7 +147,7 @@
  * @method $this mset(array $dictionary)
  * @method $this msetnx(array $dictionary)
  * @method $this psetex($key, $milliseconds, $value)
- * @method $this set($key, $value, $expireResolution = null, $expireTTL = null, $flag = null)
+ * @method $this set($key, $value, bool $getOld = false, $expireResolution = null, $expireTTL = null, $flag = null)
```
tillkruss, line 150: This will be a major breaking change. Do we need to break the argument order?

tillkruss (changes requested):

src/ClientContextInterface.php
vladvildanov: @tillkruss No, we actually don't. The only reason I did like this is just to keep and order that Redis describes in docs, but I found that it can be in any order, so I just fix it to not being breaking change

tillkruss (approved):

PR #1357, opened 2023-08-08
Added support for WAITAOF command

### LINKED ISSUE #1239: Add support for new redis command WAITAOF
7.2.0 adds support for the new redis command WAITAOF. We need to add support, documented [here](https://redis.io/commands/WAITAOF).

vladvildanov:
Closes #1239

### DIFF
--- a/src/Command/Redis/WAITAOF.php
+++ b/src/Command/Redis/WAITAOF.php
@@ -1,0 +1,29 @@
+<?php
+
+/*
+ * This file is part of the Predis package.
+ *
+ * (c) 2009-2020 Daniele Alessandri
+ * (c) 2021-2023 Till Krüss
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Predis\Command\Redis;
+
+use Predis\Command\Command as RedisCommand;
+
+/**
+ * @see https://redis.io/commands/waitaof/
+ *
+ * This command blocks the current client until all the previous write commands are acknowledged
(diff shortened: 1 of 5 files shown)


tillkruss (approved):

src/Command/Redis/WAITAOF.php
```diff
@@ -0,0 +1,29 @@
...
+
+use Predis\Command\Command as RedisCommand;
+
+/**
+ * @see https://redis.io/commands/waitaof/
+ *
+ * This command blocks the current client until all the previous write commands are acknowledged
+ * as having been fsynced to the AOF of the local Redis and/or at least the specified number of replicas.
+ */
+class WAITAOF extends RedisCommand
```
chayim, line 23: @vladvildanov 7.2.0 only we can do in 3.x yeah?
But also - missing numlocal, numreplicas, timeout [command docs](https://redis.io/commands/waitaof/)

vladvildanov: @chayim Actually, Redis server version shouldn't affect our predis version. Users with predis 2.x should be able to use it with 7.2.0, we do not break anything, so I my vote for including this changes in 2.x and 3.x.

We're not missing anything, here's command signature:

https://github.com/predis/predis/pull/1357/files#diff-7fe9d7bab1528c58fb39365621c6f02430dd01828c26fe8b00f0408de72c71b8R325

tillkruss: Yeah, v3.0 is just architecture changes of Predis itself.

chayim (approved):

PR #1362, opened 2023-08-09
Revert "[3.x] Spellcheck as part of CI"

tillkruss:
Reverts predis/predis#1354.


tillkruss:
@chayim @vladvildanov: Let's stick to PR reviews before merging them.

PR #1369, opened 2023-08-21
Added EXPIRETIME command to KeyPrefixProcessor

### LINKED ISSUE #1368: Expiretime ignores key prefix
**Describe the bug**
`expiretime` ignores key prefix

**To Reproduce**
1. Create Client with key prefix set
2. Add a key with expiration time
3. Call expiretime
4. expiretime returns -2. If you run expiretime again with prefix set manually, it will return correct timestamp

**Expected behavior**
Expiretime should not ignore key prefix

- Predis: 2.2.1
- PHP 8.1
- Redis Server: 7.0.5

vladvildanov:
Closes #1368

### DIFF
--- a/tests/Predis/Command/Redis/OBJECT_Test.php
+++ b/tests/Predis/Command/Redis/OBJECT_Test.php
@@ -82,7 +82,7 @@ class OBJECT_Test extends PredisCommandTestCase
 
     /**
      * @group connected
-     * @requiresRedisVersion >= 2.2.3
+     * @requiresRedisVersion < 7.2.0
      */
     public function testObjectEncoding(): void
     {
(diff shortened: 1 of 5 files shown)


tillkruss (approved):

tests/Predis/Command/Redis/OBJECT_Test.php
```diff
@@ -92,6 +92,19 @@ public function testObjectEncoding(): void
         $this->assertMatchesRegularExpression('/[zip|quick]list/', $redis->object('ENCODING', 'list:metavars'));
```
kylekatarnls, line 92: As per my test, only `quicklist` can happen here.

Also `[zip|quick]` should be `(zip|quick)` and have `^` and `$` if you mean to check for ziplist or quicklist

PR #1370, opened 2023-08-24
Add parameter opt out for setinfo commands

kylekatarnls:


src/Connection/Factory.php
```diff
@@ -176,13 +176,17 @@ protected function prepareConnection(NodeConnectionInterface $connection)
...
+                $connection->addConnectCommand(
+                    new RawCommand('CLIENT', ['SETINFO', 'LIB-NAME', 'predis'])
+                );
+            }
+
+            if ($parameters->set_lib_version ?? true) {
+                $connection->addConnectCommand(
+                    new RawCommand('CLIENT', ['SETINFO', 'LIB-VER', Client::VERSION])
+                );
+            }
```
Okhoshi, line 189:
```suggestion
            if ($parameters->set_client_info ?? true) {
                $connection->addConnectCommand(
                    new RawCommand('CLIENT', ['SETINFO', 'LIB-NAME', 'predis'])
                );
                
                $connection->addConnectCommand(
                    new RawCommand('CLIENT', ['SETINFO', 'LIB-VER', Client::VERSION])
                );
            }
```

src/Connection/ParametersInterface.php
```diff
@@ -31,6 +31,8 @@
  * @property string $database           Database index (see the SELECT command).
  * @property bool   $async_connect      Performs the connect() operation asynchronously.
  * @property bool   $tcp_nodelay        Toggles the Nagle's algorithm for coalescing.
+ * @property bool   $set_lib_name       Set to false not the set the LIB-NAME on connect.
+ * @property bool   $set_lib_version    Set to false not the set the LIB-VER on connect.
```
Okhoshi:
```suggestion
 * @property bool   $set_client_info    Sends the CLIENT INFO on connect (defaults to true).
```

Okhoshi (commented):
Seems overkill to have 2 parameters to control this

kylekatarnls:
Fix for https://github.com/predis/predis/pull/1347#issuecomment-1691308594

src/Connection/ParametersInterface.php
```diff
@@ -31,6 +31,8 @@
  * @property string $database           Database index (see the SELECT command).
  * @property bool   $async_connect      Performs the connect() operation asynchronously.
  * @property bool   $tcp_nodelay        Toggles the Nagle's algorithm for coalescing.
+ * @property bool   $set_lib_name       Set to `false` to not set the LIB-NAME on connect.
+ * @property bool   $set_lib_version    Set to `false` to not set the LIB-VER on connect.
```
Okhoshi, line 35:
```suggestion
 * @property bool   $set_client_info    Sends the CLIENT INFO on connect (defaults to true).
```

src/Connection/Factory.php
kylekatarnls: @vladvildanov @tillkruss Feel free to apply those changes if it makes better sense to group both. 👌

BafS: Also I would favor `?? false` to have a sain configuration by default (to add overhead by default).

tillkruss (dismissed):

src/Connection/Factory.php
```diff
@@ -176,13 +176,15 @@ protected function prepareConnection(NodeConnectionInterface $connection)
...
-            );
-
-            $connection->addConnectCommand(
-                new RawCommand('CLIENT', ['SETINFO', 'LIB-VER', Client::VERSION])
-            );
+            if ($parameters->set_client_info ?? true) {
+                $connection->addConnectCommand(
+                    new RawCommand('CLIENT', ['SETINFO', 'LIB-NAME', 'predis'])
+                );
+
```
kylekatarnls, line 183:
```suggestion

```

kylekatarnls:
Trimmed the whitespace eclint complained about:
https://github.com/predis/predis/actions/runs/5967543294/job/16189493022?pr=1370#step:6:17

tillkruss (approved):

vladvildanov:
Test coverage?

kylekatarnls:
Test added. You need to "Approve and run workflows" to trigger tests and get result from `coveralls`

kylekatarnls:
Redis 7 errors unrelated (might have been introduced by last week update of Redis 7.2)

Can be verified with #1372

tests/Predis/Connection/FactoryTest.php
```diff
@@ -322,6 +322,39 @@ public function testCreateConnectionWithInitializationCommands(): void
...
+            ->method('addConnectCommand')
+            ->withConsecutive(
+                [$this->isRedisCommand('AUTH', ['foobar'])],
+                [$this->isRedisCommand('SELECT', ['0'])]
+            );
+
+        $factory = new Factory();
+
+        // TODO: using reflection to make a protected method accessible :facepalm:
+        $reflection = new ReflectionObject($factory);
```
vladvildanov, line 352: We can avoid using reflection object here, just call the public interface (e.g `$factory->create()`), `prepareConnection()` will be called as a part of execution script

kylekatarnls: IMO there should be a refactor of the whole file for that for consistency. It's all over the place.

vladvildanov: Agreed, there's many things that can be refactored, but we still can avoid using Reflection, would you mind If I help you to fix all this issues in CI?

vladvildanov: UPD: Let's leave it for now, since all the tests are using this style of Reflection. Will be a part of another refactoring

vladvildanov (approved):

src/Connection/Parameters.php
```diff
@@ -25,6 +26,9 @@ class Parameters implements ParametersInterface
         'scheme' => 'tcp',
         'host' => '127.0.0.1',
         'port' => 6379,
+        'set_client_info' => true,
+        'client_name' => 'predis',
+        'client_version' => Client::VERSION,
```
tillkruss, line 31: Should these be configurable @vladvildanov?

vladvildanov: @tillkruss It is configurable, check [there](https://github.com/predis/predis/pull/1370/files#diff-7171e490ba29ca3a384a9f5d48189766b2f8c9f8af0371d523c8482770d65a0eR36). In this place we set a default values

tillkruss: I mean should it be? Why wouldn't it always be Predis, if it's enabled?

tillkruss (changes requested):
Clarify `client_*` options.

vladvildanov:
@tillkruss https://github.com/predis/predis/pull/1370/files#diff-7171e490ba29ca3a384a9f5d48189766b2f8c9f8af0371d523c8482770d65a0eR35

Isn't it clear enough?

src/Connection/Parameters.php
kylekatarnls: As for what I'm concerned, we just need a way to opt-out to solve the performance issue. IMO [Added overriding for client name and version, increased test coverage](https://github.com/predis/predis/pull/1370/commits/1f6c5525e1d47f7b2167d26419f5f2365dc2775f) is another feature, this could be submitted and discussed in a separated merge-request.

tillkruss: Yeah, let's go with only `set_client_info` for now, so we can release it. 

I don't see the need for name/version configuration.

vladvildanov:
@tillkruss I believe @chayim knows more than me about that. It looks like clients should be able to advertise their own products this way. F.E if magento uses predis as Redis client, they can override this values, so cloud instances can collect this kind of data about magento clients.

src/Connection/Parameters.php
tillkruss: https://github.com/predis/predis/pull/1370#issuecomment-1697735969

chayim:
The reason for allowing setting names and versions is to enable "whatever people want to create". It's entirely possible to build client solutions on top of existing clients (nredisstack and stackexchange.redis for example). This is the vision of how these things would be used.

It also allows (same example) for a client to advertise something like:
Name: Foo [predis]

Thus, sharing all.

tillkruss (approved):

tillkruss:
@kylekatarnls: I'll merge a quick fix in #1399, because #1398 is making this more complicated.

PR #1372, opened 2023-08-25
Fix Redis 7 test compatibility

kylekatarnls:
`testObjectEncoding`
Make expected type explicitly depending on Redis version.

`testDryRunSimulateExecutionOfGivenCommandByUser`
Align message with new format on Redis 7.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -62,14 +62,20 @@ jobs:
       - name: Run tests
         if: ${{ matrix.php != '8.1' || matrix.redis != '7' }}
         run: vendor/bin/phpunit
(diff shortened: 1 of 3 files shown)


vladvildanov:
@kylekatarnls This changes already fixed within another PR #1369, closing as it's a duplicate

kylekatarnls:
OK, please not that I think `assertMatchesRegularExpression('/[zip|quick]list/'` is working by luck but likely not have the intended result as `[]` classes select only 1 character and the whole expression has no "start-with" / "end-with" constraint: https://github.com/predis/predis/pull/1369/files#r1305409415

PR #1394, opened 2023-09-05
ClientInterface: using set with NX flag can return null

glaubinix:
Calling `$client->set('key', '1', 'EX', 7200, 'NX')` for a key that is already set will return null instead of Status

### DIFF
--- a/src/ClientInterface.php
+++ b/src/ClientInterface.php
@@ -155,7 +155,7 @@ use Predis\Response\Status;
  * @method mixed             mset(array $dictionary)
  * @method int               msetnx(array $dictionary)
  * @method Status            psetex(string $key, $milliseconds, $value)
- * @method Status            set(string $key, $value, $expireResolution = null, $expireTTL = null, $flag = null)
+ * @method Status|null       set(string $key, $value, $expireResolution = null, $expireTTL = null, $flag = null)
  * @method int               setbit(string $key, $offset, $value)
  * @method Status            setex(string $key, $seconds, $value)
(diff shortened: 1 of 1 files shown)


tillkruss (approved):

src/ClientInterface.php
```diff
@@ -155,7 +155,7 @@
  * @method mixed             mset(array $dictionary)
  * @method int               msetnx(array $dictionary)
  * @method Status            psetex(string $key, $milliseconds, $value)
- * @method Status            set(string $key, $value, $expireResolution = null, $expireTTL = null, $flag = null)
+ * @method Status|null       set(string $key, $value, $expireResolution = null, $expireTTL = null, $flag = null)
```
tillkruss, line 158: Can you add a test case for this?

tillkruss (changes requested):

src/ClientInterface.php
glaubinix: Happy to add a test. Can you point me into the right direction? Are you looking for a test in `ClientTest` or are there integrations tests that interact with Redis?

tillkruss: I'd suggest adding a new method to `SET_Test.php` and doing the `$this->assetNull()` to ensure that's the case.

glaubinix: It appears there already is a test that asserts that calling `SET` with `NX` will return `null` if the key is set already: `testSetStringValueWithModifierNX`

tillkruss: Perfect.

PR #1397, opened 2023-09-07
Make Relay work with redis cluster

dorrogeray:
Currently, when predis client is configured to use a combination of Relay and Redis Cluster, it does not work correctly when redis replies with `MOVED`. Following set of fixes should get it working:

- Ensure connection is open when RelayConnection::getIdentifier is called
- Return error responses instead of throwing them in RelayConnection::executeCommand so that RedisCluster::executeCommand can intercept MOVED errors and handle them with RedisCluster::onErrorResponse
- Make parsing of connection id in RedisCluster::onMovedResponse more resilient by culling off details from the exception message - this is required because relay exception message format is "IP:port (details about exception)"

CC: @tillkruss


tillkruss:
Thanks! Can you re-enable some of the cluster tests to ensure this works?

phpunit.relay.xml
```diff
@@ -25,7 +25,6 @@
             <group>realm-stack</group>
             <group>ext-curl</group>
             <group>ext-phpiredis</group>
-            <group>cluster</group>
```
dorrogeray, line 28: @tillkruss did you mean this?

tillkruss: Yes, let's see if the tests run.

dorrogeray: @tillkruss I have run following locally with all tests passing on PHP 8.1 and Redis 7.2:

- `vendor/bin/phpunit` with relay extension disabled
- `vendor/bin/phpunit -c phpunit.relay.xml` with relay extension enabled

I found [one failing test which i fixed](https://github.com/predis/predis/blob/de89d9d4ec8e6c9e5bc83a72f54da9ecf942cf6e/tests/Predis/Command/Redis/CLUSTER_Test.php#L99), but that one was failing even before my changes (i tested that on the `v2.x` branch).

Can you please run the checks to see if they pass this time?

tillkruss: I’ve changed the permissions, so they should run automatically now.

dorrogeray: @tillkruss Turns out you cant remove `cluster` from exclusions in `phpunit.relay.xml` because `predis` tests run without redis cluster. There is a separate job `predis-cluster` which spins up a redis cluster, but that one defines the group explicitly to be `cluster`. I have therefore instead opted for expanding `predis-cluster` job to include a phpunit run with `phpunit.relay.xml`.

dorrogeray:
@tillkruss ready for review

src/Connection/RelayConnection.php
```diff
@@ -198,9 +199,10 @@ public function getClient()
     /**
      * {@inheritdoc}
      */
-    protected function getIdentifier()
+    public function getIdentifier()
     {
-        return $this->client->endpointId();
+        // Workaround until Relay::endpointId() stops throwing "Not Connected" error when not yet connected
```
tillkruss, line 204: I've raised this internally, and we'll get this resolved in the next release.

tillkruss (approved):

tillkruss:
Thanks for the work @dorrogeray on this. I've pushed a few formatting changes.

BTW, we're currently working on full `Relay\Cluster` support internally. /ping @yatsukhnenko

tillkruss:
Can we get your review @vladvildanov?

.gitignore
```diff
@@ -1,5 +1,6 @@
 /build
 /vendor
+.coveralls.yml
```
tillkruss, line 3: @szepeviktor How can we fix these? https://github.com/predis/predis/actions/runs/6140711631/job/16660032679?pr=1397

tests/Predis/Connection/Cluster/RedisClusterTest.php
```diff
@@ -1212,11 +1213,12 @@ public function testFetchSlotsMapFromClusterWithClusterSlotsCommand(): void
...
     /**
      * @group disconnected
+     * @dataProvider onMovedResponsesDataProvider
      */
-    public function testAskSlotMapToRedisClusterOnMovedResponseByDefault(): void
+    public function testAskSlotMapToRedisClusterOnMovedResponseByDefault(string $movedErrorMessage): void
     {
         $cmdGET = Command\RawCommand::create('GET', 'node:1001');
-        $rspMOVED = new Response\Error('MOVED 1970 127.0.0.1:6380');
+        $rspMOVED = new Response\Error($movedErrorMessage);
```
tillkruss, line 1221: @vladvildanov I just noticed that we're hardly performing ANY cluster operations. Should we see if we can run some of the command tests against cluster?

I gotta say, I hate cluster and never use it.

.gitignore
szepeviktor:
```diff
- .coveralls.yml,LICENSE,README.md,autoload.php,composer.json
+ LICENSE,README.md,autoload.php,composer.json
```

`.coveralls.yml` is added to the repo. It should not be ignored.
Stop exporting it in `.gitattributes`
```
.coveralls.yml export-ignore
```

tests/Predis/Connection/Cluster/RedisClusterTest.php
dorrogeray: That's an interesting statement, how do you solve fail-over in your use-cases without cluster? Or is it that you just don't need a fail-over in your use-cases?

.gitignore
szepeviktor, line 3: Please do not ignore added files.
```suggestion
```

src/Connection/RelayConnection.php
```diff
@@ -198,9 +199,11 @@ public function getClient()
     /**
      * {@inheritdoc}
      */
-    protected function getIdentifier()
+    public function getIdentifier()
     {
-        return $this->client->endpointId();
+        // Using `spl_object_id()` workaround until `Relay::endpointId()`
+        // works without requiring a connection
+        return (string) spl_object_id($this->client);
```
tillkruss, line 206: This will be more accurate.

```suggestion
        try {
            return $this->client->endpointId();
        } catch (RelayException $ex) {
            return parent::getIdentifier();
        }
```

tests/Predis/Connection/Cluster/RedisClusterTest.php
tillkruss: We just don't use any HA Redis setups.

tillkruss:
@dorrogeray: The PR looks great. The only concern that I have, which isn't directly related is `OK (3 tests, 4 assertions)`. 

Would you mind seeing if we can run the one simple Redis command test case against cluster?

https://github.com/predis/predis/blob/v2.x/tests/Predis/Command/Redis/GET_Test.php

dorrogeray:
@tillkruss I have added some cluster tests. The `$redis->del('metavars');` is needed because FLUSHDB flushes database on a single node, it does not clear the entire redis cluster. Leftovers from previous tests were causing problems in those particular tests..

tillkruss (approved):

Issue #1329, opened 2023-06-22
PHP 8.2 deprecation: Use of "static" in callables is deprecated


nicolas-grekas:
Both in branches 1.x and 2.x, KeyPrefixProcessor is relying on deprecated callables and triggers this deprecation, as explained in this RFC (there may be other classes affected, I didn't check):
https://wiki.php.net/rfc/deprecate_partially_supported_callables

tillkruss:
@nicolas-grekas Do you have a suggestion on how to avoid deprecated callables?

nicolas-grekas:
For v2.x, this has been fixed in #797
For v1.x I propose #1403

PR #1402, opened 2023-09-18
Fix prefix error if prefix equals a php command

mmodler:
Fix php error if prefix equals a php command like "chr" for example.

Replaced is_callable with instanceof Closure, which should preserve the desired behavior.

Hopefully better fix for #1339

### DIFF
--- a/src/Configuration/Option/Prefix.php
+++ b/src/Configuration/Option/Prefix.php
@@ -12,6 +12,7 @@
 
 namespace Predis\Configuration\Option;
 
+use Closure;
 use Predis\Command\Processor\KeyPrefixProcessor;
 use Predis\Command\Processor\ProcessorInterface;
 use Predis\Configuration\OptionInterface;
@@ -28,7 +29,7 @@ class Prefix implements OptionInterface
      */
     public function filter(OptionsInterface $options, $value)
     {
-        if (is_callable($value)) {
+        if (is_callable($value) && (is_object($value) || $value instanceof Closure)) {
             $value = call_user_func($value, $options);
         }
 


mmodler:
@tillkruss Now tests are passing, only code style test failing but has nothing to do with my change?

src/Configuration/Option/Prefix.php
```diff
@@ -28,7 +29,7 @@ class Prefix implements OptionInterface
      */
     public function filter(OptionsInterface $options, $value)
     {
-        if (is_callable($value)) {
+        if (is_callable($value) && (is_object($value) || $value instanceof Closure)) {
```
tillkruss, line 32: @vladvildanov Can we get your eyes on this?

The `is_callable()` check has been around for 7+ years and the issue is that if the prefix is set to `"chr"` or `"abs"` the PHP functions would be called.

I can't think of an elegant solution to prevent this, without creating edge cases.

tillkruss:
See #1329.

src/Configuration/Option/Prefix.php
vladvildanov: @tillkruss I think here it's enough to keep `$value instanceof Closure`, because that's exactly what we want to ensure. We have two different flows for string and Closure, so the only issue is that `is_callable()` consider strings matching php default functions as callable.

tillkruss: Okay, are you certain that only checking `$value instanceof Closure` won't break peoples installations?

vladvildanov: I am, but test coverage should prove this better

tillkruss: The `object` + `Closure` check makes sense, but I'm uncomfortable merging this because we have one bug report of this causing an issue and an unknown number of people using:

```php
$client = new Predis\Client($server, [
    'prefix' => 'my_prefix_callback',
    'prefix' => 'MyPrefix::callback',
]);
```

vladvildanov: @tillkruss Okay, lets leave `is_callable()` call for now, so people could see deprecation message and have time to fix it in their code. We can substitute all `call_user_func($callable)` with `$callable()` later on.

For now, we can add a hack like this `if (is_callable($prefix) && !function_exists($prefix))`, so prefixes corresponding to php function name won't consider as callables.

tillkruss: That seems reasonable, even tho it's a breaking change. @mmodler what do you think?

mmodler: I think this would solve the issue in my case.

src/Configuration/Option/Prefix.php
```diff
@@ -28,7 +29,7 @@ class Prefix implements OptionInterface
      */
     public function filter(OptionsInterface $options, $value)
     {
-        if (is_callable($value)) {
+        if (is_callable($value) && ! function_exists($value)) {
```
tillkruss, line 32: @mmodler Great, let's add a new test case that `"chr"` won't be called as a function and we can merge it.

tillkruss (changes requested):

tillkruss:
@vladvildanov Should we move this into v3?

vladvildanov:
@tillkruss +

tillkruss:
@vladvildanov yeah, wdyt?

PR #1408, opened 2023-10-19
added if condition befor fclose

jayesh-razorpay:
fixes fclose(): supplied resource is not a valid stream resource which (as per my understanding) may occur since PHP 8.1 if the resource was already closed (or is otherwise invalid).

- Added if condition to fclose when the resource is available. 
- This is for v1.x releases.

Fix in v2.x - https://github.com/predis/predis/pull/1199


tillkruss:
v1.x is no longer maintained.

Issue #595, opened 2020-01-28
Cluster Mode - No READONLY commands on Slave Nodes - All commands forwarded to Master Node


dntosas:
**Expected behaviour**
Predis should evenly distribute readonly commands between masters and slaves

**Actual behaviour**
Predis forwards readonly cmd to SLAVE but SLAVE responses with MOVED to MASTER node as Predis doesnt instruct the SLAVE that wants to get readonly information. As a result, all commands are executed on MASTER and that break the cluster offerings for scaling and traffic distribution.

**Why that happens**
>There's one important caveat: if a client connects to a replica and sends the command GET foo, the reply will be a MOVED error message even if that replica is in charge of the hash slot where the key "foo" is located. That behavior can be changed by sending first the command READONLY, which instruct the replica to serve the client's request even if the data is stale. The READWRITE command can be used to restore the connection to the original state.

dntosas:
@nrk hello man and thank you for your work

Do you have any clue on this?

Issue #579, opened 2019-09-02
Need an example of how to create a predis client with clustering AND replication


abbood:
The docs provides examples with either or, but not both

Basically I'm creating a predis client with the following parameters:

```

$parameters = [
                               // masters
                               'localhost:7000?alias=master', 
                               'localhost:7001?alias=master', 
                               'localhost:7002?alias=master', 

                               // slaves
                               'localhost:7003', 
                               'localhost:7004', 
                               'localhost:7005', 
                          ]

$options    = ['cluster' => 'redis', 'replication' => true];
           
$client = new Predis\Client($parameters, $options);
```

But when it [creates the connection](https://github.com/nrk/predis/blob/v1.1/src/Client.php#L103-L138) in predis/createConnection, i get this in the debugger:

```
- Locals at /Users/Shared/dev/php/toters-api/vendor/predis/predis/src/Client.php:138

 ▾ $connection = (Predis\Connection\Aggregate\MasterSlaveReplication [6])
  \
   ▸ $connection->strategy = (Predis\Replication\ReplicationStrategy [3])
   |
   ⬦ $connection->master = (null)
   |
   ▾ $connection->slaves = (array [6])
    \
     ▸ $connection->slaves['slave-localhost?alias=master:7000'] = (Predis\Connection\StreamConnection [4])
     |
     ▸ $connection->slaves['slave-localhost?alias=master:7001'] = (Predis\Connection\StreamConnection [4])
     |
     ▸ $connection->slaves['slave-localhost?alias=master:7002'] = (Predis\Connection\StreamConnection [4])
     |
     ▸ $connection->slaves['slave-localhost:7003'] = (Predis\Connection\StreamConnection [4])
     |
     ▸ $connection->slaves['slave-localhost:7004'] = (Predis\Connection\StreamConnection [4])
     |
     ▸ $connection->slaves['slave-localhost:7005'] = (Predis\Connection\StreamConnection [4])
    /
   ⬦ $connection->current = (null)
   |
   ⬦ $connection->autoDiscovery = (bool) 0
   |
   ⬦ $connection->connectionFactory = (null)
  /
 ⬦ $initializer = (uninitialized)
 |
 ▾ $options = (Predis\Configuration\Options [3])
  \
   ▸ $options->input = (array [1])
   |
   ▸ $options->options = (array [2])
   |
   ▸ $options->handlers = (array [6])
  /
 ▾ $parameters = (array [6])
  \
   ▸ $parameters[0] = (array [4])
[...]

Issue #392, opened 2016-11-15
Redis Cluster with multiple master and replica nodes


Abam:
As described in [the official redis documentation Redis Cluster master-slave model](http://redis.io/topics/cluster-tutorial), it is possible to have a cluster with master-slave model.

> so that the final cluster is composed of A, B, C that are masters nodes, and A1, B1, C1 that are slaves nodes, the system is able to continue if node B fails.

But it seems that is not supported by predis configuration. We have to choose between either a cluster model or a master slave model with only one master. The code in [Client.php](https://github.com/nrk/predis/blob/v1.1/src/Client.php#L123) seems to confirm it (if replication is choosen, the else case dealing with cluster is never reached).

Any plan on supporting this configuration ?

Edit: After having a deep dive into https://github.com/nrk/predis/blob/v1.1/src/Connection/Aggregate/PredisCluster.php it seems that if useClusterSlots is true, the client will be able to update slots and therefore use slaves if master fail. Isn't it ?

nrk:
This is indeed not yet supported by Predis. Slaves are ignored by the client when it [fetches the slot map](https://github.com/nrk/predis/blob/v1.1.1/src/Connection/Aggregate/RedisCluster.php#L260-L272) from one of the Redis nodes, furthermore it is not possibile to specify them manually via the client's configuration (but this may change with Predis 2.0 thanks to some internal but breaking changes to support a `role` parameter).

When `useClusterSlots` is set to `TRUE` (which is the default behaviour) this is a actually a minor issue since redis-cluster takes care of electing a new master from the pool of slaves connected to a dead master, so this feature would useful mostly to offload master nodes by redirecting read operations to their slaves (like it happens when using replication).

Abam:
Ok, thanks for the reply. So I can rely on the userClusterSlots parameter for predis v1.1.

nrk:
@Abam my plans anyway are to make Predis aware of slaves in a redis-cluster configuration in the future, so I'll just leave this issue open as a reminder. Thanks.

binwiederhier:
Any movement on this? Anything that can be done to help with this?

binwiederhier:
I briefly looked into implementing cluster support for your library, but it seems to be non-trivial (at least to me). Thanks for all the work on it. It works wonderfully in a non-clustered environment. 

[...]

Issue #173, opened 2014-05-28
Support for both cluster and replication


vasyap:
Current Redis Cluster's documentation suggests to create replicas for every cluster node, and it seems to be a good failure-proof idea. Predis, however, could not act this way and refuses to use 'cluster' option when used with master/slave replicasets.

It looks to be a little complicated way to implement this, though this issue is probably linked with Sentinel support or anything else I could not figure out yet.

PR #1411, opened 2023-11-10
Use Replicas For Read

Cardmarket-Sijing:
This is a proposition on how load could be balanced for replicated clusters.

See:
#595 #392 #173 #579

* Added separate pool and slot map for cluster replicas
* Use replicas to serve read requests

This adds a dependency on the replication strategy for detecting read commands. The detection could be moved to the redis cluster strategy or to a separate class included in both strategies.

Also this doesn't allow for the user to configure how the replicas should be used. Future improvements could be:
* prefer replicas for reads, fallback to primary
* randomly select between replicas and primaries to evenly distribute

This splits up the slots of the primary between replicas. Another idea would be to share the slots between all replicas and select the replica (randomly) for each command

### DIFF
--- a/src/Connection/Cluster/RedisCluster.php
+++ b/src/Connection/Cluster/RedisCluster.php
@@ -74,14 +74,15 @@ class RedisCluster implements ClusterInterface, IteratorAggregate, Countable
      */
     public function __construct(
         FactoryInterface $connections,
-        StrategyInterface $strategy = null
+        StrategyInterface $strategy = null,
+        ReplicationStrategy $replicationStrategy = null
     ) {
         $this->connections = $connections;
         $this->strategy = $strategy ?: new RedisClusterStrategy();
         $this->slotmap = new SlotMap();
 
         $this->replicaSlotmap = new SlotMap();
-        $this->replicaStrategy = new ReplicationStrategy();
+        $this->replicaStrategy = $replicationStrategy ?: (new ReplicationStrategy())->disableLoadBalancing();
     }
 
     /**
@@ -296,7 +297,6 @@ class RedisCluster implements ClusterInterface, IteratorAggregate, Countable
 
         foreach ($response as $slots) {
             [$start, $end, $master] = $slots;
-            $replicas = array_slice($slots, 3);
 
             if ($master[0] === '') {
                 $this->slotmap->setSlots($start, $end, (string) $connection);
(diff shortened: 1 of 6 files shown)


src/Connection/Cluster/RedisCluster.php
```diff
@@ -75,6 +79,9 @@ public function __construct(
         $this->connections = $connections;
         $this->strategy = $strategy ?: new RedisClusterStrategy();
         $this->slotmap = new SlotMap();
+
+        $this->replicaSlotmap = new SlotMap();
```
vladvildanov, line 83: Why should we use another SlotMap if we have a SlotMap above?

vladvildanov:
@tillkruss @Cardmarket-Sijing Looks good, but I have few suggestions:

1. Make it configurable. Load balancing should be configurable from client configuration. I suggest to add a support for new option that enables it. `ReplicationStrategy::loadBalancing` should help you:

```
$options = ['cluster' => 'redis, 'loadBalancing' => true];
```

2. Test coverage. We need unit tests and feature tests coverage against cluster.
3. Documentation. Load balancing feature should be documented

vladvildanov:
@tillkruss And since this is a breaking change feature we should include it in 3.x

src/Connection/Cluster/RedisCluster.php
Cardmarket-Sijing: The current slot map only tracks primaries and does not seem to have functionality for tracking lists of connections per slot. 
I did not want to introduce too many changes at once

Cardmarket-Sijing:
> @tillkruss And since this is a breaking change feature we should include it in 3.x

It could probably be changed to a non breaking change by making it configurable as suggested above and disabled by default.

tillkruss:
> It could probably be changed to a non breaking change by making it configurable as suggested above and disabled by default.

Yeah, I'll leave that decision up to you. Configuration would be nice.

Cardmarket-Sijing:
I've changed it to be configurable through 

`$options = ['cluster' => 'redis, 'loadBalancing' => true];`

And updated the readme accordingly.

Also added some coverage tests. Let me know if there should be additional test cases

tests/Predis/Connection/Cluster/RedisClusterTest.php
```diff
@@ -1455,4 +1458,202 @@ public function testQueryClusterNodeForSlotMapPauseDurationOnRetry()
 
         $this->assertCount(16384, $cluster->getSlotMap());
     }
+
```
vladvildanov, line 1461: Mark tests, with `@group disconnected`, represents unit tests

vladvildanov:
@Cardmarket-Sijing Looks great! The only missing piece is an integration tests against cluster. Here's an instruction how to perform them locally:

1. Go to directory `docker/unstable_cluster` (make sure you have docker and docker-compose installed) and run `docker-compose up -d`.
2. Write integration tests within the same test class and mark them with annotations: `@group connected`, `@group cluster`
3. Now you can run them and they will be executed against cluster

Cardmarket-Sijing:
Something broke while trying to rebase to the main branch, so I made another PR targeting the 2.x branch again.
#1414

Issue #1292, opened 2023-05-09
Improve Relay integration


tillkruss:
- [ ] Add `dispatcher_loop.php` example?
- [ ] Auto-mutate data
- [ ] Handle `Predis\Command\RawCommand()` calls?
- [x] Review `relay-fixme` test group
- [x] Fix `RelayConnection` hierarchy
- [x] Make it possible to use `RelayConnection` within aggregate connections (OSS cluster support) #1423
- [x] Test coverage (included cluster tests) #1424
- [x] RESP2 tests
Follow-up to #1263.

tillkruss:
> Hey! I just had a closer look on Relay support and notice that it's really conceptually wrong.
RelayConnection class isn't really represents new connection interface (I assume it's still socket under the hood of RelayClient used for communication, so it's the same as StreamConnection). It looks like the architecture wise it should be a separate client object RelayClientAdapter that will adapt RelayClient to ClientInterface. Client should send commands over connection, but not vice versa.
For example, right now it's impossible to use Relay against cluster and to be able to support it you will need new object smth like RelayCluster and it will lead to incremental changes and workarounds (edited) 

>Hey, Till! I just faced with another problem of bad Relay implementation, since you're using another client within RelayConnection (which means that we're using client that uses another client), feature tests will fail for some commands, because implementations can differ in both clients.
For example, I implemented SSUBSCRIBE  command and my feature tests works as expected, but then when the same feature tests called with RelayConnection it fails because it calls via RelayClient where this command implemented different and it requires at least 2 arguments to be passed. Mixing clients leads to unexpected problems more and more

tillkruss:
@vladvildanov: How long do you reckon this would take you to change? 👆

vladvildanov:
@tillkruss I need to make another brainstorm on this to choose the best approach for the use case, since it's a huge change.

vladvildanov:
@tillkruss @chayim  The list of mandatory improvements from my POV:

1. `RelayConnection` hierarchy.
For communication with Redis via Relay, we're using `RelayClient` which is a "black-box" for us, but `StreamConnection` represents a connection via streams and sockets that we have control on and it's a "white-box". It's conceptually, wrong to consider `StreamConnection` as ancestor of `RelayConnection`.

If it's a black-box, let's treat it as a black-box. `RelayConnection` object should be an adapter and implements `NodeConnectionInterface`. Decorators and proxies should be used in future in case if more business logic should be added.

[...]

PR #1419, opened 2023-11-22
Fixed Relay objects structure

vladvildanov:
Enhancements:

1. `RelayConnection` now is implementation of `NodeConnectionInterface`, not a children of `StreamConnection`.
2. Added `RelayFactory`, separate object that responsible for `RelayConnection` instantiating. Single factory for all connection types makes it's harder to maintain and requires additional `if` statements per type if logic is differ, which leads to changes into existing code.
3. Test coverage increased, including feature tests.

#1292


tillkruss (approved):

tillkruss:
@vladvildanov Should we merge this into v3 directly?

tillkruss:
> @tillkruss If you don't mind, I created a separate branch for whole relay improvement "feature", my plan is to push it into 3.x when it's will be completely finished

Sure thing 👍

PR #1421, opened 2023-12-20
Added OSS cluster tests

vladvildanov:
Tests to prove that Relay connection can be used within cluster(aggregate) connection

### DIFF
--- a/src/Connection/RelayConnection.php
+++ b/src/Connection/RelayConnection.php
@@ -125,14 +125,6 @@ class RelayConnection extends AbstractConnection
     }
 
     /**
-     * {@inheritdoc}
-     */
-    protected function getIdentifier()
-    {
-        return $this->client->endpointId();
-    }
-
-    /**
      * @param  ParametersInterface $parameters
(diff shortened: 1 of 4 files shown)


vladvildanov:
@tillkruss Do you think we need more cluster test cases?

src/Connection/RelayConnection.php
```diff
@@ -124,14 +124,6 @@ public function getClient()
         return $this->client;
     }
 
-    /**
-     * {@inheritdoc}
-     */
-    protected function getIdentifier()
-    {
-        return $this->client->endpointId();
```
vladvildanov, line 132: In the context of connection this method doesn't need at all, cause it's just used to get a unique identifier for connection used on the client side. Here in RelayConnection this method requires connection to server, before actual connection is established.

tillkruss:
> @tillkruss Do you think we need more cluster test cases?

No, I think having minimal tests for Cluster is fine, it's an edge-case IMO.

tillkruss (approved):

PR #1423, opened 2023-12-26
Improved Relay integration

### LINKED ISSUE #1292: Improve Relay integration

vladvildanov:
See #1292.


vladvildanov:
@tillkruss Lmk if there's anything else we need to improve

src/Connection/RelayFactory.php
```diff
@@ -0,0 +1,147 @@
...
+
+class RelayFactory extends Factory
+{
+    /**
+     * @var string[]
+     */
+    protected $schemes = [
+        'tcp' => RelayConnection::class,
+        'unix' => RelayConnection::class,
+        'redis' => RelayConnection::class,
```
tillkruss, line 28: This is missing TLS and `rediss`, right?

src/Connection/RelayFactory.php
```diff
@@ -0,0 +1,147 @@
...
+                : [$parameters->password];
+
+            $connection->addConnectCommand(
+                new RawCommand('AUTH', $cmdAuthArgs)
+            );
+        }
+
+        if (isset($parameters->database) && strlen($parameters->database)) {
+            $connection->addConnectCommand(
+                new RawCommand('SELECT', [$parameters->database])
```
tillkruss, line 143: When using Relay `auth()` and `select()` are special methods that should be executed, without using a raw command. 

Correct me if I'm wrong @michael-grunder.

tillkruss (changes requested):

tillkruss:
@vladvildanov This needs a rebase when you have a moment.

src/Connection/RelayFactory.php
michael-grunder: I think `select` can be used via `rawCommand` but `auth` should not be really.

tillkruss: Ah. I thought it switches in the internal database for caching or something similar.

vladvildanov: @tillkruss But this commands will be executed as `auth()` and `select()`, because `commandId` matching `$atypicalCommands` values.

https://github.com/predis/predis/blob/feature/relay-improvements/src/Connection/RelayConnection.php#L70
https://github.com/predis/predis/blob/feature/relay-improvements/src/Connection/RelayConnection.php#L176

src/Connection/RelayFactory.php
vladvildanov: Yeah, you're right. But they should be handled by `StreamConnection`, forgot to add it

tillkruss, line 28:
```suggestion
        'tcp' => RelayConnection::class,
        'tls' => RelayConnection::class,
        'unix' => RelayConnection::class,
        'redis' => RelayConnection::class,
        'rediss' => RelayConnection::class,
```

src/Connection/RelayFactory.php
tillkruss: Right, glad you confirmed.

tillkruss (approved):

PR #1428, opened 2024-01-10
Removed NULL assignement for username parameter

vladvildanov:
Since, Redis 6.2 sentinels supports ACL authentication and AUTH command with both username and password. So for now we don't need to override `username` on Sentinel connections

### DIFF
--- a/src/Connection/Replication/SentinelReplication.php
+++ b/src/Connection/Replication/SentinelReplication.php
@@ -264,14 +264,10 @@ class SentinelReplication implements ReplicationInterface
         }
 
         if (is_array($parameters)) {
-            // NOTE: sentinels do not accept AUTH and SELECT commands so we must
-            // explicitly set them to NULL to avoid problems when using default
-            // parameters set via client options. Actually AUTH is supported for
-            // sentinels starting with Redis 5 but we have to differentiate from
-            // sentinels passwords and nodes passwords, this will be implemented
-            // in a later release.
+            // NOTE: sentinels do not accept SELECT command so we must
+            // explicitly set it to NULL to avoid problems when using default
+            // parameters set via client options.
             $parameters['database'] = null;
-            $parameters['username'] = null;
 
             // don't leak password from between configurations
             // https://github.com/predis/predis/pull/807/#discussion_r985764770


src/Connection/Replication/SentinelReplication.php
```diff
@@ -264,14 +264,10 @@ protected function createSentinelConnection($parameters)
...
-            // explicitly set them to NULL to avoid problems when using default
-            // parameters set via client options. Actually AUTH is supported for
-            // sentinels starting with Redis 5 but we have to differentiate from
-            // sentinels passwords and nodes passwords, this will be implemented
-            // in a later release.
+            // NOTE: sentinels do not accept SELECT command so we must
+            // explicitly set it to NULL to avoid problems when using default
+            // parameters set via client options.
             $parameters['database'] = null;
-            $parameters['username'] = null;
```
tillkruss, line 274: I recall these changes causing issues in the past, but I can only find #807.

tests/Predis/Connection/Replication/SentinelReplicationTest.php
```diff
@@ -51,17 +51,17 @@ public function testParametersForSentinelConnectionShouldUsePasswordForAuthentic
     /**
      * @group disconnected
      */
-    public function testParametersForSentinelConnectionShouldIgnoreDatabaseAndPassword(): void
+    public function testParametersForSentinelConnectionShouldIgnoreDatabase(): void
```
tillkruss, line 54: I think we should merge this into `main` instead, unless we can keep the existing unit tests unchanged.

tillkruss (changes requested):

tests/Predis/Connection/Replication/SentinelReplicationTest.php
vladvildanov: Well, functionality was changed and unit test should be changed according to this. Or you mean that the whole PR should be merged into main?

tillkruss: What I mean is that people might rely on the removed/changed functionality and if we want to change behavior, we have to either merge into `main` because it's breaking.

I'm usually being overly precautious and assume people have weird Sentinel setups with varying usersnames for each node, or some other oddities.

If you like we can merge this into `2.x`, but we might crash someone's setup and need to revert.

vladvildanov: You're right, it's a breaking change. Here's new PR against main #1429

PR #1438, opened 2024-03-04
Improved pipeline abstractions

vladvildanov:
The goal of this PR is to enhance pipelines to make them "real" pipelines. For now all commands are still pushed into socket one by one. Instead of this, the behaviour was changed to create a buffer from pipelined commands and push them all at once.

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -478,7 +478,7 @@ class Client implements ClientInterface, IteratorAggregate
      *
      * @return Pipeline|array
      */
-    protected function createPipeline(array $options = null, $callable = null)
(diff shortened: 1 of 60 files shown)


tillkruss (commented):
This is excellent. Does it have any implication with Relay, since `write()` isn't supported?

vladvildanov:
@tillkruss I don't think it has, Relay is an another layer above connection that does not provide an API for direct socket communication. Also, RelayPipeline overrides `executePipeline()` method, so the parent class changes does not affect it

tillkruss (approved):

PR #1432, opened 2024-02-09
[2.x] ZPOPMIN & ZPOPMAX prefixes

sngrl:

PR #1443, opened 2024-03-19
Define prefix strategy for ZMSCORE command

mohorev:
**Description:**

When using Predis, if a client prefix is set and then a `ZMSCORE` command is executed, the client prefix is ignored. This leads to incorrect results, as the command operates on keys without the applied prefix.

**Steps to Reproduce:**

- Set a prefix for the Predis client.
- Use the `ZMSCORE` command to retrieve the score from a sorted set.

**Expected Behavior:**

The `ZMSCORE` command should consider the client prefix when retrieving the score from the sorted set, ensuring that the correct key is used.

**Actual Behavior:**

The `ZMSCORE` command retrieves the score from the sorted set without considering the client prefix. As a result, incorrect keys are accessed, leading to unexpected behavior.

**Additional Information:**

Predis Version: v2.2.2

Similar issue with prefix problems in `ZPOPMIN` & `ZPOPMAX` commands https://github.com/predis/predis/pull/1432 already reported.

**Sample Code:**

```php
<?php

require __DIR__ . '/vendor/autoload.php';

// Set up Predis client with a prefix
$redis = new Predis\Client('tcp://localhost:6383', ['prefix' => 'test_']);

// Add values to a sorted set
$redis->zadd('mykey', ['foo' => 10, 'bar' => 20]);

// ZRANGE works correct, return: ['foo' => '10', 'bar' => '20']
var_dump($redis->zrange('mykey', 0, -1, ['WITHSCORES' => true]));

// ZMSCORE ignores the prefix. Outputs incorrect scores: [null, null]
var_dump($redis->zmscore('mykey', 'foo', 'bar'));
```

### DIFF
--- a/examples/transaction_using_cas.php
+++ b/examples/transaction_using_cas.php
@@ -32,7 +32,7 @@ function zpop($client, $key)
         'cas' => true,      // Initialize with support for CAS operations
         'watch' => $key,    // Key that needs to be WATCHed to detect changes
         'retry' => 3,       // Number of retries on aborted transactions, after
-        // which the client bails out with an exception.
+                            // which the client bails out with an exception.
     ];
 
     $client->transaction($options, function ($tx) use ($key, &$element) {
--- a/examples/transaction_using_cas.php
+++ b/examples/transaction_using_cas.php
@@ -32,7 +32,7 @@ function zpop($client, $key)
         'cas' => true,      // Initialize with support for CAS operations
         'watch' => $key,    // Key that needs to be WATCHed to detect changes
         'retry' => 3,       // Number of retries on aborted transactions, after
-                            // which the client bails out with an exception.
+        // which the client bails out with an exception.
     ];
 
     $client->transaction($options, function ($tx) use ($key, &$element) {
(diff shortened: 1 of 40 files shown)


mohorev:
@vladvildanov @tillkruss Hey team, can we release with these changes? This code is my salvation in production. I'd be praying for you in church if you help speed up this release 😺

mohorev:
To kill two birds with one stone, I decided to go further, and within the scope of this pull request, I also fixed the prefix issue with the ZPOPMAX and ZPOPMIN commands, which was mentioned in a [neighboring pull request](https://github.com/predis/predis/pull/1432). However, unlike that solution, I added unit tests here that verify the functionality of the prefixes.

tillkruss (changes requested):
Please undo all the styling changes.

mohorev:
@tillkruss Done. As per your request, I've reverted the commit regarding the styling formatting.

vladvildanov:
@mohorev Thanks for your input! I added a PR to fix codestyle issues and bind a version of php-cs-fixer package. After it will be merged please rebase your branch and we will release both of this changes

https://github.com/predis/predis/pull/1445

tillkruss:
I've merged this in #1451 without the test changes.

PR #1445, opened 2024-03-21
Limit version of php-cs-fixer >= 3.4.0 <= 3.5.0

vladvildanov:
Php-cs-fixer package continue constantly updating, so are codestyle rules. This leads often to CI failures and requires separate PR so it won't be mixed with actual changes. I propose to limit to the latest minor version, to avoid this codestyle changes and lately we have to choose more stable version.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,7 @@
         "php": "^7.2 || ^8.0"
     },
     "require-dev": {
-        "friendsofphp/php-cs-fixer": "^3.3",
+        "friendsofphp/php-cs-fixer": "~3.4.0",
         "phpstan/phpstan": "^1.9",
         "phpunit/phpunit": "^8.0 || ^9.4"
     },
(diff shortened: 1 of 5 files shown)


tests/Predis/Command/Redis/GEOSEARCH_Test.php
```diff
@@ -186,7 +186,7 @@ public function responsesProvider(): array
                 [
                     'member1' => ['lng' => 1.1, 'lat' => 2.2],
                     'member2' => ['lng' => 2.2, 'lat' => 3.3],
-                    'member3' => ['lng' => 3.3, 'lat' => 4.4]],
+                    'member3' => ['lng' => 3.3, 'lat' => 4.4], ],
```
mohorev, line 189: My hands are itching to move this bracket to a new line:

```suggestion
                    'member3' => ['lng' => 3.3, 'lat' => 4.4],
                ],
```

composer.json
```diff
@@ -25,7 +25,7 @@
         "php": "^7.2 || ^8.0"
     },
     "require-dev": {
-        "friendsofphp/php-cs-fixer": "^3.3",
+        "friendsofphp/php-cs-fixer": "~3.4.0",
```
tillkruss, line 28: I don't think we should lock the version of `php-cs-fixer`, maybe we just adjust the rules and code as they update?

PR #1448, opened 2024-03-27
explicitly mark nullable parameters as nullable

xabbuh:
not doing this will be deprecated in PHP 8.4 (see e.g. https://github.com/symfony/symfony/actions/runs/8459191951/job/23175036483?pr=54426#step:8:2335)

PR #1457, opened 2024-05-02
Explicitly mark nullable parameters as nullable

nicolas-grekas:
Same as #1448 but for branch v1.x :pray:


nicolas-grekas:
:facepalm: not possible because this is not valid syntax in older php versions of course.

PR #1468, opened 2024-06-28
Fix prefix for SMISMEMBER

mohorev:
**Description:**

When using the `SMISMEMBER` command, it has been observed that the client's prefix is ignored. This leads to incorrect membership checks within a set, as the key passed to the command does not include the prefix specified for the client.

**Changes:**

- Updated the handling of the `SMISMEMBER` command to include the client's prefix in the key.
- Added tests to verify correct behavior of the command with a prefix.

**Additional Information:**

Predis Version: v2.2.2


tillkruss (approved):

tillkruss:
@vladvildanov Good?

mohorev:
@tillkruss @vladvildanov I rebased my branch from v2.x because code style issues are already fixed in the base branch.

PR #1471, opened 2024-07-09
keep key and value when filtering SET arguments

xabbuh:
With the changes done in #1470 the value that is going to be set is also filtered which prevents setting a key to `false` or `null`.


vladvildanov (approved):

vladvildanov:
@xabbuh Thank you, it makes a lot of sense

tillkruss (approved):

PR #1476, opened 2024-08-29
ClientInterface: update watch command accepting string and string[]

glaubinix:
`watch` accepts one or more strings as argument https://redis.io/docs/latest/commands/watch/


tillkruss (changes requested):
Needs tests.

glaubinix:
There already was `WATCH_Test::testFilterArgumentsAsSingleArray` which asserts that passing a list gets converted to the correct structure. Added a test that shows watching works both via `$redis1->watch('foo', 'bar')` and `$redis1->watch(['foo', 'bar'])`

tillkruss (approved):

PR #1479, opened 2024-09-27
Check if countable

streamingsystems:
Every once in a while I get weird errors about the value not being countable and putting this code solved.


tillkruss (changes requested):
Please remove dot files.

streamingsystems:
Hi,

I think I removed them correctly.

PR #1482, opened 2024-10-14
Add TOPK.COUNT command

acb:
I'm starting to use TOPK for some work in our application and noticed that TOPK.COUNT was missing. It'll be handy for us to check some thresholds so I figured I could add it.

If this could be baked into a 2.x release soonish I would love it, otherwise we'll have to use my fork for a while.


tillkruss (changes requested):
Needs tests.

tillkruss:
Closing due to lack of tests. @vladvildanov

PR #1484, opened 2024-10-21
Added support for transactions in OSS Cluster

vladvildanov:
This PR adds a support for transaction execution against distributed hash-slot environment. 

The core idea is to allow a transaction execution against same hash slot to keep data atomicity. In Cluster environments the commands are queued until it's possible to calculate hash slot and ensure that commands are operated against a same slot


vladvildanov:
@tillkruss Why does it marked as breaking change?

vladvildanov:
@tillkruss Blocked by #1485

dmaier-redislabs (dismissed):
Looks good from my point of view. @tillkruss It seems that an approval by you is needed. Could you please review and approve?

tillkruss:
> Why does it marked as breaking change?

@vladvildanov: Some interfaces changed and new return types for methods. That would break if someone has their own implementation.

vladvildanov:
@tillkruss Well, the interfaces was extended with new methods and there's no changes to existing one. The return types wasn't changed they were defined in the places where the definition is missing, also changes to functions and constructor signatures is backward compatible. So I disagree that this can be consider as breaking.

Let me know if you meant something specific in this PR.

tillkruss:
> Well, the interfaces was extended with new methods and there's no changes to existing one.

Adding a new method to an interface is a breaking change.

If you want to dig up some official SemVer rules on this we can nitpick it, maybe PHP-FIG or a similar group has some?

vladvildanov:
@tillkruss Hmm, PHP-FIG doesn't elaborate on semantic versioning, so I guess we can jump into official docs from https://semver.org/.
```
Given a version number MAJOR.MINOR.PATCH, increment the:

1. MAJOR version when you make incompatible API changes
2. MINOR version when you add functionality in a backward compatible manner
3. PATCH version when you make backward compatible bug fixes
```

Well, extending the existing interface with a new method can be consider as a `add functionality in a backward compatible manner`, it doesn't break any existing behaviour and doesn't impact the end user. The same could be applied for any function signature that was extended with additional optional arguments, if arguments order was kept.

tillkruss:
>Well, extending the existing interface with a new method can be consider as add functionality in a backward compatible manner

Not in my books. IMO it's a breaking change.

- https://github.com/phpredis/phpredis/issues/2562
- https://github.com/phpredis/phpredis/issues/2564

```php
<?php

interface StrategyInterface {
  public function getSlot(string $command);
  public function checkSameSlotForKeys(array $keys): bool;
}

class StrategyTest implements StrategyInterface {
  public function getSlot(string $command) {
    // Your implementation here
  }
}

$test = new StrategyTest();
```

vladvildanov:
@tillkruss Okay, it's up to you. Appreciate your review in any case

tillkruss (changes requested):
Looks good, just need to change the branch to v3.x

PR #1486, opened 2024-10-28
Read operations from a random Redis Cluster replica

disc:


disc:
Hello @tillkruss and @vladvildanov. I've just slightly improved PR by supporting a few modes for scaling read operations in redis-cluster. In this version, read operations can be scaled between master and replica nodes in a `random` mode, or just by replicas in `replicas` mode.

disc:
Hello @tillkruss. Could you look into my changes and share your opinion about the feature?

tillkruss:
Looks like not breaking changes, which is great. I'd like @vladvildanov to review this and give his thumbs up.

disc:
> Looks like not breaking changes, which is great. I'd like @vladvildanov to review this and give his thumbs up.

Hello @vladvildanov. Have you had a chance to look into the suggested changes?

vladvildanov:
@disc Sorry for late response guys, I'm on it

vladvildanov:
@disc @tillkruss 

I see some major issues with this PR that will only worsen an existing problems in a client:

1. Current cluster implementation are not designed to support master/slave replication

Adding new connection manager (e.g ReadConnectionSelector class) on top of existing RedisCluster, just to manage replica connections is a workaround that ignores the general issue that master/slave replication isn't supported. Load balancing is one of the features that comes up with replication, but the problem should be scoped accordingly - "we need to add a support for master/slave replication in Cluster API". And the access to master/replica nodes should be managed by Cluster object.

https://github.com/predis/predis/blob/v2.x/src/Connection/Cluster/RedisCluster.php#L287

2. **ReadonlyOperationDetector class**

I disagree that we need an external class that should distinguish read and write commands. The information about the command mode should be exposed by command object itself, so the `CommandInterface` should be extended with something like `getMode()` method. In other case we need to change a code in different places when we add a new command, we already have the same detector for Replication:

https://github.com/predis/predis/blob/v2.x/src/Replication/ReplicationStrategy.php#L213

Apart of it we have to specify separately if command is supported by cluster:

https://github.com/predis/predis/blob/v2.x/src/Cluster/ClusterStrategy.php#L41

If we will add another detector, each new command that will be added needs to be appended to each of this detectors in 3 different places. So my point that we shouldn't make it even worse.

3. **ReadConnectionSelector class**

As I mentioned above, a separate class that will manage connections seems like workaround, connections are already managed by RedisCluster and new object should not share this responsibilities, also it will be aligned with Replication logic.

https://github.com/predis/predis/blob/v2.x/src/Connection/Replication/MasterSlaveReplication.php#L166

PR #1495, opened 2024-12-05
Check if data is countable

streamingsystems:
Hi,

Every once in a while we see this in our logs where Redis seems to be returning data that is not countable. 

If I check if the data is_countable prior prior to the loop it fixes it. 

[2024-12-05 15:23:03] worker.ERROR: count(): Argument #1 ($value) must be of type Countable|array, int given {"exception":"[object] (TypeError(code: 0): count(): Argument #1 ($value) must be of type Countable|array, int given at /home/ss/application/vendor/predis/predis/src/Command/Redis/ZRANGE.php:96)
[stacktrace]
#0 /home/ss/application/vendor/predis/predis/src/Command/Redis/ZRANGE.php(96): count(0)
#1 /home/ss/application/vendor/predis/predis/src/Client.php(391): Predis\\Command\\Redis\\ZRANGE->parseResponse(0)

### DIFF
--- a/src/Command/Redis/ZRANGE.php
+++ b/src/Command/Redis/ZRANGE.php
@@ -92,12 +92,13 @@ class ZRANGE extends RedisCommand
     {
         if ($this->withScores()) {
             $result = [];
-
-            for ($i = 0; $i < count($data); ++$i) {
-                if (is_array($data[$i])) {
-                    $result[$data[$i][0]] = $data[$i][1]; // Relay
-                } else {
-                    $result[$data[$i]] = $data[++$i];
+            if (is_countable($data)) {
+                for ($i = 0; $i < count($data); ++$i) {
+                    if (is_array($data[$i])) {
+                        $result[$data[$i][0]] = $data[$i][1]; // Relay
+                    } else {
+                        $result[$data[$i]] = $data[++$i];
+                    }
                 }
             }
 


tillkruss:
@vladvildanov this seems like an issue with not detecting the mode in 6.2 doesn't it?

@streamingsystems What Redis version are you using?

streamingsystems:
Hi,

From regis-cli I run info: 

# Server
redis_version:7.2.4

The strange this is that we have many thousands of requests this go through our redis system and randomly will get that in the logs. 

If it matters we are running sentinel and master / replica setup.

streamingsystems:
I saw this again again this morning: 

[2024-12-06 07:53:04] worker.CRITICAL: count(): Argument #1 ($value) must be of type Countable|array, int given  
[2024-12-06 07:53:04] worker.ERROR: count(): Argument #1 ($value) must be of type Countable|array, int given {"exception":"[object] (TypeError(code: 0): count(): Argument #1 ($value) must be of type Counta
ble|array, int given at /home/ss/application/vendor/predis/predis/src/Command/Redis/ZRANGE.php:96)
[stacktrace]
#0 /home/ss/application/vendor/predis/predis/src/Command/Redis/ZRANGE.php(96): count(0)


Also, instead of is_countable($data) you can just put if ($data) {}, around the for loop - anything to stop $data which is an integer from being passed into count().

vladvildanov:
@tillkruss @streamingsystems Hmm, that's weird because response schema for ZRANGE command is an array, so you should never get integer. Could you confirm you get a reply a different from what is specified in documentation?

https://redis.io/docs/latest/commands/zrange/

PR #1497, opened 2025-01-13
Added support for transactions in OSS Cluster

vladvildanov:
This PR adds a support for transaction execution against distributed hash-slot environment.

The core idea is to allow a transaction execution against same hash slot to keep data atomicity. In Cluster environments the commands are queued until it's possible to calculate hash slot and ensure that commands are operated against a same slot


vladvildanov:
@tillkruss May I have a review for this one?

vladvildanov:
@tillkruss Done, also I added an entry in README

PR #1500, opened 2025-01-14
Update README.md to include that the client is compatible with Valkey

asafpamzn:
Added documentation clarifying that predis is compatible with Valkey. This information could be really helpful for developers using Valkey in their projects, saving them time and ensuring they are confident in their choice of predis.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@
 [![Build status][ico-build]][link-actions]
 [![Coverage Status][ico-coverage]][link-coverage]
 
-A flexible and feature-complete [Redis](http://redis.io) and [Valkey](https://github.com/valkey-io/valkey) client for PHP 7.2 and newer.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -7,7 +7,7 @@
 [![Build status][ico-build]][link-actions]
 [![Coverage Status][ico-coverage]][link-coverage]
 
-A flexible and feature-complete [Redis](http://redis.io) client for PHP 7.2 and newer.
+A flexible and feature-complete [Redis](http://redis.io) / [Valkey](https://github.com/valkey-io/valkey) client for PHP 7.2 and newer.
```
tillkruss, line 10: Let's update `composer.json` as well.

tillkruss (changes requested):

README.md
asafpamzn: ok

asafpamzn: Done

PR #1508, opened 2025-02-11
CommandFactory: reject named arguments as those are not supported

janedbal:
Since predis [2.3.0](https://github.com/predis/predis/releases/tag/v2.3.0), named arguments are not supported (at least in `SET` command) due to [this change](https://github.com/predis/predis/pull/1471/files) as the `value` argument does not meet [the condition there](https://github.com/predis/predis/pull/1471/files#diff-786e3de4bbf9be306c2c5a6aba6933c6fc3fb7aa1de753be55f5b8a4cb519858R33).

At first, I was about to fix this by supporting it inside `SET::setArguments`, but as all the command calls are implemented via `__call`, I believe it is safer to rely on the fact that your users are not using named arguments at all.

### DIFF
--- a/src/Command/Factory.php
+++ b/src/Command/Factory.php
@@ -67,6 +67,18 @@ abstract class Factory implements FactoryInterface
             throw new ClientException("Command `$commandID` is not a registered Redis command.");
         }
 
+        $containsNamedArgument = array_reduce(
+            array_keys($arguments),
+            static function (bool $carry, $key): bool {
+                return $carry || is_string($key);
+            },
+            false
+        );
+
+        if ($containsNamedArgument) {
+            throw new ClientException('Named arguments are not supported.');
+        }
+
         $command = new $commandClass();
         $command->setArguments($arguments);
 
--- a/tests/Predis/Command/Redis/SET_Test.php
+++ b/tests/Predis/Command/Redis/SET_Test.php
@@ -12,6 +12,8 @@
 
 namespace Predis\Command\Redis;
 
+use Predis\ClientException;
+
 /**
  * @group commands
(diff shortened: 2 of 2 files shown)


tillkruss (changes requested):
Tests failing.

janedbal:
> Tests failing.

Adjuted. The remaining failures are unrelated.

vladvildanov (approved):

tillkruss:
This feels a bit overkill, especially since we merged #1509?

janedbal:
> This feels a bit overkill, especially since we merged https://github.com/predis/predis/pull/1509?

@tillkruss Not everybody uses PHPStan, so I think your library should either:

1. Note BC break in 2.3.0 that named args are no longer supported for SET
2. Fix the BC break by supporting named args again
3. Decide to no longer support named args to enable [similar changes](https://github.com/predis/predis/pull/1471/files) without a risk of breaking things _(this MR)_

tillkruss:
> 1. Note BC break in 2.3.0 that named args are no longer supported for SET

I don't see this as a BC issue, was this feature listed anywhere ever or did it just happen to work?

janedbal:
> was this feature listed anywhere ever or did it just happen to work?

I believe users are expecting native PHP features to work.

> I don't see this as a BC issue

If my code was working fine before upgrade, but fails hard on `ERR syntax error` after upgrade, I'd say it is a BC break. Even when unintentional.

janedbal:
If you dislike this MR, I can send PR fixing the broken support of named args.

tillkruss:
@vladvildanov Any thoughts on this?

PR #1510, opened 2025-02-13
Added testing with 8.0

vladvildanov:
Added testing against 8.0. Changed test infrastructure to use client specific docker images.

Client specific images is a separate docker images provided and maintained by Clients & Ecosystems team within Redis LTD. They're used specifically for unified creation of test infrastructure across different clients, and provides out of the box different types of environments including Standalone, Cluster, Sentinel, Stack. Full support for TLS connections


tests/Predis/Command/Redis/ZREVRANGE_Test.php
```diff
@@ -108,6 +108,7 @@ public function testAddsWithscoresModifiersOnlyWhenOptionIsTrue(): void
 
     /**
      * @group connected
+     * @group relay-incompatible
```
tillkruss, line 111: This is a bug in Relay, will be fixed in the next release.

.github/workflows/docker-compose.yml
tillkruss, line 1: This is showing up in the action runs, can we move it?

tillkruss (changes requested):

.github/workflows/docker-compose.yml
vladvildanov: Okay, I'll move it to a root directory

tillkruss: Or move it to a `.github/support`.

tillkruss (approved):
This looks great. Can you summarize the changed in a changelog entry?

Do we want to reduce the number of PHP runs as well? We still need to test against PHP 7.2, but maybe we test fewer Redis versions and only run 8.4 against every version?

.github/workflows/docker-compose.yml
vladvildanov: @tillkruss I would suggest to keep it in the root directory. This way for developers it would be much easy to bring a whole test infrastructure that could be used during the development, cause it's pre-configured. It would be something like:

```
1. docker compose --profile all up -d
2. vendor/bin/phpunit
```

vladvildanov:
@tillkruss I agree, we can make it as a part of this PR. I'll take care of it

vladvildanov:
@tillkruss What about keep 7.2, 8.3, 8.4? PHP provides official support for two latest versions, so it's still possible that something might be broken in 8.3 as well

tillkruss (changes requested):
Can you restore testing against 4.0 as well. I doubt people use 3.0 still, but we do advertise it.

tillkruss:
Thanks! 4.0 so we test against some common old versions.

tillkruss (approved):

dmaier-redislabs (approved):
LGTM

PR #1519, opened 2025-03-05
fix(stream): Removed resource disconnection on GC

vladvildanov:
Stream object shouldn't close a resource during GC, it breaks a conception of persistent connections. The resource would be closed in connection during GC, respecting a persistent connection logic.

https://github.com/predis/predis/blob/main/src/Connection/StreamConnection.php#L78

### DIFF
--- a/.github/docker-compose.yml
+++ b/.github/docker-compose.yml
@@ -37,7 +37,7 @@ services:
       - standalone
 
   redis-cluster:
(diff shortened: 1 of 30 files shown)


tillkruss (commented):
Does this affect v2.x as well?

vladvildanov:
@tillkruss Nope, Stream class doesn't exists there, it's purely 3.x functionality to decouple things a bit

PR #1521, opened 2025-03-18
Added missing FT._LIST and BITFIELD_RO commands

vladvildanov:
Closes #1400

PR #1522, opened 2025-03-20
Added missing types for commands API

### LINKED ISSUE #1460: Type parameter that are not currently typed
**Is your feature request related to a problem? Please describe.**
Some methods, for instance [setex](https://github.com/predis/predis/blob/v2.x/src/ClientInterface.php#L160) have some untyped parameters `$seconds` and `$value`
- Is there any other valid type than `int` for `$seconds`?
- Is there any other relevant  type than `string` for `$value`? AFAIK it always ends up being stored as `string` in Redis.

The reason why the absence of typing is a problem is because it's error-prone, new people discovering the library or the code using the libraries or older people who don't have a their knowledge of the library fresh might ask the questions above or other ones such as
- Is this value nullable, if so what happens if I send `null` `false` here?
- Is it really `mixed`? I can put `array` or `object` and they will get serialized?

Those are all doubts that can actually be cleared out right away if the type-hints reflect accurately what is relevant to be passed.

**Describe the solution you'd like**
Method such as:
```
@method Status setex(string $key, $seconds, $value)
```
To be transformed into:
```
@method Status setex(string $key, int $seconds, string $value)
```

**Describe alternatives you've considered**
Other ways I could think of to enforce types are:
- quite tedious to implement
- are not as good in term of discoverability

vladvildanov:
Closes #1460


src/ClientContextInterface.php
```diff
@@ -136,23 +136,23 @@
...
- * @method $this psetex($key, $milliseconds, $value)
- * @method $this set($key, $value, $expireResolution = null, $expireTTL = null, $flag = null)
- * @method $this setbit($key, $offset, $value)
- * @method $this setex($key, $seconds, $value)
- * @method $this setnx($key, $value)
- * @method $this setrange($key, $offset, $value)
+ * @method $this psetex(string $key, int $milliseconds, string|int|float $value)
+ * @method $this set(string $key, string|int|float $value, ?string $expireResolution = null, ?int $expireTTL = null, ?string $flag = null)
+ * @method $this setbit(string $key, int $offset, string|int|float $value)
+ * @method $this setex(string $key, int $seconds, string|int|float $value)
```
tillkruss, line 153: This might be a bit strict. Keys could be numeric as well and get casted to a string no? Same for values, could they not be objects that implement `__toString()`?

tillkruss (changes requested):

src/ClientContextInterface.php
vladvildanov: Yeah, you're right, I double checked and it's possible to use integers as keys as they anyway will be encoded as bulk strings. And yeah, it's possible to pass objects that implements `__toString()`, it's even possible to pass boolean as value however string conversion works non-intuitive (true -> '1', false -> ''), the only exception is arrays. I think it makes sense to mark all of it as `mixed`?

kylekatarnls: I think explicit `(string)` cast should be done on user-side. If you go this way all way down, then values can also be `null` `bool` or stringify-able objects, so you end up with `mixed` everywhere which adds no value.

And BTW `mixed` will allow `array` while it's not actually valid.

tillkruss: We might need to revise #1521 as well @vladvildanov.

tillkruss: `mixed` is what Relay used as well, I'm not sure it's worth adding for key-names and values TBH because it's implied already.

vladvildanov: @tillkruss In this way it will be inconsistent to all other commands, if you walk through the interface case are usually typed as strings

tillkruss: My vote is to not touch the stubs, it's been working and restricting them will unleash dozens of tickets I don't want to tackle.

PR #1530, opened 2025-04-15
 [2.x] Fix the problem of protocol loss during cluster MOVED/ASK

vladvildanov:
Fix the problem of protocol loss during redis cluster MOVED/ASK


src/Connection/Cluster/RedisCluster.php
```diff
@@ -337,10 +337,19 @@ protected function createConnection($connectionID)
...
+        $parameters = [
             'host' => substr($connectionID, 0, $separator),
             'port' => substr($connectionID, $separator + 1),
-        ]);
+        ];
+
+        $existConnection = current($this->pool);
+        if ($existConnection instanceof NodeConnectionInterface) {
+            $existParameters = $existConnection->getParameters()->toArray();
+            unset($existParameters['alias'], $existParameters['slots']);
```
vladvildanov, line 348: @tillkruss Removed slots as well, shouldn't be applied to different connection if already covered by another one

tillkruss (approved):

PR #1535, opened 2025-05-05
fix(auth): Adds default user if only password used for authentication

### LINKED ISSUE #1534: WRONGPASS invalid username-password pair or user is disabled after upgrading to predis 3
**Describe the bug**
Just after upgrading to [predis v3.0.0](https://github.com/predis/predis/releases/tag/v3.0.0) connection fails with:

```console
  Predis\Connection\ConnectionException 

Failed: WRONGPASS invalid username-password pair or user is disabled. [tcp://127.0.0.1:6379]

at vendor/predis/predis/src/Connection/AbstractConnection.php:157
  153▕      */
  154▕     protected function onConnectionError($message, $code = 0): void
  155▕     {
  156▕         CommunicationException::handle(
➜ 157▕             new ConnectionException($this, "$message [{$this->getParameters()}]", $code)
  158▕         );
  159▕     }
  160▕ 
  161▕     /**

    +28 vendor frames 

29  artisan:35
    Illuminate\Foundation\Console\Kernel::handle()
```

**To Reproduce**
1. Change `composer.json` requirements from `"predis/predis": "^2.4"` to `"predis/predis": "^3.0"`;
2. Run `composer update`;
3. Run `artisan cache:clear`;
4. See error above.

**Expected behavior**
`artisan cache:clear` finishing with message `INFO  Application cache cleared successfully.` and no error.

**Versions**
- Laravel: 11.44.7
- Predis: 3.0.0
- PHP: 8.3.20
- Redis Server: 7.4+ (incl. 8)
- OS: Ubuntu 24.04.2 LTS

**Code sample**
N/A

**Additional context**
Laravel example for completion purposes. It happens just the same on a test app with predis being the single requirement. predis v2.4.0 works fine though.

- `.env` contains:

    ```env
    CACHE_DRIVER=redis
    SESSION_DRIVER=redis
    REDIS_PASSWORD=secret
    ```

- `config/database.php` contais:

    ```php
    'redis' => [
        'client' => env('REDIS_CLIENT', 'predis'),

        'options' => [
            'cluster' => env('REDIS_CLUSTER', 'redis'),
            'prefix' => env(
                'REDIS_PREFIX',
                Str::slug(env('APP_NAME','laravel'), '_').'_database_'
            ),
            'scheme' => env('REDIS_SCHEME', 'unix'),
        ],

        'default' => [
            'url' => env('REDIS_URL'),
            'path' => env('REDIS_PATH', '/var/run/redis/redis-server.sock'),
            'password' => env('REDIS_PASSWORD'),
            'database' => env('REDIS_DB', 0),
        ],

        'cache' => [
            'url' => env('REDIS_URL'),
            'path' => env('REDIS_PATH', '/var/run/redis/redis-server.sock'),
            'password' => env('REDIS_PASSWORD'),
            'database' => env('REDIS_CACHE_DB', 1),
        ],
    ],
    ```

vladvildanov:
Closes #1534


src/Response/Iterator/MultiBulk.php
```diff
@@ -40,6 +40,8 @@ public function __construct(NodeConnectionInterface $connection, $size)
      */
     public function __destruct()
     {
+        // TODO: Disconnect on every iterator makes it fail during handshake
+        // TODO: (e.g connect() iterates over commands and eventually disconnects which kill the resource and makes connection broken)
```
tillkruss, line 44: Do you want to resolve this before merging?

tests/Predis/ClientTest.php
```diff
@@ -1343,6 +1343,7 @@ public function testClientsCreateDifferentPersistentConnections(): void
     /**
      * @group connected
      * @group cluster
+     * @group relay-incompatible
```
tillkruss, line 1346: Relay sets `default` as username if only password is specified.

src/Connection/Factory.php
```diff
@@ -167,10 +167,11 @@ protected function prepareConnection(NodeConnectionInterface $connection)
...
 
-            isset($parameters->username) && strlen($parameters->username)
-                ? array_push($cmdAuthArgs, $parameters->username, $parameters->password)
-                : $cmdAuthArgs[] = $parameters->password;
+            if (!isset($parameters->username) || !strlen($parameters->username)) {
+                $parameters->username = 'default';
+            }
 
+            array_push($cmdAuthArgs, $parameters->username, $parameters->password);
             array_push($cmdAuthArgs, 'SETNAME', 'predis');
```
elena-kolevska, line 175: Is it ok we're pushing a different command into the same cmd auth arguments?

src/Connection/Factory.php
```diff
@@ -167,10 +167,11 @@ protected function prepareConnection(NodeConnectionInterface $connection)
         if (isset($parameters->password) && strlen($parameters->password)) {
             $cmdAuthArgs = [$parameters->protocol, 'AUTH'];
 
-            isset($parameters->username) && strlen($parameters->username)
-                ? array_push($cmdAuthArgs, $parameters->username, $parameters->password)
-                : $cmdAuthArgs[] = $parameters->password;
+            if (!isset($parameters->username) || !strlen($parameters->username)) {
+                $parameters->username = 'default';
+            }
```
elena-kolevska, line 172:
```suggestion
            if (!empty($parameters->username)) {
                $parameters->username = 'default';
            }
```

tests/PHPUnit/PredisConnectionTestCase.php
```diff
@@ -539,9 +541,9 @@ protected function assertNonPersistentConnection($resource): void
      *
      * @return NodeConnectionInterface
      */
-    protected function createConnection(bool $initialize = false): NodeConnectionInterface
+    protected function createConnection(bool $initialize = false, bool $noAuth = false): NodeConnectionInterface
```
elena-kolevska, line 544: As mentioned above, I think we should have both cases, with and without pass

src/Connection/StreamConnection.php
```diff
@@ -316,9 +316,20 @@ private function handleError(ErrorResponseInterface $error, CommandInterface $fa
             if (in_array('AUTH', $failedCommand->getArguments(), true)) {
                 $parameters = $this->getParameters();
 
+                // If Redis > 6.2
```
elena-kolevska, line 319:
```suggestion
                // If Redis >= 6.0
```

src/Connection/StreamConnection.php
```diff
@@ -316,9 +316,20 @@ private function handleError(ErrorResponseInterface $error, CommandInterface $fa
...
                 $auth = new RawCommand('AUTH', [$parameters->username, $parameters->password]);
```
elena-kolevska, line 320: If I understand correctly, this would run after we've gotten an error on the AUTH we sent in the factory. If that's the case, why repeat the same command (with username and pass), why not go straight to the pre 6.0 version with no username?

.github/docker-compose.yml
```diff
@@ -23,9 +25,10 @@ services:
     environment:
       - TLS_ENABLED=yes
       - REDIS_CLUSTER=no
+      - REDIS_PASSWORD=foobar
       - PORT=6379
       - TLS_PORT=6666
-    command: ${REDIS_EXTRA_ARGS:---enable-debug-command yes --enable-module-command yes --tls-auth-clients optional --save ""}
+    command: ${REDIS_EXTRA_ARGS:---enable-debug-command yes --enable-module-command yes --tls-auth-clients optional --save "" --requirepass "foobar"}
```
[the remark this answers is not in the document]
vladvildanov: Hmm, I think it's really unnecessary. We will have twice more containers, we will need to distinguish tests for different type of infrastructure and I don't see so much use cases when we want to test against specifically password or passwordless infra, do you think there is?

src/Connection/Factory.php
vladvildanov: Unfortunately, we cannot use pipeline abstraction on connection level. Also the interface of connection accepts handshake commands one by one and add them into queue that will be flushed in connection time

`public function addConnectCommand(CommandInterface $command);`

.github/docker-compose.yml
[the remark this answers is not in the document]
vladvildanov: AFAIR other clients also uses password protected infrastructure only, f.e Jedis

https://github.com/redis/jedis/blob/master/Makefile

src/Connection/StreamConnection.php
vladvildanov: Yeah, I think you're right, before 6.0 username wasn't an option so there's no way there's case when before 6.0 when AUTH wouldn't work without username

tests/Predis/ClientTest.php
vladvildanov: Yeah, I think the problem is not in this test, Relay cluster setup just randomly fails probably cause infrastructure sometimes not yet ready when tests are running

src/Connection/StreamConnection.php
vladvildanov: Done

src/Response/Iterator/MultiBulk.php
vladvildanov: Is it possible to postpone it a bit and remove from the scope of this PR? Have other priorities in meantime

.github/docker-compose.yml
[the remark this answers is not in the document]
elena-kolevska: I don't know the whole codebase and don't know what we're already covering, but we definitely have to have a good coverage of all the possible combinations. What happens if we end up sending `auth default..` to a non-password protected server? What happens if we send default, but a username has been set.. and so on. Good coverage will protect us from future regressions like the one this PR is fixing.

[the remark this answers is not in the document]
tillkruss: Probably best to test against password, less do against unsecured instances. Much less common.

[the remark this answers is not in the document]
vladvildanov: Now all infra is password protected and I added one standalone unprotected for very specific test cases, I think that's the best we can do here to provide full coverage

tests/PHPUnit/PredisConnectionTestCase.php
vladvildanov: Resolved

src/Response/Iterator/MultiBulk.php
vladvildanov: @tillkruss My bad, it wasn't a bug it was just an issue with test setup, now Iterator tests are executing correctly against protected server

vladvildanov:
@tillkruss Ready for review

elena-kolevska (commented):
lgtm

src/Connection/StreamConnection.php
```diff
@@ -316,9 +316,13 @@ private function handleError(ErrorResponseInterface $error, CommandInterface $fa
...
 
-                $auth = new RawCommand('AUTH', [$parameters->username, $parameters->password]);
+                // If Redis <= 6.0
+                $auth = new RawCommand('AUTH', [$parameters->password]);
                 $response = $this->executeCommand($auth);
-                $this->handleOnConnectResponse($response, $auth);
+
+                if ($response instanceof ErrorResponseInterface) {
+                    $this->onConnectionError("Failed: {$response->getMessage()}");
+                }
```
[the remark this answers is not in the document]
vladvildanov: This is basically authentication for Redis < 6.0 when it was done with password only. Call to `handleOnConnectResponse()` is incorrect as it leads to infinite recursion, so if if error was thrown eventually it raises ConnectionError now

tillkruss (approved):

PR #1537, opened 2025-05-06
Update tests against Relay

yatsukhnenko:


tillkruss (changes requested):
Looks like only a few cluster failures. This is great!

yatsukhnenko:
It is because of unstable GitHub CI, try to re-run failed job

tillkruss (approved):

.github/docker-compose.yml
```diff
@@ -44,11 +44,11 @@ services:
       - NODES=6
       - REPLICAS=1
       - TLS_ENABLED=yes
-      - PORT=16379
+      - PORT=6372
```
vladvildanov, line 47: Is there's a reason why port was changed? It looks like this variable is used internally for container setup and since it's internal it has nothing to do with public port

yatsukhnenko: The reason is that in tests cluster exposes internal container port to client and not public ports
https://github.com/predis/predis/actions/runs/14864669804/job/41738341970#step:20:46

vladvildanov: Got it!

vladvildanov (approved):

PR #1540, opened 2025-05-10
Added support for UNLINK command

markyys:
I initially implemented this against 2.4.0 at work, so this is from memory. I also guessed at the test implementation, basically just copying the DEL tests.

If not exactly this, something like this should be done to allow UNLINK to work. A similar change should likely be make against the 2.x branch.

### DIFF
--- a/src/Command/Redis/UNLINK.php
+++ b/src/Command/Redis/UNLINK.php
@@ -1,0 +1,34 @@
+<?php
+
+/*
+ * This file is part of the Predis package.
+ *
+ * (c) 2009-2020 Daniele Alessandri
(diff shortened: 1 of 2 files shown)


tillkruss (changes requested):
Tests are failing and it needs a changelog entry 👍

markyys:
I'll be honest, I have no desire to do the work on this. I almost just opened an issue saying "you're missing UNLINK" but I figured it was a bit friendlier to put in some of the work, especially since I had already hacked this in on a local copy of 2.4.0.

But I have no interest in investigating what it takes to make all the tests pass, I don't even run redis at home.

PR #1546, opened 2025-05-26
Fix return type for ZCOUNT

TimWolla:
ZCOUNT returns an `int`:

> Integer reply: the number of members in the specified score range.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## Changelog
 
+## Unreleased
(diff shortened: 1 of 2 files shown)


vladvildanov:
@TimWolla Hi! I think v2.x is CVE and bugfixes only, correct me if I'm wrong @tillkruss

TimWolla:
> and bugfixes

I would say that fixing the return type to match the reality is a bugfix :smile:

PR #1548, opened 2025-05-28
fix(stream): Fixed return type to match function signature

vladvildanov:
Fixing thrown exception when stream is EOF, as specified in a signature CommunicationException should be thrown

### DIFF
--- a/src/Connection/StreamConnection.php
+++ b/src/Connection/StreamConnection.php
@@ -145,7 +145,7 @@ class StreamConnection extends AbstractConnection
         $stream = $this->getResource();
 
         if ($stream->eof()) {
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 ### Fixed
 - Fixed PHP 8.4 deprecated call to `stream_context_set_option()` (#1545)
 - Fixed return type for `ZCOUNT` to be `int` (#1547)
+- fix(stream): Fixed exception type to match function signature (#1548)
```
tillkruss, line 10: Can we be more specific? Something like:

```
Fixed throwing `CommunicationException` when stream is EOF.

vladvildanov: Done

PR #1550, opened 2025-06-03
feat(commands): Add support for VectorSet commands

vladvildanov:
Provides an API to interact with VectorSet data structure.

@tillkruss Vector Sets are currently in preview and the API may change. Could we please mark this feature as experimental in release notes so we can adjust API depends on future updates without introducing new major version?

https://redis.io/docs/latest/develop/data-types/vector-sets/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,9 @@
 ## Changelog
 
 ## Unreleased
-### Experimental
-- feat(commands): Add support for VectorSet commands (#1550)
+### Added
+- Add experimental support for vector sets commands (#1550)
+
 ### Changed
 - Handle and retry `LOADING` errors from Sentinel replicas (#1536)
 


src/ClientContextInterface.php
```diff
@@ -45,6 +45,7 @@
 use Predis\Command\Container\Search\FTCONFIG;
 use Predis\Command\Container\Search\FTCURSOR;
 use Predis\Command\Container\XGROUP;
+use Predis\Command\Redis\VADD;
```
tillkruss, line 48: Is this needed?

vladvildanov: Yes, we refer to a constant associated with this class as a default argument.

`* @method $this vadd(string $key, string|array $vector, string $elem, int $dim = null, bool $cas = false, string $quant = VADD::QUANT_DEFAULT, ?int $BEF = null, string|array $attributes = null, int $numlinks = null)`

htemelski (approved):

tillkruss (approved):

PR #1512, opened 2025-02-20
fix: Bug with single persistent connection to the same resource

vladvildanov:
Fixes #1506

PR #1551, opened 2025-06-04
fix(conn): Removed automatic conn_uid assignment

### LINKED ISSUE #1539: Incising connections: AUTH failed: ERR max number of clients reached
**Describe the bug**
Increasing connections in Redis cluster during write operations.
It might be related to https://github.com/predis/predis/pull/1530/files#diff-235ebc13f5c611f9f7be1e8856f07fbf12cd6dc2668e41c9fe4b180eae09c682R352

**To Reproduce**
1. When you use SET in iteration, this causes an increase in connections on Redis side with `AUTH failed: ERR max number of clients reached`
2. The v2.3 doesn't have this problem

**Expected behavior**
There is an assumption that new connections are created, but the previous ones are not closed. This is just an assumption.

- Predis: [e.g. 2.4.0]
- PHP [e.g. 8.2.0]
- Redis Server [e.g. 6.0.0]
- OS [e.g. Ubuntu 24.40]

**Code sample**
If applicable, a small snippet of code that reproduces the issue.

**Additional context**

vladvildanov:
Closes #1539 

Automatic `conn_uid` assignment was added as a part of #1512, but it leads to a connection spikes in case if failover happens and connections is recreated by client. Since, `conn_uid` is a parameter option it must be assigned by user instead to avoid sharing of the socket between different clients within a single process.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## Unreleased
 ### Fixed
 - Fixed return type for `ZCOUNT` to be `int` (#1546)
-- fix(conn): Removed automatic conn_uid assignment (#1551)
+- Removed automatic `conn_uid` parameter assignment (#1551)
 
 ## v2.4.0 (2025-04-30)
 ### Added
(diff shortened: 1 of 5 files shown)


vladvildanov:
@tillkruss The general code coverage was slightly decreased because a line of code was removed

tillkruss (approved):
Do we need to port this to v3.x?

PR #1557, opened 2025-06-11
implement XCLAIM command

alschastny:
Added implementation of the XCLAIM command.

A more complex command tests can be performed after implementing the XPENDING command.


src/ClientInterface.php
```diff
@@ -298,6 +299,7 @@
  * @method int               xack(string $key, string $group, string ...$id)
  * @method string            xadd(string $key, array $dictionary, string $id = '*', array $options = null)
  * @method array             xautoclaim(string $key, string $group, string $consumer, int $minIdleTime, string $start, ?int $count = null, bool $justId = false)
+ * @method array             xclaim(string $key, string $group, string $consumer, int $minIdleTime, string|array $ids, ?XClaimOptions $options = null)
```
vladvildanov, line 302: I would say in this case additional abstraction around optional arguments is not the best approach, it should be as simple as possible so everyone could run it like this `$client->xclaim(...$arguments)`

Something like this would work better: 
`xclaim(... ?int idle = null, ?int $time = null, ?int count = null, bool $force = false, bool $justId = false, ?string $lastId = null)`

src/Command/Redis/XCLAIM.php
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    public function prefixKeys($prefix)
+    {
+        $this->applyPrefixForFirstArgument($prefix);
+    }
+
+    public function parseResponse($data): array
+    {
+        if (!is_array($data) || !$data) {
```
vladvildanov, line 53: We don't need this condition, according to official docs the response type is always array, so we shouldn't expect any other types of responses

https://redis.io/docs/latest/commands/xclaim/

src/Command/Argument/Stream/XClaimOptions.php
```diff
@@ -0,0 +1,57 @@
...
+    public function __construct(
+        bool $justId = false,
+        ?int $idleMs = null,
+        ?int $timeMs = null,
+        ?int $retryCount = null,
+        bool $force = false,
+        ?string $lastId = null
+    ) {
+        if (null !== $idleMs) {
+            $this->options[] = 'IDLE';
```
vladvildanov, line 30: use `array_push()` function to append multiple items

src/ClientInterface.php
alschastny: Thanks for the review. I have made changes according to your recommendations.

vladvildanov (approved):

tillkruss (approved):

PR #1559, opened 2025-06-12
implement XSETID command

alschastny:
This is the last of the missing commands for redis streams (https://redis.io/docs/latest/commands/xsetid/). And although it is marked as internal in the documentation, sometimes I had to use it.


vladvildanov:
@alschastny The RESP2 test doesn't work with Relay, as Relay is RESP3 only and it returns already parsed response (associative array), we have to check if XINFO STREAM response is already associative array and does not apply parsing in this case

vladvildanov (approved):

PR #1560, opened 2025-06-13
XINFO command parse response changes + more tests

alschastny:
This PR corrects the parsing of responses from the XINFO command, and brings responses for the RESP2 and RESP3 protocols to a single form.

### DIFF
--- a/src/Command/Redis/XINFO.php
+++ b/src/Command/Redis/XINFO.php
@@ -14,6 +14,7 @@ namespace Predis\Command\Redis;
 
 use Predis\Command\Argument\ArrayableArgument;
 use Predis\Command\Command as RedisCommand;
+use Predis\Command\Redis\Utils\CommandUtility;
 
 class XINFO extends RedisCommand
 {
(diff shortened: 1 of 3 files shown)


src/Command/Redis/XINFO.php
```diff
@@ -47,17 +48,61 @@ private function setStreamArguments(array $arguments): void
...
+
+    private function parseStreamResponse($data): array
+    {
+        $result = CommandUtility::arrayToDictionary($data, null, false);
+
+        if (isset($result['entries'])) {
+            $result['entries'] = $this->parseDict($result['entries']);
+        }
+
+        if (isset($result['groups']) && is_array($result['groups'])) {
```
vladvildanov, line 67: The logic of parsing is quite nested because some arrays are semantically arrays and other is dictionaries?

alschastny: Yes, I added tests to make it clearer what’s going on inside. I also added a test that checks the consistency between RESP2 and RESP3 responses.

vladvildanov: @alschastny If you don't mind, I think `arrayToDictionnary()` method could be enhanced so it will accept a list of keys that should be excluded from parsing, WDYT? I can jump on it later this week

vladvildanov (approved):

PR #1566, opened 2025-06-18
BITOP added operation validation

htemelski-redis:
Added a check that only valid operations can be sent to the server.
Modified the tests to accommodate the changes.
Included as valid operations the four new operations that are going to be introduced with Redis 8.2

### DIFF
--- a/src/Command/Redis/BITOP.php
+++ b/src/Command/Redis/BITOP.php
@@ -41,9 +41,11 @@ class BITOP extends RedisCommand
             array_unshift($arguments, $operation, $destination);
         }
 
-        $operation = strtoupper($arguments[0]);
-        if (!in_array($operation, self::VALID_OPERATIONS, true)) {
-            throw new InvalidArgumentException('BITOP operation must be one of: AND, OR, XOR, NOT, DIFF, DIFF1, ANDOR, ONE');
+        if (!empty($arguments)) {
+            $operation = strtoupper($arguments[0]);
+            if (!in_array($operation, self::VALID_OPERATIONS, false)) {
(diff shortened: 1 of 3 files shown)


tests/Predis/Command/Redis/BITOP_Test.php
```diff
@@ -181,15 +182,15 @@ public function testBitwiseNOTAcceptsOnlyOneSourceKey(): void
     }
 
     /**
-     * @group connected
-     * @requiresRedisVersion >= 2.6.0
+     * @group disconnected
```
vladvildanov, line 185: @htemelski-redis could we also add a test with new BITOP operators? Is it already released in 8.0.2?

src/Command/Redis/BITOP.php
```diff
@@ -38,6 +41,13 @@ public function setArguments(array $arguments)
             array_unshift($arguments, $operation, $destination);
         }
 
+        if (!empty($arguments)) {
```
vladvildanov, line 44: We can skip this check, according to the interface BITOP command have 3 required arguments

`* @method int               bitop($operation, $destkey, $key)`

vladvildanov (changes requested):

src/Command/Redis/BITOP.php
htemelski-redis: The first commit was without this check. Turns out there are some tests that call BITOP with no arguments
https://github.com/predis/predis/actions/runs/15729316263/job/44326453078

vladvildanov (approved):

PR #1571, opened 2025-06-27
V2 fix method annotation clientinterface

thePanz:


vladvildanov:
I think v2.x is for critical bugs only, changes shouldn't be backport, right @tillkruss ?

tillkruss:
See https://github.com/predis/predis/pull/1570#issuecomment-3418767529

PR #1580, opened 2025-07-24
Added new stream commands, extended existing ones

htemelski-redis:
Added support for the new `XDELEX` and `XACKDEL` commands that will be introduced with Redis 8.2
Extended the existing `XTRIM` and `XADD` commands with the newly introduced parameters


src/Command/Redis/XACKDEL.php
```diff
@@ -0,0 +1,64 @@
...
+ */
+
+namespace Predis\Command\Redis;
+
+use Predis\Command\PrefixableCommand as RedisCommand;
+
+/**
+ * @see http://redis.io/commands/xackdel
+ */
+class XACKDEL extends RedisCommand
```
vladvildanov, line 20: You need to update `ClientInterface` and `ClientContextInterface` files to reflect new commands. This is were our public API defined

src/Command/Redis/XACKDEL.php
```diff
@@ -0,0 +1,64 @@
...
+    {
+        $processedArguments = [$arguments[0], $arguments[1]];
+
+        $argIndex = 2;
+        if (isset($arguments[$argIndex]) && in_array(strtoupper($arguments[$argIndex]), ['KEEPREF', 'DELREF', 'ACKED'])) {
+            $processedArguments[] = strtoupper($arguments[$argIndex]);
+            $argIndex++;
+        }
+
+        while (isset($arguments[$argIndex])) {
```
vladvildanov, line 43: I would suggest to define API the way `ids` argument will be `array`. This way the arguments handling should looks something like this:

```
array_push($processedArguments, 'IDS', count($arguments[3]), ...$arguments[3])
```

Very short and clear. Ids is a required argument so we don't even need to check if it exists

src/Command/Redis/XADD.php
```diff
@@ -35,7 +35,23 @@ public function setArguments(array $arguments)
...
+        $optionsOffset = 3;
+        $idOffset = 2;
+        $pushRefArg = false;
+
+        $refArguments = ['KEEPREF', 'DELREF', 'ACKED'];
+        $stringArgs = array_filter($arguments, function ($arg) {
+            return is_string($arg);
+        });
+
+        if (count(array_intersect($refArguments, $stringArgs)) == 1) {
```
vladvildanov, line 48: This whole block is overcomplicated. I think new argument could be another entry in `$options` something like `['trimming' => 'keepref']`.

This way we would follow the existing approach and we can much easier handle this by checking if key 'trimming' exists and if the value corresponds to one of the possible trimming types

src/Command/Redis/XDELEX.php
```diff
@@ -0,0 +1,64 @@
...
+    {
+        $processedArguments = [$arguments[0]];
+
+        $argIndex = 1;
+        if (isset($arguments[$argIndex]) && in_array(strtoupper($arguments[$argIndex]), ['KEEPREF', 'DELREF', 'ACKED'])) {
+            $processedArguments[] = strtoupper($arguments[$argIndex]);
+            $argIndex++;
+        }
+
+        while (isset($arguments[$argIndex])) {
```
vladvildanov, line 43: Same as `XACKDEL`

tests/Predis/Command/Redis/XACKDEL_Test.php
```diff
@@ -0,0 +1,176 @@
...
+
+namespace Predis\Command\Redis;
+
+use Predis\Command\PrefixableCommand;
+
+/**
+ * @group commands
+ * @group realm-stream
+ */
+class XACKDEL_Test extends PredisCommandTestCase
```
vladvildanov, line 21: Test needs to be updated after API changes I suggest

vladvildanov (changes requested):

src/Command/Redis/XACKDEL.php
htemelski-redis: I've noticed that other commands can accept the list of IDs as both variadic and array, so I want to keep the command usage consistent.

`HDEL_Test.php`

```php
    /**
     * @group disconnected
     */
    public function testFilterArguments(): void
    {
        $arguments = ['key', 'field1', 'field2', 'field3'];
        $expected = ['key', 'field1', 'field2', 'field3'];

        $command = $this->getCommand();
        $command->setArguments($arguments);

        $this->assertSame($expected, $command->getArguments());
    }

    /**
     * @group disconnected
     */
    public function testFilterArgumentsFieldsAsSingleArray(): void
    {
        $arguments = ['key', ['field1', 'field2', 'field3']];
        $expected = ['key', 'field1', 'field2', 'field3'];

        $command = $this->getCommand();
        $command->setArguments($arguments);

        $this->assertSame($expected, $command->getArguments());
    }
```

src/ClientContextInterface.php
```diff
@@ -285,10 +285,12 @@
  * @method $this tsrange(string $key, $fromTimestamp, $toTimestamp, ?RangeArguments $arguments = null)
  * @method $this tsrevrange(string $key, $fromTimestamp, $toTimestamp, ?RangeArguments $arguments = null)
  * @method $this xack(string $key, string $group, string ...$id)
+ * @method $this xackdel(string $key, string $group, string $mode, string ...$ids)
```
vladvildanov, line 288: `ids` arg is an array type, not enumeration

src/ClientContextInterface.php
```diff
@@ -285,10 +285,12 @@
...
  * @method $this xadd(string $key, array $dictionary, string $id = '*', array $options = null)
  * @method $this xautoclaim(string $key, string $group, string $consumer, int $minIdleTime, string $start, ?int $count = null, bool $justId = false)
  * @method $this xclaim(string $key, string $group, string $consumer, int $minIdleTime, string|array $ids, ?int $idle = null, ?int $time = null, ?int $retryCount = null, bool $force = false, bool $justId = false, ?string $lastId = null)
  * @method $this xdel(string $key, string ...$id)
+ * @method $this xdelex(string $key, string $mode, string ...$ids)
```
vladvildanov, line 293: same here

vladvildanov (changes requested):

PR #1592, opened 2025-10-06
Added prefixable interface for HEXPIRE, HEXPIRETIME

vladvildanov:


tillkruss:
@yatsukhnenko Can you take a look at these failures (now being ignored), they should be compatible with v0.12.x

Issue #1595, opened 2025-10-17
Infinite loop on TLS "Broken pipe"


esler:
**Describe the bug**
We experienced the issue when TLS connection was severed by network infrastructure. `fwrite()` in `Predis\Connection\Resource\Stream` throws following warning but crucially doesn't return `false` which code expects in case of an error.
```
PHP Warning:  fwrite(): SSL operation failed with code 5. OpenSSL Error messages:
error:80000020:system library::Broken pipe
```
Because an exception is not thrown `Predis\Connection\StreamConnection::write()` never breaks the loop. We get **infinite loop** and warnings are thrown in fast pace.
This could also happen even buffer is full - just without warning.

**To Reproduce**
It's tricky to simulate on actual Redis server but we can write small server which closes connection and generates similar problem.
```php
$context = stream_context_create([
    'ssl' => [
        'local_cert' => __DIR__ . '/server.pem', // self-signed cert
        'allow_self_signed' => true,
    ]
]);

$server = stream_socket_server("ssl://127.0.0.1:9443", $errno, $errstr, STREAM_SERVER_BIND|STREAM_SERVER_LISTEN, $context);
if (!$server) die("Server error: $errstr ($errno)\n");

echo "SSL server listening...\n";

while ($conn = @stream_socket_accept($server, -1)) {
    echo "Client connected, closing immediately...\n";
    fclose($conn);
}
```
Connecting to this server will throw following warning and falls in the same infinite loop.
```
Warning: fwrite(): SSL operation failed with code 5. OpenSSL Error messages:
error:80000020:system library::Broken pipe
```

**Expected behavior**
Code recognize issue with the socket and throws an exception.

**Versions (please complete the following information):**
- Predis 3.2.0
- PHP 8.3.25
- Redis Server 7.4.6
- OS Linux Alpine 3.22

vladvildanov:
@esler Thanks for reaching out! Looks like we're missing exception handling somewhere. I'm a bit busy now, so if you need it ASAP feel free to open a PR, otherwise would be able to jump on it in a couple of weeks

PR #1596, opened 2025-10-17
fail on broken TLS connection

esler:
Fix for #1595

There's a draft of the test, it requires a server simulating the issue. It can be omitted or I can add the server to the test suite as well.

### DIFF
--- a/src/Connection/Resource/Stream.php
+++ b/src/Connection/Resource/Stream.php
@@ -211,6 +211,16 @@ class Stream implements StreamInterface
             throw new RuntimeException('Unable to write to stream', 1);
         }
 
+        if ($result === 0) {
+            if ($this->eof()) {
+                throw new RuntimeException('Stream closed', 1);
+            }
+
+            if (stream_get_meta_data($this->stream)['timed_out']) {
+                throw new RuntimeException('Stream timed out', 1);
+            }
+        }
+
         return $result;
     }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 ### Changed
 - Refactor pipeline data writing depends on connection type (#1586)
 - Improved compatiblity with Relay (#1597)
+- Fail on broken TLS connection (#1595)
 
 ### Maintenance
 - Added testing with 8.4-M01 (#1593)


tillkruss (changes requested):
@vladvildanov I'm fine to merge this without a test. Just needs a changelog entry @esler.

esler:
Hi @vladvildanov, it's updated

src/Connection/Resource/Stream.php
```diff
@@ -211,6 +211,18 @@ public function write(string $string): int
             throw new RuntimeException('Unable to write to stream', 1);
         }
 
+        if ($result === 0) {
```
vladvildanov: I think there's no point to introduce a separate check. Currently the issue is that if statement above is to strict about type. Something like this would fix an issue:

`if ($result == false)`

src/Connection/Resource/Stream.php
```diff
@@ -211,6 +211,18 @@ public function write(string $string): int
...
+            if ($this->eof()) {
+                throw new RuntimeException('Stream closed', 1);
+            }
+
+            if (stream_get_meta_data($this->stream)['timed_out']) {
+                throw new RuntimeException('Stream timed out', 1);
```
vladvildanov: It doesn't make sense to add this additional exception handling, the exception type is the same as well as error code. We already throws exception with same inputs:

`throw new RuntimeException('Unable to write to stream', 1);`

esler: Yes we're unable to write to stream but for different reason. It might be useful for the user IMHO (timeout might indicated insufficient network configuration).

src/Connection/Resource/Stream.php
esler: `$result` contains the number of bytes successfully written to the socket buffer. A value of `0` is a **valid** and **successful** result (different from `false`). It typically indicates that the remote side is busy and the local send buffer is full, preventing additional data from being written at the moment.

src/Connection/Resource/Stream.php
vladvildanov: But how user will distinguish timeout from another reason? By parsing an exception message?

src/Connection/Resource/Stream.php
```diff
@@ -211,6 +211,18 @@ public function write(string $string): int
...
+            }
+
+            usleep(10000);
```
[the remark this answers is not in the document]
vladvildanov: Yeah, concerns me a bit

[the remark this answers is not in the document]
esler: It's gone. I'll create separate PR to deal with _busy loop_ on higher level.

PR #1597, opened 2025-10-17
Improve Relay compatibility

yatsukhnenko:


src/Command/Redis/VINFO.php
```diff
@@ -32,7 +32,11 @@ public function getId(): string
     public function parseResponse($data): ?array
     {
         if (!is_null($data)) {
-            return CommandUtility::arrayToDictionary($data);
+            if ($data === array_values($data)) {
+                return CommandUtility::arrayToDictionary($data);
+            } else {
+                return $data; // Relay
```
tillkruss: Do we need a handler for `< 0.20.0`?

tillkruss (changes requested):

src/Command/Redis/VINFO.php
yatsukhnenko: I thought about `atypicalCommands`, this allow to use native command and it may work (I need to check this)

yatsukhnenko: Nope, native command doesn't helped, looks like we can't make it work before `0.20`

PR #1607, opened 2025-11-09
Added experimental support for FT.HYBRID

vladvildanov:
Feature is consider experimental as interface and response structure might be flacky


vladvildanov:
@tillkruss Yeah, but most of it is tests cause API is quite extensive 😄 Coverage decrease is actually wrong, all of the methods is tested within integration tests

PR #1616, opened 2025-11-26
Added retry support

### LINKED ISSUE #1595: Infinite loop on TLS "Broken pipe"

vladvildanov:
Closes #1595 
Closes #1596
Closes #1613 

This PR aims to bring in-built retry functionality, for the communication exceptions that might sense to retry (e.g broken connection, timeout, socket initialisation failure). By default, all the currently existing retries (for cluster, replication) is kept, but it's possible to override them by providing custom retry configuration

### DIFF
--- a/tests/Predis/ClientTest.php
+++ b/tests/Predis/ClientTest.php
@@ -1556,6 +1556,7 @@ class ClientTest extends PredisTestCase
 
     /**
      * @group connected
+     * @group relay-incompatible
      * @return void
      * @requiresRedisVersion >= 7.0.0
      */


vladvildanov:
@tillkruss Does Relay using different type of underneath resource then by default PHP uses? It looks like it's not compatible in terms of metadata, I would like to know how to check whether resource has been timed out

vladvildanov:
@esler Hi! Would appreciate your review on this part, I think it covers what you introduced in another PR

https://github.com/predis/predis/pull/1616/files#diff-ef84ec023a9c60c0c568e9d11ee1e92aaebfe84fb5f7fd68eeb7378e82cdad54R210

src/Connection/Resource/Stream.php
```diff
@@ -207,7 +207,27 @@ public function write(string $string): int
...
+            }
+
+            if ($metadata['timed_out']) {
+                throw new RuntimeException(
+                    'Stream has been timed out',
+                    2
+                );
+            }
+
             throw new RuntimeException('Unable to write to stream', 1);
```
esler: `fwrite()` returning `0` is valid state if EOF and timeout doesn't appear. This could be quite common on slower non-blocking connection especially when SSL is involved. I would suggest to return `0` and add small sleep to avoid busy loop #1596 #1613

tillkruss:
@vladvildanov We just tagged Relay v0.20.0 that changes retry behaviour yfyi.

src/Connection/Replication/SentinelReplication.php
```diff
@@ -566,7 +569,10 @@ private function getConnectionInternal(CommandInterface $command)
     protected function assertConnectionRole(NodeConnectionInterface $connection, $role)
     {
         $role = strtolower($role);
-        $actualRole = $connection->executeCommand(RawCommand::create('ROLE'));
+        $retry = $connection->getParameters()->retry;
```
petyaslavova: Why in some cases you are using the retry object from `this` and in others, from the connection? Can those retry objects' configurations be different?

src/Pipeline/Atomic.php
```diff
@@ -63,24 +66,24 @@ protected function getConnection()
     protected function executePipeline(ConnectionInterface $connection, SplQueue $commands)
     {
         $commandFactory = $this->getClient()->getCommandFactory();
-        $connection->executeCommand($commandFactory->create('multi'));
-
-        if ($connection instanceof AggregateConnectionInterface) {
-            $this->writeToMultiNode($connection, $commands);
```
petyaslavova: What happens with this case? It appears that it is not covered at this time. Is it because here we are in multi context and we should do everything on one node only?

README.md
```diff
@@ -509,6 +509,44 @@ $client = new Predis\Client('tcp://127.0.0.1', [
 For a more in-depth insight on how to create new connection backends you can refer to the actual
 implementation of the standard connection classes available in the `Predis\Connection` namespace.
 
+### Retry exceptions
+
+You can enable automatic retry that is disabled by default, to be able to reduce the amount
```
petyaslavova: amount of

vladvildanov:
@tillkruss As Relay have it's own internal retry maybe it makes sense to exclude RelayConnection from retry mechanism at all?

vladvildanov:
@tillkruss Yes, Relay is fully omitted

tillkruss:
Yep, @yatsukhnenko is gonna open a PR.

PR #1630, opened 2026-01-28
Added support for HOTKEYS container command

vladvildanov:


tests/Predis/Command/Redis/HOTKEYS_Test.php
```diff
@@ -0,0 +1,272 @@
...
+ */
+
+namespace Predis\Command\Redis;
+
+use Predis\Command\Container\HOTKEYS as Container;
+use ValueError;
+
+/**
+ * @group commands
+ * @group relay-incompatible
```
tillkruss: @yatsukhnenko Should we use raw command for these, or our unreleased functions?

PR #1645, opened 2026-02-20
Add upstream_driver option for CLIENT SETINFO upstream reporting

vchomakov:
## Summary

Adds `upstream_driver` option for [`CLIENT SETINFO LIB-NAME`](https://redis.io/docs/latest/commands/client-setinfo/) to allow upstream libraries to identify themselves, per Redis client guidelines.

## Usage

```php
// String
$client = new Client($params, [
    'upstream_driver' => 'laravel_v11.0.0',
]);
// → CLIENT SETINFO LIB-NAME predis(laravel_v11.0.0)

// Array (for multiple upstream drivers)
$client = new Client($params, [
    'upstream_driver' => ['laravel_v11.0.0', 'my-app_v1.0.0'],
]);
// → CLIENT SETINFO LIB-NAME predis(laravel_v11.0.0;my-app_v1.0.0)
```

When not specified, sends `predis` (current behavior unchanged).

## Changes

- `src/Configuration/Option/UpstreamDriver.php` (new) - option handler
- `src/Configuration/Options.php` - register `upstream_driver` option
- `src/Configuration/OptionsInterface.php` - add `@property` annotation
- `src/Configuration/Option/Connections.php` - pass upstream driver to factory
- `src/Connection/Factory.php` - builds library name with upstream info
- Unit tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ## v3.4.1 (2026-02-23)
 ### Added
-- Added `driver_info` option for CLIENT SETINFO upstream reporting (#1645)
+- Added `upstream_driver` option for CLIENT SETINFO upstream reporting (#1645)
 - Made `H(P)TTL` commands prefixable (#1639)
 - Made `(B)LMPOP` commands prefixable (#1643)
 


src/Connection/Factory.php
```diff
@@ -25,6 +25,11 @@ class Factory implements FactoryInterface
 {
     private $defaults = [];
 
+    /**
+     * @var string
+     */
+    private $upstreamDriver = '';
```
vladvildanov: I think the default should be `null` and not empty string, which is more suitable for cases when something is "not set"

src/Connection/Factory.php
```diff
@@ -197,4 +222,18 @@ protected function prepareConnection(NodeConnectionInterface $connection)
...
+     *
+     * @return string
+     */
+    protected function buildLibraryName(): string
+    {
+        if ($this->upstreamDriver === '') {
+            return 'predis';
+        }
+
+        return 'predis(' . $this->upstreamDriver . ')';
```
vladvildanov: You can use ternary operator here for simplicity

`return ($this->upstreamDriver) ? 'predis(' . $this->upstreamDriver . ')' : 'predis';`

PR #1669, opened 2026-04-22
Change lpush and rpush methods to accept single values

JoseGonzalez84:
Related with #1146 

PR for fix a couple of methods documentation.


vladvildanov:
@JoseGonzalez84 This is incorrect interpretation of Redis API. Both LPUSH and RPUSH accept enumeration of elements, for now it's defined as an array in our public API (which kind of suboptimal), but it shouldn't be a string. PHP enumeration is the closest semantically (e.g `...values`). However, this would be a breaking change in terms of API.

https://redis.io/docs/latest/commands/lpush/

JoseGonzalez84:
I have Predis version 2.4.0 (as seen in composer.lock). My PHP version is 8.4.17.

If I pass an array ($data) as the second parameter of the `rpush` method, I receive the exception `Warning: Array to string conversion`.

This method only works if `json_encode($data)` is passed as the second parameter (or a typed string variable).

Example:

the next code works

```php
public function addProcess(string $agentId, string $actionId, array $data): void
{
    $this->connection->rpush($this->path.":".self::KEY_PROCESS.":$agentId:$actionId", json_encode($data));
}
```

the next code throws an exception `Array to string conversion`

```php
public function addProcess(string $agentId, string $actionId, array $data): void
{
    $this->connection->rpush($this->path.":".self::KEY_PROCESS.":$agentId:$actionId", $data);
}
```

This is a strange situation, because if you work with the documentation, the problem is guaranteed. 

The Redis documentation is irrelevant here; PHP is in charge. And it won't let me pass a string sequence.

vladvildanov:
@JoseGonzalez84 Okay, I understand it. But it should go the other way around, the internal code should be fixed so it could work as specified in the public API

JoseGonzalez84:
> @JoseGonzalez84 Okay, I understand it. But it should go the other way around, the internal code should be fixed so it could work as specified in the public API

You're right. I'll try to work on this to resolve it.

PR #1687, opened 2026-06-10
Fix UNLINK missing single-array argument normalization

Brajk19:
I noticed that with predis 3.5.0 i started getting some errors when using it with symfony/rate-limiter.
https://github.com/symfony/symfony/blob/e8d62b2f2d20ab87ce21fcf50584bfa65cad8777/src/Symfony/Component/Cache/Traits/RedisTrait.php#L683-L687

symfony cache component would try to call unlink with array argument, and if function does not exist it would fallback to `del`. That's why everything worked with older versions. Now with 3.5.0 unlink is available but does not support array arg.

Since UNLINK is very similar to DEL, i think the  signature should be the same.

UNLINK does not have the `setArguments()` override (like DEL), so `unlink(['key1', 'key2'])` keeps the keys nested instead of flattening them like del() does. On a client with a key prefix, `applyPrefixForAllArguments()` then concatenates the prefix with an array, silently sending UNLINK <prefix>Array — the real keys are never deleted.

This PR adds the `normalizeArguments` call to match `DEL`.
Test `testFilterArgumentsAsSingleArray` is copied from `DEL_Test`

### DIFF
--- a/src/ClientContextInterface.php
+++ b/src/ClientContextInterface.php
@@ -77,7 +77,7 @@ use Predis\Command\Redis\VADD;
  * @method $this sort_ro(string $key, ?string $byPattern = null, ?LimitOffsetCount $limit = null, array $getPatterns = [], ?string $sorting = null, bool $alpha = false)
  * @method $this ttl($key)
  * @method $this type($key)
- * @method $this unlink(string ...$keys)
+ * @method $this unlink(array|string $keys)
  * @method $this append($key, $value)
  * @method $this arcount(string $key)
  * @method $this ardel(string $key, int ...$index)
--- a/src/ClientContextInterface.php
+++ b/src/ClientContextInterface.php
@@ -77,7 +77,7 @@ use Predis\Command\Redis\VADD;
  * @method $this sort_ro(string $key, ?string $byPattern = null, ?LimitOffsetCount $limit = null, array $getPatterns = [], ?string $sorting = null, bool $alpha = false)
  * @method $this ttl($key)
  * @method $this type($key)
- * @method $this unlink(array|string $keys)
+ * @method $this unlink(string[]|string $keyOrKeys, string ...$keys = null)
  * @method $this append($key, $value)
  * @method $this arcount(string $key)
  * @method $this ardel(string $key, int ...$index)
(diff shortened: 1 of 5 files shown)


src/ClientInterface.php
```diff
@@ -88,7 +88,7 @@
  * @method array             sort_ro(string $key, ?string $byPattern = null, ?LimitOffsetCount $limit = null, array $getPatterns = [], ?string $sorting = null, bool $alpha = false)
  * @method int               ttl(string $key)
  * @method mixed             type(string $key)
- * @method int               unlink(string ...$keys)
+ * @method int               unlink(string[]|string $keyOrKeys, string ...$keys = null)
```
Brajk19, line 91: now matches
https://github.com/predis/predis/blob/98180048d0f9fb3331acf78f0599dbd310163401/src/ClientInterface.php#L68

src/ClientContextInterface.php
```diff
@@ -77,7 +77,7 @@
  * @method $this sort_ro(string $key, ?string $byPattern = null, ?LimitOffsetCount $limit = null, array $getPatterns = [], ?string $sorting = null, bool $alpha = false)
  * @method $this ttl($key)
  * @method $this type($key)
- * @method $this unlink(string ...$keys)
+ * @method $this unlink(array|string $keys)
```
vladvildanov: API should match the one from `ClientInterface`

vladvildanov (changes requested):

src/ClientContextInterface.php
Brajk19: i wasn't sure about that so i matched it with `del` from same interface: https://github.com/predis/predis/blob/98180048d0f9fb3331acf78f0599dbd310163401/src/ClientContextInterface.php#L57

Brajk19: but i can use the one from ClientInterface if that is correct

vladvildanov:
@tillkruss The changes are backward-compatible, so it would be okay to consider this as a bug fix

src/ClientContextInterface.php
vladvildanov: Yeah, use this one. Not sure why DEL is not aligned, but it should be

Brajk19: fixed it, thanks
https://github.com/predis/predis/pull/1687/changes/61d67103d42c9a5d50382a220a9c8a3bd2501975

vladvildanov (approved):

tillkruss (approved):
@vladvildanov looks bc-safe to me!

PR #1698, opened 2026-07-05
Fix typing in ClientContextInterface.php

kylekatarnls:
Hi, 👋 

Related to https://github.com/vimeo/psalm/issues/10336#issuecomment-1790519219

Current version of predis:
```
ERROR: UndefinedMagicMethod
at foo.php:5
Magic method Predis\ClientContextInterface::zremrangebyscore does not exist (see https://psalm.dev/219)
            $pipe->zremrangebyscore($key, $min, $max);


ERROR: UndefinedMagicMethod
at foo.php:6
Magic method Predis\ClientContextInterface::zcard does not exist (see https://psalm.dev/219)
            $pipe->zcard($key);
```

After adding the space, those errors are gone.

Thanks 🙏


tillkruss:
@copilot add a changelog entry

PR #1608, opened 2025-11-10
Added temporary XREADGROUP_CLAIM command

vladvildanov:
Closes #1606

PR #1701, opened 2026-07-17
Added support for MAXCOUNT and MAXSIZE arguments for XREAD and XREADGROUP

vladvildanov:
Redis 8.10 adds two new options to XREAD and XREADGROUP:

- MAXCOUNT maxcount — a cumulative cap on the total number of entries returned across all streams in the command (unlike COUNT, which remains a per-stream limit)
- MAXSIZE maxsize — a soft cumulative cap on the total reply size in bytes across all streams

### API Changes

`xreadgroup_claim()` (transitional XREADGROUP API) — two optional parameters appended, fully backward compatible:

```
$redis->xreadgroup_claim(                                                                                                                                                                                      
    'group', 'consumer',                                                                                                                                                                                       
    ['stream1' => '>', 'stream2' => '>'],                                                                                                                                                                      
    50,     // count (per stream)                                                                                                                                                                              
    null,   // block                                                                                                                                                                                           
    false,  // noAck                                                                                                                                                                                           
    null,   // claim                                                                                                                                                                                           
    80,     // maxCount (cumulative)                                                                                                                                                                           
    65536   // maxSize (bytes, soft cap)                                                                                                                                                                       
); 
```

`xread_v4()` — new transitional command that will replace `xread()` in the next major version (same approach as XREADGROUP_CLAIM, #1608). The legacy `xread()` signature ends with variadic ids, so new options cannot be added to it without breaking positional callers. The new command uses a dict-based streams argument, so future arguments can be appended safely:                                   

```                                                                                                                                                                                                       
$redis->xread_v4(                                                                                                                                                                                             
    ['stream1' => '0-0', 'stream2' => '0-0'],                                                                                                                                                                  
    50,     // count (per stream)                                                                                                                                                                              
    null,   // block                                                                                                                                                                                           
    80,     // maxCount (cumulative)                                                                                                                                                                           
    65536   // maxSize (bytes, soft cap)                                                                                                                                                                       
);
```
`xread_v4()` sends XREAD on the wire, reuses the existing response parser, and — unlike the legacy `xread()` — supports key prefixing, locating keys dynamically from the STREAMS block. The legacy `XREAD` class is now marked @deprecated pointing to `XREAD_V4`, mirroring the `XREADGROUP → XREADGROUP_CLAIM` deprecation.

### Explicit deprecation of xread() and xreadgroup()

Both methods were previously "virtual" — declared only as `@method` annotations and dispatched through `__call()` — which left no way to signal deprecation to users, since PHPDoc cannot deprecate individual `@method` lines.

They are now declared as explicit methods marked `@deprecated`, proxying `func_get_args()` to `__call()` so the runtime behavior is unchanged:                                                                       
                                                                                                                                                                                                                 
- `Predis\Client::xread()` / `Predis\Client::xreadgroup()`                                                                                                                                                     
- `Predis\Pipeline\Pipeline::xread() / xreadgroup()` (inherited by all pipeline variants)                                                                                                                        
- `Predis\Transaction\MultiExec::xread() / xreadgroup()`

With real methods in place, IDEs render call sites with a strikethrough and static analyzers (PHPStan/Psalm with deprecation rules) report usages in consumer CI — neither was possible with annotation-only declarations.


src/Command/Redis/XREAD_NEW.php
```diff
@@ -0,0 +1,84 @@
...
+ */
+
+namespace Predis\Command\Redis;
+
+use Predis\Command\PrefixableCommand as RedisCommand;
+
+/**
+ * This is a transitional command. In the next major version this command will replace XREAD.
+ */
+class XREAD_NEW extends RedisCommand
```
tillkruss: Not sure I'm loving the `_NEW` naming. Let me think about this.

vladvildanov: Sure thing, if you have a better one let me know

tillkruss: Using `rawCommand` until we do a major version. `NEW` is too vague.

vladvildanov: We already have a case with `xreadgroup` and `xreadgroup_claim` it just wasn't properly deprecated which I did in this PR, so I think it would be consistent to do this with `xread`. Having new API would enable a smoother transition for the end users, also it would be easy to miss this out for the next major. Deprecation is also a way to control what we need to figure out before next major.

The idea of the new API is not only to allow usage of two new arguments, it's also a way to be able to add a new arguments in backward-compatible manner in a future, so we can call it something like `XREAD_STREAMS` that represents a core change (streams as dictionnary), or `XREAD_V4` so we make it clear that it will be used in 4.x

tillkruss: I like the `v4` that seems clear, eh?

vladvildanov: Yeah, that seems more clearer and refers to future changes in the next major

src/ClientInterface.php
```diff
@@ -346,8 +346,9 @@
  * @method array             xrevrange(string $key, string $end, string $start, ?int $count = null)
  * @method array             xrange(string $key, string $start, string $end, ?int $count = null)
  * @method array|null        xread(int $count = null, int $block = null, array $streams = null, string ...$id)
+ * @method array|null        xread_v4(array $keyIdDict, ?int $count = null, ?int $blockMs = null, ?int $maxCount = null, ?int $maxSize = null)
  * @method array             xreadgroup(string $group, string $consumer, ?int $count = null, ?int $blockMs = null, bool $noAck = false, string ...$keyOrId)
- * @method array             xreadgroup_claim(string $group, string $consumer, array $keyIdDict, ?int $count = null, ?int $blockMs = null, bool $noAck = false, ?int $claim = null)
+ * @method array             xreadgroup_claim(string $group, string $consumer, array $keyIdDict, ?int $count = null, ?int $blockMs = null, bool $noAck = false, ?int $claim = null, ?int $maxCount = null, ?int $maxSize = null)
```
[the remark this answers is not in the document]
vladvildanov: Updated

tillkruss (approved):

PR #1708, opened 2026-07-20
Added stream commands to ClusterStrategy

RobiNN1:


src/Cluster/ClusterStrategy.php
```diff
@@ -464,6 +480,65 @@ protected function getKeyFromSUnsubscribeCommand(CommandInterface $command): ?st
...
+     *
+     * @return string|null
+     */
+    protected function getKeyFromStreamReadCommands(CommandInterface $command)
+    {
+        $arguments = $command->getArguments();
+        $position = null;
+
+        foreach ($arguments as $index => $argument) {
+            if (is_string($argument) && strtoupper($argument) === 'STREAMS') {
```
vladvildanov: There's one edge case worth to take into a count. `XREADGROUP` command accept consumer group and consumer name as required arguments and if one of those would be named `streams` (which isn't that impossible, especially for consumer group name) it would be resolved incorrectly.

For `XREADGROUP` command it make sense to start iterating from `arguments[3]` omitting `GROUP group consumer` arguments

vladvildanov (changes requested):

vladvildanov (approved):

tillkruss (approved):

vladvildanov:
@RobiNN1 Would you be able to follow Copilot suggestion to fix the merge conflict introduced bug? My apologise for that

PR #1712, opened 2026-07-28
Added experimental support for HIMPORT bulk hash import feature

vladvildanov:
## Summary

Adds client support for Redis 8.10's `HIMPORT` command family — a faster ingestion mechanism for loading many hashes that share the same field names. Field names are sent once with `HIMPORT PREPARE`, registering them under a fieldset name in the connection's session; hashes are then created with `HIMPORT SET` by sending only the values. The resulting keys are ordinary hashes; the memory-efficient encoding is a server-side detail.

```
$client->himport->prepare('users', ['name', 'email', 'age']);                                                                                                                                                  
$client->himport->set('user:1', 'users', ['alice', 'alice@example.com', '25']);                                                                                                                                
$client->himport->set('user:2', 'users', ['bob', 'bob@example.com', '30']);                                                                                                                                    
$client->himport->discard('users');   // 1                                                                                                                                                                     
$client->himport->discardAll();        // number of fieldsets removed
```

Values map by position to the fields supplied to `prepare()`, in the caller's order — never reordered. Hash enumeration order (`HGETALL`) is not guaranteed to match `PREPARE` order; only the value↔field pairing is.                                                                                                                                                                                                            
                                                                                                                                                                                                               
The hard part of this feature isn't the wire format — **it's that a fieldset is server-side session state bound to a single physical connection and is lost on `disconnect / RESET / failover`**, while Predis reconnects lazily and (on a cluster) routes different keys to different nodes. Most of this PR is the state management that keeps `HIMPORT SET` working transparently without re-sending field names on every write.

###  State management

- `FieldsetRegistry` (client-level, canonical) — held in the himport client option **so it's created once, cached, and shared with derived per-node clients**. It maps fieldset name → ordered field list: the application's intent. It's the authoritative source used to rebuild a `PREPARE` for any connection that needs one, and it gates recovery. It deliberately keeps no per-connection "already prepared" flags — those can't stay honest across transparent reconnects (especially ext-relay); the server's `no such fieldset` error is the only reliable per-connection signal.                                                                                                                                                                                         
- `$sessionCommands` (per physical connection, derived) — gains a keyed queue of PREPAREs to replay when that specific socket reconnects.                                          
                                                                                                                                                                                                               
**Mutate-on-success**. `prepare()` records into the registry and pins the session command only after the server confirms the `PREPARE` (all masters, on a cluster). `discard()/discardAll()` remove from the registry and unpin only after the server confirms. This keeps the two layers consistent on failure and, critically, prevents a rejected `PREPARE` (e.g. duplicate field name) from leaving a poison registry entry that every later SET would keep trying to re-prepare.                                                                                                                                                                     
                                                                                                                                                                                                               
**Zero cost when unused**. The registry is created lazily (only the container touches the himport option), `$sessionCommands` is an empty-array property replayed as a no-op, and the command hot path is untouched — normal traffic never runs a line of `HIMPORT` code.

### Recovery: proactive replay + reactive re-prepare                                                                                                                                                               
                                                                                                                                                                                                               
Two mechanisms, with clearly separated triggers:                                                                                                                                                               
                                                                                                                                                                                                               
**Proactive replay** — always on, independent of any option. `StreamConnection::connect()` (and `RelayConnection::connect()`) replay the pinned session commands right after the init commands on every reconnect,     
before the next command is written. Because reconnect is lazy (driven by the next `writeRequest`), a `SET` issued after a dropped connection reconnects, replays its `PREPARE`, and then goes out on the already-prepared socket — it just succeeds, with no error and no retry. Replay is best-effort and lenient: a server error for one entry drops only that entry and never tears down the connection (unlike init-command failures, which raise `ConnectionException`).                                                                                                                                                       
                                                                                                                                                                                                               
**Reactive re-prepare + retry** — reuses the client's configured Retry. This covers the cases proactive replay can't: a fieldset that was never pinned on the connection serving the SET (a new cluster node reached via MOVED, a configured/pre-loaded fieldset, or after failover).

### Cluster specifics                                                                                                                                                                                              
                                                                                                                                                                                                               
A fieldset lives inside a single connection session, so on a cluster it must exist on every node that may run a `SET` for it:                                                                                    
                                                                                                                                                                                                               
- `prepare() / discard() / discardAll()` **fan out to every master shard**. Fan-out uses all-succeeded semantics: any node error is surfaced and the client-side state is not mutated. Each master carries its own `$sessionCommands`.                                       
- `HIMPORT SET` routes by the hash slot of its key like any other write.
- After a MOVED to a node created after `prepare()`, that node lacks the fieldset; with retries enabled, `getConnectionByCommand()` (post-move()) resolves to the new node, the container re-prepares there, pins  
it, and retries. Known limitation: during an active -ASK migration the single retry can still fail (the slot map isn't updated by ASK); the error is then surfaced.                                            
                                                                                                                                                                                                               
**Sentinel / replication**: `HIMPORT` is unregistered in `ReplicationStrategy`, so it defaults to a write → master. Fan-out collapses to the single master; after failover the first `SET` on the new master reactively re-prepares (when retries are enabled). ext-relay: proactive replay is bypassed by relay's internal reconnects, so the reactive path is the guarantee there.

### DIFF
--- a/README.md
+++ b/README.md
@@ -475,6 +475,24 @@ case server errors are propagated unchanged:
 $client = new Predis\Client($parameters, ['himport' => ['auto_prepare' => false]]);
 ```
 
+Fieldsets can also be declared up front through the same option. Fieldsets declared this way are
+prepared on demand the first time a `HIMPORT SET` references them on a connection, so the application
+never has to call `prepare()` for them (this uses the on-demand mechanism above, so keep `auto_prepare`
+enabled):
+
+```php
+$client = new Predis\Client($parameters, [
+    'himport' => [
+        'fieldsets' => [
+            'users' => ['name', 'email', 'age'],
+        ],
+    ],
+]);
+
+// No prepare() call needed — "users" is known from configuration:
+$client->himport->set('user:1', 'users', ['alice', 'alice@example.com', '25']);
+```
+
 On a cluster, `prepare()`, `discard()` and `discardAll()` fan out to every master shard, while `set()`
 is routed by the hash slot of its key like any other write. This ensures a `HIMPORT SET` succeeds on
 whichever shard owns its key.
--- a/README.md
+++ b/README.md
@@ -22,6 +22,7 @@ More details about this project can be found on the [frequently asked questions]
 - Command pipelining on both single nodes and clusters (client-side sharding only).
 - Abstraction for Redis transactions (Redis >= 2.0) and CAS operations (Redis >= 2.2).
 - Abstraction for Lua scripting (Redis >= 2.6) and automatic switching between `EVALSHA` or `EVAL`.
+- Abstraction for Hinted Hash Templates (`HIMPORT`, Redis >= 8.10) with automatic per-connection fieldset replay.
 - Abstraction for `SCAN`, `SSCAN`, `ZSCAN` and `HSCAN` (Redis >= 2.8) based on PHP iterators.
 - Connections are established lazily by the client upon the first command and can be persisted.
 - Connections can be established via TCP/IP (also TLS/SSL-encrypted) or UNIX domain sockets.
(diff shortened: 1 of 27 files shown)


README.md
```diff
@@ -434,6 +435,83 @@ $response = $redis->transaction(function (MultiExec $tx) {
 ```
 
 
+### Hinted Hash Templates (HIMPORT) ###
+
+> **Experimental:** this feature is experimental and its API (the `himport` container and option) may
+> change in a future release.
+
+`HIMPORT` (Redis >= 8.10) speeds up loading many hashes that share the same field names. The field
```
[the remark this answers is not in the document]
vladvildanov: The server version and API version are different, 8.9 is correct guard

src/Connection/RelayConnection.php
```diff
@@ -369,15 +369,30 @@ protected function initializeUnixConnection(ParametersInterface $parameters)
...
+            } elseif ($response instanceof ErrorResponseInterface) {
+                $this->onConnectionError("`{$command->getId()}` failed: {$response->getMessage()}", 0);
+            }
+        }
+
+        // Best-effort replay of session state (e.g. HIMPORT fieldsets): a failed
+        // command drops only its own entry and never tears down the connection.
+        // Note the "relay" extension reconnects internally without re-entering
+        // this method, so this replay does not cover relay-driven reconnects;
+        // those are healed reactively when the next dependent command runs.
```
tillkruss: Smells like Claude 😅

Do we need the comment in here for a specific command, refactor seems fine to me.

vladvildanov: Yeah, it's not best-effort here it's actual a proper implementation for Relay. Removed it 😅

vladvildanov:
@tillkruss The PR ready for another review 👍

tillkruss (approved):
Man, 3000 lines of code for one function 😬

vladvildanov:
@tillkruss Well, this is what connection state management costs. When you extrapolate it to the different client architectures and consistency guarantees for failovers or disconnections, here we are 🥲

PR #1716, opened 2026-08-21
Fix RESP3 double parsing returning positive INF for -inf

semx:
With `protocol => 3`, `Resp3Strategy::parseDouble()` mishandles the special double payloads:

- `,-inf\r\n` is returned as **positive** `INF` - the sign is inverted:
  ```php
  if ($string === 'inf' || $string === '-inf') {
      return INF;
  }
  ```
- `,nan\r\n` falls through to the `(float)` cast, and `(float) 'nan'` is `0.0` in PHP, so NaN silently becomes zero.

Both are reachable with real commands, e.g. `ZSCORE` / `ZRANGEBYSCORE WITHSCORES` on a member whose score is `-inf`:

```php
$client = new Predis\Client(['protocol' => 3]);
$client->zadd('zset', ['member' => '-inf']);
$client->zscore('zset', 'member'); // float(INF) instead of float(-INF)
```

This PR returns `-INF` for `-inf` and `NAN` for NaN payloads. Per the RESP3 specification, Redis before 7.2 may emit any libc representation of NaN (`-nan`, `NAN`, `nan(char-sequence)`), and "client implementations should be able to handle this correctly", so those spellings are accepted too.

The existing infinity test only asserted `is_infinite()`, which cannot catch the sign inversion; it now checks exact values, and NaN cases are added. `tests/Predis/Protocol/` passes (73 tests); the full `--group disconnected` run shows the same pre-existing failures as `main` (tests that expect a live server on 127.0.0.1:6379), with no new ones.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 ### Added
 ### Changed
 ### Fixed
+- Fixed RESP3 double parsing returning positive `INF` for `-inf` payloads (#1716)
 
 ## v3.6.0 (2026-08-14)
 ### Added
--- a/src/Protocol/Parser/Strategy/Resp3Strategy.php
+++ b/src/Protocol/Parser/Strategy/Resp3Strategy.php
@@ -63,8 +63,12 @@ class Resp3Strategy extends Resp2Strategy
      */
     protected function parseDouble(string $string): float
     {
-        if ($string === 'inf' || $string === '-inf') {
+        if ($string === 'inf') {
             return INF;
+        }
+
+        if ($string === '-inf') {
+            return -INF;
         }
 
         return (float) $string;
--- a/tests/Predis/Protocol/Parser/Strategy/Resp3StrategyTest.php
+++ b/tests/Predis/Protocol/Parser/Strategy/Resp3StrategyTest.php
@@ -60,11 +60,11 @@ class Resp3StrategyTest extends PredisTestCase
      * @param  string $data
      * @return void
      */
-    public function testParseDataReturnsFloatInfinityOnInfinityOrNegativeInfinity(string $data): void
+    public function testParseDataReturnsFloatInfinityOnInfinityOrNegativeInfinity(string $data, float $expectedValue): void
     {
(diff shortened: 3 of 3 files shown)


tillkruss (approved):

tillkruss:
@vladvildanov I'd consider this a bug fix but technically its a breaking change?

vladvildanov (approved):

vladvildanov:
@tillkruss Yeah, since it's now returns NAN as well it's a breaking change

semx:
From the reporter side, for what it is worth: the -inf half is a pure correctness fix, the sign was inverted and consumers received +INF for a score of -inf, so I would not treat code relying on the flipped sign as a contract. The NaN half is the debatable one, since ",nan" used to surface as 0.0 and now surfaces as NAN. If that part is a concern for a minor release, I am happy to split the PR: land the -inf sign fix as a bugfix now and hold the NaN handling for the next major.

tillkruss:
@vladvildanov I could be persuaded to call this a bugfix for correctness and if causes trouble we revert it, acceptable to you?

vladvildanov:
@tillkruss I would rather agree with @semx and split the PR's to separate real bug with -INF apart of current NaN interpretation as 0.0. Changing it to NaN would break `json_encode()` of the result.

semx:
Done, split as suggested. This PR now contains only the -inf sign correction (plus the exact-value infinity assertions and its changelog line); the TDigest test updates went with the NaN half since they only matter once NaN surfaces. The NaN behavior change lives in #1718 with the json_encode() consideration spelled out, so you can schedule it independently.

semx:
CI note: the single red job (PHP 8.0 / Redis 8.8) is HOTKEYS_Test::testRetrieveHotKeys failing with "array does not contain key1", a sampling-dependent assertion unrelated to this diff (this branch only changes the -inf sign in parseDouble). The RESP3 variant of the same test failed on a different combo over in #1718, whose diff does not overlap with this one, which also points at flakiness of that test on the Redis 8.8 image.

tillkruss (approved):

Issue #1713, opened 2026-07-31
`SentinelReplication` retry callback rejects `StreamInitException`


marden:
## Problem

`SentinelReplication::retryCommandOnFailure()` defines its failure callback with
`CommunicationException` as the only accepted argument type. However, the retry
mechanism also treats `StreamInitException` as retryable and passes it to the
same callback.

As a result, an initial connection failure is masked by a `TypeError`.

## Actual behavior

When a Redis node or Sentinel cannot be reached, Predis throws:

```
Predis\Connection\Replication\SentinelReplication::{closure:...retryCommandOnFailure():756}():
Argument #1 ($exception) must be of type Predis\CommunicationException,
Predis\Connection\Resource\Exception\StreamInitException given,
called in /path/to/vendor/predis/predis/src/Retry/Retry.php on line 134
```

The original transport error (`StreamInitException`) is therefore not exposed
to the application.

## Expected behavior

`SentinelReplication` should either:

- accept and handle `StreamInitException` in the failure callback; or
- exclude `StreamInitException` from the retryable exceptions used by this
  specific retry path.

The original connection error should be preserved if all retries fail.

## Relevant code

`src/Connection/Replication/SentinelReplication.php`:

```php
$failCallback = function (CommunicationException $exception) {
    $this->wipeServerList();
    $exception->getConnection()->disconnect();
};
```

`src/Retry/Retry.php` includes `StreamInitException` in the default list of
catchable exceptions:

```php
protected $catchableExceptions = [
    TimeoutException::class,
    ConnectionException::class,
    StreamInitException::class,
];
```

## Environment

- Predis version: `3.5.1`
- Redis topology: Sentinel
- PHP version: `8.4`

vladvildanov:
@marden Thanks for reporting this, would have a look ASAP

vladvildanov:
@marden I think #1694 already fixes your issue, it just not yet released. Could you confirm it from your side?

marden:
Yes, it seems related. Is there a plan to release the fix in coming days?

vladvildanov:
@marden Yes, after we merge #1712 we're planing to release next minor

vladvildanov:
@marden The new release was cut and should fix this issue, feel free to re-open

Issue #1615, opened 2025-11-25
StreamInitException does not extend ConnectionException


tolry:
Is this intended or an oversight? 

We implement a retry mechanism for all redis commands, since we have flaky network connections at times, catching all `ConnectionException` occurrences. After upgrading to the most recent predis version, we noticed `StreamInitException` instances popping up.

We can, of course, also catch StreamInitExceptions, but I think this would better be suited as a child class of ConnectionException. What do you think?

vladvildanov:
@tolry Hi! Thanks for reporting! I'm currently working on in-built retry mechanism, I'll take it into a count

PR #1717, opened 2026-08-25
Retry cluster slot map discovery on StreamInitException

holystix:
`RedisCluster::queryClusterNodeForSlotMap()` overrides `Retry`'s default catchable exception list with `[ConnectionException::class]`. `StreamInitException` extends `PredisException` directly, so a `CLUSTER SLOTS` refresh that cannot open a connection to the node it was just redirected to is neither retried nor passed to the fail callback that would evict the unreachable node. It escapes uncaught, and since `onErrorResponse()` runs outside `retryCommandOnFailure()`, nothing upstream catches it either.

This is the same defect #1577 and #1650 fixed for `SentinelReplication` (the callback signature half came later, in #1712 for issue #1713). `RedisCluster` is the remaining call site. `Retry`'s default catchable list already contains `StreamInitException` — per issue #1615 that was the chosen resolution instead of re-parenting the exception — so dropping the override is enough.

**Change** Drops the override. `$failCallback` widens to `Throwable`, because `StreamInitException` is raised before a connection object exists and so carries none; the node being queried is evicted instead. `onFailCallback()` rediscovers the slot map for the same reason on the command path. `ConnectionException` behaviour is unchanged.

Retrying is safe: `StreamInitException` is thrown only by `StreamFactory` during stream creation, so the command provably never reached a server and a retry cannot duplicate a write.

**Behaviour change worth a look** On pool exhaustion the fail callback now rethrows the original exception instead of a `ClientException`. This is required rather than cosmetic — once `StreamInitException` is retryable, a single-seed cluster that cannot complete a TLS handshake empties the pool and would report pool exhaustion, masking the root cause that `SSLTest::testClusterExecuteCommandOverSSLConnectionWithoutSSLConfig` asserts must reach the caller. It also matches what `Retry` already does when the retry limit is reached, and no test asserted the `ClientException` message.

Seen in production on Azure Managed Redis (OSS clustering policy) behind Laravel on 3.5–3.6, during Azure-side failover and scaling events.


holystix:
`PipelineTest::testClusterExecutePipeline` is a pre-existing flake, not something this PR causes. It fails on `main` with no PR involved — runs [31516569237](https://github.com/predis/predis/actions/runs/31516569237) and [30991212363](https://github.com/predis/predis/actions/runs/30991212363) — the failing matrix cells vary between runs, and nothing here touches `Pipeline`. It looks cluster-shape dependent: the pipeline resolves against a cold slot map and takes a `MOVED` depending on which node owns the key's slot.

CI is green on the latest push.

vladvildanov (approved):

tillkruss (changes requested):
we don't need that many and long inline comments added, smells like Claude. If the code doesn't explain itself cut down the comment length

holystix:
Thanks for the review. I’ve removed the long inline comments, updated the branch with current main, and resolved the conflicts. Ready for another look.

CHANGELOG.md
```diff
@@ -4,7 +4,9 @@
 ### Added
 ### Changed
 - Deprecated `CommandInterface::deserializeCommand()` (CVE GHSA-w6f5-v2h6-g786)
+- `RedisCluster` now surfaces the original transport error instead of a `ClientException` when the pool is exhausted during `CLUSTER SLOTS` discovery (#1717)
 ### Fixed
+- Fixed Redis Cluster not retrying slot map discovery on `StreamInitException` (#1717)
```
vladvildanov: Please remove additional entry in CHANGELOGS

CHANGELOG.md
```diff
@@ -4,7 +4,9 @@
 ### Added
 ### Changed
 - Deprecated `CommandInterface::deserializeCommand()` (CVE GHSA-w6f5-v2h6-g786)
+- `RedisCluster` now surfaces the original transport error instead of a `ClientException` when the pool is exhausted during `CLUSTER SLOTS` discovery (#1717)
```
vladvildanov: Make it matching to PR name, there's no need to add such details in CHANGELOGs

vladvildanov (changes requested):

vladvildanov (approved):

PR #1718, opened 2026-08-26
Return NAN for RESP3 NaN double payloads

semx:
Split out of #1716 as discussed there: this is the behavior-change half, while #1716 now carries only the -inf sign correction.

With protocol=3, ",nan" currently falls through to a float cast, and (float) "nan" is 0.0 in PHP, so a NaN score silently becomes a valid-looking zero. This PR returns NAN instead. Per the RESP3 specification, Redis before 7.2 may emit any libc representation of NaN ("-nan", "NAN", "nan(char-sequence)") and clients should handle them, so those spellings are accepted too.

The TDigest RESP3 tests asserted 0 or null for empty sketches, which only passed because of the collapsed 0.0 (null == 0.0 loosely); their RESP2 counterparts already assert the string "nan" for the same replies. They now assert NaN. Verified against a live redis-stack: the full realm-stack group passes (1356 tests).

On the json_encode() concern from the #1716 discussion: it indeed throws on NAN, so consumers serializing raw replies must handle it, whereas today NaN hides behind a plausible 0.0. Scheduling this for whichever release you consider appropriate is entirely your call.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unlreleads
 ### Added
 ### Changed
+- Changed RESP3 double parsing to return `NAN` for NaN payloads instead of `0.0`
 ### Fixed
 
 ## v3.6.0 (2026-08-14)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,12 +1,8 @@
 ## Changelog
 
-## Unlreleads
-### Added
+## Unreleased
 ### Changed
 - Changed RESP3 double parsing to return `NAN` for NaN payloads instead of `0.0`
-### Fixed
-- Fixed RESP3 double parsing returning positive `INF` for `-inf` payloads (#1716)
-- Fixed `client_info` connection parameter being ignored (#1722)
 
 ## v3.6.0 (2026-08-14)
 ### Added
--- a/src/Protocol/Parser/Strategy/Resp3Strategy.php
+++ b/src/Protocol/Parser/Strategy/Resp3Strategy.php
@@ -66,7 +66,7 @@ class Resp3Strategy extends Resp2Strategy
         if ($string === 'inf') {
             return INF;
         }
-      
+
         if ($string === '-inf') {
             return -INF;
         }
--- a/src/Protocol/Parser/Strategy/Resp3Strategy.php
+++ b/src/Protocol/Parser/Strategy/Resp3Strategy.php
@@ -67,6 +67,12 @@ class Resp3Strategy extends Resp2Strategy
             return INF;
         }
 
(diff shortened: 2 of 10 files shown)


semx:
CI note: the two red jobs are HOTKEYS_Test::testRetrieveHotKeysResp3 ("array does not contain key1", sampling-dependent) and FTSEARCH_Test::testSearchWithEnhancedMatchingCapabilities, both on single PHP 7.2 combos; neither touches double parsing. The full realm-stack group passes against a live redis-stack locally.

vladvildanov:
@tillkruss How would we handle this breaking change PR's so we do not forced to release new major version? Are we keeping it open until we plan major release as next?

tillkruss:
@vladvildanov tbh with this one, I'd be happy to squeeze it in since it's more of an extension, but yeah I guess we keep it open and tag it as 4.x?

semx:
A few data points for the scheduling decision, from the reporter side:

1. **The current behavior is silent data corruption, not a lenient default.** 0.0 is a perfectly plausible score, so consumers cannot tell "actually zero" from "was NaN". The concrete case that surfaced this: TDigest on an empty sketch returns nan for MIN/MAX/quantiles, and predis reports it as 0.0 - monitoring built on top shows fake zero percentiles instead of "no data". Your own TDigest RESP2 tests assert the string "nan" for exactly these replies; RESP3 users get a fabricated number for the same server response.

2. **The protocol switch already changes this behavior unversioned.** The same server data yields "nan" (visible) on protocol=2 and 0.0 (invisible) on protocol=3 today. Anyone migrating to RESP3 already experiences a silent semantic change - this PR makes the two protocols agree instead.

3. **The json_encode() concern cuts the other way, in my view.** Today json_encode() happily serializes the fabricated 0.0, so the corruption propagates downstream cleanly. With NAN it throws - a loud failure at the boundary for a value that has no JSON representation, at a point where the caller can decide on a policy. That is the same class of fix as #1716: an obviously-wrong value replaced by the truthful one, except NaN->0.0 is arguably worse than -inf->+inf because the wrong value looks plausible.

4. RESP3 spec-wise, ",nan" (and the pre-7.2 libc spellings) are values clients "should be able to handle"; parsing them as 0.0 is non-compliant rather than conservative.

If a straight land in a minor still feels too hot, a middle ground would be shipping it now behind a connection/parser option defaulting to the current behavior, and flipping the default in 4.0 - happy to rework the PR that way if you prefer. Sitting unreleased until a major mostly means more RESP3 adopters accumulating on the silent-zero behavior in the meantime.

vladvildanov:
@tillkruss Well, maybe for this one we can make an exception. It's really on a borderline between real bug and incorrect interpretation. The impact should be really low potential RESP3 + TDigest users

PR #1719, opened 2026-09-01
Fixed CRLF command injection / smuggling in AbstractAggregateConnection::write()

vladvildanov:
## Summary

`AbstractAggregateConnection::write()` reconstructed commands by splitting the serialized buffer on \r\n and re-routing each chunk via `Command::deserializeCommand()`. Because that second parser ignores RESP bulk-length prefixes, \r\n sequences inside an attacker-influenced value or key (e.g. a cache key built from a URL slug) were interpreted as command boundaries and routed to a node — remote command injection on cluster connections (`FLUSHDB, DEL, SET`, same-slot `GET`) and a repeatable DoS on replication connections. This is `CVE GHSA-w6f5-v2h6-g786 / CWE-93`.                                                           
                                                                                                                                                                                                                 

The exploitable pipeline route was already closed in 3.3.0 (#1586), which made pipelines write each command through its real `Command` object. But the vulnerable parser itself remained in place — orphaned from Predis's own pipeline, yet still reachable through the public `ConnectionInterface::write()` API. This PR removes the sink.

## Change

1. `write()` no longer re-parses a raw buffer. Since an aggregate connection cannot route an opaque multi-command buffer without re-implementing a RESP framer — and nothing legitimately needs to after #1586 — it now refuses the operation.
`write()` is a public interface method, so making it throw is technically a behavioral change — but that call path is the vulnerability, and Predis itself has no caller that reaches it (pipelines use `writeToMultiNode() / writeRequest()`). A loud `NotSupportedException` was chosen over silent acceptance so any misuse fails visibly rather than corrupting data.

2. Deprecate `deserializeCommand()`. With the sink gone, `deserializeCommand()` has no production callers, but it survives on the public `CommandInterface`. It is unsafe by construction, not just in the removed caller: it `explode("\r\n")s` and steps two-at-a-time, discarding the `$<len>` prefixes that are the only reliable RESP framing. A single SET mykey "line1\r\nline2" round-trips into three corrupted arguments ['mykey', 'line1', ''], and it instantiates a command class from the parsed token. Leaving it in place invites the next contributor to reopen this exact CVE.                                                                                 
                                                                                                                                                                                                                   It is marked @deprecated on CommandInterface, Command, and RawCommand, scheduled for removal in the next major.


src/Command/Command.php
```diff
@@ -158,6 +158,9 @@ public function serializeCommand(): string
 
     /**
      * {@inheritDoc}
+     *
+     * @deprecated Not binary-safe; see CommandInterface::deserializeCommand().
+     *             Scheduled for removal in the next major.
```
tillkruss: @vladvildanov Should we split this up and mark some methods as deprecated in `main`, and fully remove them in `next`?

tillkruss (changes requested):

src/Command/Command.php
vladvildanov: Yeah, I'll take care

tillkruss (approved):

